PR #6, opened 2018-10-22
Create Basic CLI with Flags

### LINKED ISSUE #4: Need first implementation of CLI
First implementation of the CLI with a start command and basic options (port, etc.)

andrewnguyen22:
Fixes #4


andrewnguyen22:
Will resolve conflict upon approval

cmd/util/const.go
```diff
@@ -0,0 +1,12 @@
+// This package is for command line specific utilities.
+package util
```
luyzdeleon: Should we have a project global constant modules? I feel like version and client identifier are constants that could be used elsewhere.

cmd/util/flags.go
```diff
@@ -0,0 +1,53 @@
...
+ */
+func ParseFlags() {
+	flag.Parse()
+}
+
+/*
+"PrintClientInfo" prints startup information about the client
+ */
+func PrintClientInfo(){
+	fmt.Println("CLIENT ID:", clientIdentifier)
```
luyzdeleon: This is fine for now, but I think there's a better way to retrieve these values. Maybe the CMD should build some sort of configuration object that gets created based on the CLI flags the user typed in, then we can just do `Println(Configuration.description())` or something along those lines.

luyzdeleon (commented):
Minor comments.

cmd/util/const.go
andrewnguyen22: Agreed will add

cmd/util/flags.go
andrewnguyen22: That's a good idea

andrewnguyen22: See issue #9 and approve this PR

cmd/util/const.go
andrewnguyen22: See [commit](https://github.com/pokt-network/pocket-core/commit/7c91b982f490fdb42aa4ab1f583dcb084621432a) for change

luyzdeleon (dismissed):

luyzdeleon (dismissed):

luyzdeleon (approved):

PR #20, opened 2018-11-09
Issue #14 & 15

### LINKED ISSUE #14: Dispatch endpoint implementation: (finding existing sessions and creating new sessions)

andrewnguyen22:
Closes #14 and #15


session/session.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/pocket_network/pocket-core/node"
+)
+
+var (
+	globalSessionList *sessionList
+)
+/*
+This is the session structure.
+ */
+ // TODO consider converting validators and servicers to map
```
luyzdeleon, line 15: I think this should actually be done as a map, to avoid possible duplications, or a Set, look here for reference: https://softwareengineering.stackexchange.com/questions/177428/sets-data-structure-in-golang

session/session.go
```diff
@@ -0,0 +1,83 @@
...
+	return globalSessionList
+}
+
+/*
+"createNewSession" creates a new session for the specific devID and adds to global sessionList (map)
+ */
+func CreateNewSession(dID string) {
+	if(SearchSessionList(dID)==nil){
+		// pulls the global list from the singleton
+		sList :=GetSessionListInstance().sessions
```
luyzdeleon: This isn't too important, but syntactically it looks awkward that you have to get the `sessions` property out of the `sessionList ` object, shouldn't the `sessionList` be the actual list of sessions?

session/session.go
```diff
@@ -0,0 +1,83 @@
...
+		// simulated List of Servicers
+		// TODO turn into real list of servicers
+		servicers :=[]node.Service{}
+		// adds a new session to the sessionlist (map)
+		sList[dID]=Session{dID,validators, servicers}
+	}
+}
+
+/*
+"searchSessionList" searches the session list for the specific devID
```
luyzdeleon: Typo on the comment: `searchSessionList ` should be `SearchSessionList `

luyzdeleon (changes requested):
Please move the simple_curl_test file into a unit test.

session/session.go
andrewnguyen22: I purposefully did this so that the sessionList structure can become more robust in the future

andrewnguyen22: If you think that this is unnecessary then I can change it to just a list and not a custom object

session/session.go
andrewnguyen22: Done

session/session.go
luyzdeleon: If that's the case, I think it would be best to create a class that extends list, and add the future functionality there. Keeping the list elements abstracted away and accessible to the list inherited functionality.

andrewnguyen22: Structures cannot be extended like classes as in other object-oriented programming languages. We can just create a new struct embedding the struct that we require to use from a different package.

luyzdeleon (approved):

PR #21, opened 2018-11-09
Added global logging system

### LINKED ISSUE #12: Add global logging system

andrewnguyen22:
closes #12

### DIFF
--- a/cmd/pocket_core/main.go
+++ b/cmd/pocket_core/main.go
@@ -2,8 +2,11 @@
 package main
 
 import (
+	"github.com/logmatic/logmatic-go"
 	"github.com/pocket_network/pocket-core/config"
+	"github.com/pocket_network/pocket-core/logs"
 	"github.com/pocket_network/pocket-core/rpc"
+	"github.com/sirupsen/logrus"
 )
 
 //TODO add logging
@@ -20,6 +23,8 @@ func main() {
 "startClient" Starts the client with the given initial configuration.
  */
 func startClient(){
+	logs.LogConstructorAndLog("test.log","pocket-core/cmd/pocket_core/main.go",
+		"27", "Testing testing 123", logrus.DebugLevel, &logmatic.JSONFormatter{})
 	config.InitializeConfiguration()
 	config.PrintConfiguration()
 	rpc.RunAPIEndpoints()
--- a/cmd/pocket_core/main.go
+++ b/cmd/pocket_core/main.go
@@ -2,11 +2,9 @@
 package main
 
 import (
-	"github.com/logmatic/logmatic-go"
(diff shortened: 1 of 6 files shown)


cmd/pocket_core/main.go
```diff
@@ -2,8 +2,11 @@
 package main
 
 import (
+	"github.com/logmatic/logmatic-go"
```
luyzdeleon: Why is this logging dependency added to the CMD package?

cmd/pocket_core/main.go
```diff
@@ -2,8 +2,11 @@
...
 	"github.com/pocket_network/pocket-core/config"
+	"github.com/pocket_network/pocket-core/logs"
 	"github.com/pocket_network/pocket-core/rpc"
+	"github.com/sirupsen/logrus"
```
luyzdeleon: Same as above

cmd/pocket_core/main.go
```diff
@@ -20,6 +23,8 @@ func main() {
 "startClient" Starts the client with the given initial configuration.
  */
 func startClient(){
+	logs.LogConstructorAndLog("test.log","pocket-core/cmd/pocket_core/main.go",
```
luyzdeleon: Is there a way to get the file or class name dynamically? Having to type that every time seems too inconvenient.

I also feel that the way this interface works is too low level, the Logger interface should just be an instance that the whole package can just pass around and use.

cmd/pocket_core/main.go
```diff
@@ -20,6 +23,8 @@ func main() {
...
+		"27", "Testing testing 123", logrus.DebugLevel, &logmatic.JSONFormatter{})
```
luyzdeleon: A higher level interface shouldn't concern itself with how the logger works, there should be an abstraction so that the logger package implementation could switch without having to change the interface.

logs/models.go
```diff
@@ -0,0 +1,17 @@
+// This package is for logging runtime data.
+package logs
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+type Log struct{ // TODO consider optimizations with enum like structure to give options to log creator
```
luyzdeleon: Noted that for MVP we want to keep things as simple as possible, but design wise this is tightly coupling the Logger implementation throughout the rest of the modules.

luyzdeleon (changes requested):
Add the /Logs/test.log to the gitignore?

cmd/pocket_core/main.go
andrewnguyen22: Check it out now, should fix all

luyzdeleon (approved):

PR #23, opened 2018-11-09
added information response struct and method

### LINKED ISSUE #19: Create JSON Information Response
Turn this into readable JSON data for information endpoints

"Endpoint: <url>:<relay-port>/v1/dispatch/serve\n" +
"Method: DispatchServe\n" +
"Parameters:\n" +
"\t1) DeveloperID\n" +
"Returns: SessionID\n" +
"Example: POST: {\"method\":\"web3_clientVersion\",\"params\":[]}"

andrewnguyen22:
closes #19

### DIFF
--- a/rpc/shared/models.go
+++ b/rpc/shared/models.go
@@ -14,7 +14,7 @@ type Example struct {
 	Data  string `json:"data"`
 }
 
-type Information struct{
+type APIReference struct{
 	Endpoint 	string `json:"endpoint"`
 	Method 		string `json:"method"`
 	Parameters 	[]string `json:"params"`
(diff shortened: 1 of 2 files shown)


rpc/shared/models.go
```diff
@@ -14,3 +14,10 @@ type Example struct {
 	Data  string `json:"data"`
 }
 
+type Information struct{
```
luyzdeleon: Information is not a good name I believe for this struct.

luyzdeleon (commented):
I believe we should find a better name for the `Information` struct.

rpc/shared/models.go
andrewnguyen22: Check it out now? "APIReference"

luyzdeleon (approved):

PR #46, opened 2018-11-21
SessionID (Key) implemented

### LINKED ISSUE #33: Create session ID
Create a dummy interface to (eventually) derive a publicly verifiable pseudorandom sessionID

For now this can be whatever, just make it cryptographically swappable

### LINKED ISSUE #34: Return a sessionID (as dispatch node) when creating a session
Couple it within the data returned during a createSession call

### LINKED ISSUE #35: Document sessionID current naive algorithm
It is important to track and document the development and implementation of the sessionID algorithm. Start a doc now under the wiki

andrewnguyen22:
Closes #33 
Closes #34 
Closes #35

### DIFF
--- a/session/sessionKey.go
+++ b/session/sessionKey.go
@@ -4,9 +4,9 @@ package session
 import "github.com/pokt-network/pocket-core/crypto"
 
 /*
-"SessionKeyAlgo" function determines the sessionKey.
+"GenerateSessionKey" function determines the sessionKey.
  */
-func SessionKeyAlgo(devID string) []byte {
+func GenerateSessionKey(devID string) []byte {
 	// Simulate block hash
 	b1 := "block 1"
 	b2 := "block 2"
--- a/session/sessionKey.go
+++ b/session/sessionKey.go
@@ -1,12 +1,24 @@
 // This package is for all 'session' related code.
 package session
 
+import "github.com/pokt-network/pocket-core/crypto"
+
 /*
-This function determines the sessionKey.
+"SessionKeyAlgo" function determines the sessionKey.
  */
-func sessionKeyAlgo(devID string){
+func SessionKeyAlgo(devID string) []byte {
 	// Simulate block hash
-
+	b1 := "block 1"
+	b2 := "block 2"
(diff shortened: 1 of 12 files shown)


session/sessionKey.go
```diff
@@ -0,0 +1,24 @@
+// This package is for all 'session' related code.
+package session
+
+import "github.com/pokt-network/pocket-core/crypto"
+
+/*
+"SessionKeyAlgo" function determines the sessionKey.
+ */
+func SessionKeyAlgo(devID string) []byte {
```
luyzdeleon: Can we rename this function? I feel like the name is prone to confusion. something like GenerateSessionKey should be better.

tests/session/SessionKey_test.go
```diff
@@ -0,0 +1,41 @@
...
+	resp, err := client.Do(req)
+	// Handle errors
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+	// Deferred: close the body of the response
+	defer resp.Body.Close()
+	// Read the body from the response using ioutil
+	body, _ := ioutil.ReadAll(resp.Body)
+	t.Log(string(body))
```
luyzdeleon: Rather than printing the result, you should use some sort of assertion to make sure the Session Key is within the reseponse, and that the response has the correct structure. Also in your assertion you can generate a Session Key using the same function `SessionKeyAlgo ` and compare it to whatever is in your response body.

util/methods.go
```diff
@@ -1,6 +1,9 @@
 package util
 
-import "runtime"
+import (
+	"encoding/hex"
+	"runtime"
+)
 
 // MyCaller returns the caller of the function that called it :)
 func MyCaller() (*runtime.Func,uintptr){
```
luyzdeleon: Can we renam this to `FunctionCaller`? or `Caller`?

luyzdeleon (changes requested):
Some small changes, mostly nomenclature.

andrewnguyen22:
Changes Completed

luyzdeleon (dismissed):

PR #86, opened 2019-01-05
Makes the config module more idiomatic. Functionality no-op.

iajrz:
A number of measures are taken here to make the code more idiomatic: the
singleton pattern is simplified for the configuration object, the
package comments are centralized in a single file of the package to
avoid duplication, and the imports are organized according to the specs
outlined in effective go (this can be achieved automagically by setting
up goimport).
Removed some superfluous comments (where code is crystal clear on
intent), and some confusing ones (eg: build.go, line 26,
"using custom logs").
In config.go, some comments were besides the code in weird indentations.
These were moved above the code. This avoids weird alignment issues
produced in some development environments that inline-print information
about the code.


andrewnguyen22 (dismissed):

iajrz:
Merged to be up-to-date with Staging.

andrewnguyen22 (approved):

andrewnguyen22:
Just rebase to merge

PR #117, opened 2019-01-10
GID Flag and Node.Self Structure

### LINKED ISSUE #104: Service Node: GID Flag
--gid "dummyID" will be sent to Dispatch node upon registration

### LINKED ISSUE #123: Check For Hosted Chains
Simple port ping should suffice, ensure that each port is at least in use before allowing operation

andrewnguyen22:
Added flag: --gid "myGID"
Added node.Self singleton structure for easy self node properties like GID ^
Closes #104 
Closes #123

### DIFF
--- a/cmd/pocket_core/main.go
+++ b/cmd/pocket_core/main.go
@@ -52,16 +52,17 @@ func chainsFromFile(){
 "startClient" Starts the client with the given initial configuration.
 */
 func startClient() {
-	config.InitializeConfiguration()                                  // initializes the configuration from flags and defaults.
-	config.BuildConfiguration()                                       // builds the proper structure on pc for core client to operate.
-	config.PrintConfiguration()                                       // print the configuration the the cmd.
-	peersFromFile()                                                   // check for manual peers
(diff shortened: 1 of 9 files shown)


iajrz (changes requested):
[rpc.RunAPIEndpoints](https://github.com/pokt-network/pocket-core/pull/117/files#diff-1318addeee65b25b7124bace036f67d9R63) seems not to be available in this change set. Maybe there was a mixup when staging. This branch can't be built. 
[`ExitGracefully`](https://github.com/pokt-network/pocket-core/pull/117/files#diff-8a1278a5855334feb584493def4bfec5R8) should log the cause for exiting, not just print it.
I expected, because of the issue, that the default GID would be "dummyGID". 0 works too; if this was ignored on purpose, all is well.

andrewnguyen22:
Ahh yes, the manual merging is what did me in! Good catch, I fixed and made the changes you requested!

iajrz:
9e37f20 created an import cycle which prevented the project from building. I eliminated it in d125d79.

iajrz (dismissed):

PR #114, opened 2019-01-10
Entering and Exiting The Network Message Construction and Handling

andrewnguyen22:
closes #97 
closes #98 
closes #100

PR #118, opened 2019-01-10
Entering and Exiting Network Upon Launch

### LINKED ISSUE #105: Network Entry And Exit
Automatic network joining and exiting upon client open and close

andrewnguyen22:
Added UDP message to enter and exit the network upon launch and shutdown

Closes #105

### DIFF
--- a/cmd/pocket_core/main.go
+++ b/cmd/pocket_core/main.go
@@ -5,6 +5,7 @@ import (
 	"bufio"
 	"fmt"
 	"github.com/pokt-network/pocket-core/config"
+	"github.com/pokt-network/pocket-core/const"
 	"github.com/pokt-network/pocket-core/crypto"
 	"github.com/pokt-network/pocket-core/logs"
 	"github.com/pokt-network/pocket-core/message"
@@ -50,7 +51,13 @@ func chainsFromFile(){
 
 //NOTE: this is for centralized dispatch and may be removed at production
 func sendEntryMessage(){
-	message.NewEnterNetMessage()
+	m:=message.NewEnterNetMessage()
+	message.SendMessage(message.RELAY, m, _const.DISPATCHIP, message.EnterNetworkPayload{})
+}
+
+func sendExitMessage(){
+	m:=message.NewExitNetMessage()
+	message.SendMessage(message.RELAY, m, _const.DISPATCHIP, message.ExitNetworkPayload{})
 }
 
 /*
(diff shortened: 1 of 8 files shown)


cmd/pocket_core/main.go
```diff
@@ -48,20 +49,33 @@ func chainsFromFile(){
...
-	message.RunMessageServers()                                       // runs servers for messages
-	fmt.Print("Press any key + 'Return' to quit: ")                   // prompt user to exit
-	input := bufio.NewScanner(os.Stdin)                               // unnecessary temporary entry
-	input.Scan()                                                      // wait
+	message.RunMessageServers()													// runs servers for messages
+	sendEntryMessage()															// send entry message
+	fmt.Print("Press any key + 'Return' to quit: ")                 			// prompt user to exit to dispatcher
+	input := bufio.NewScanner(os.Stdin)                             			// unnecessary temporary entry
+	input.Scan()                                                    			// wait
+	sendExitMessage()															// send exit message to dispatcher
```
iajrz, line 81: Maybe this should be "deferred", so that if we exit in another way, we still get the message to go through most times.

const/dispatcher.go
```diff
@@ -0,0 +1,6 @@
+package _const
+
+const(
+	DISPATCHGID ="DISPATCH"
+	DISPATCHIP="127.0.0.1"
```
iajrz, line 5: I guess this is going to be changed in the future? Probably dispatcher IP should come from the config file.

util/ip.go
```diff
@@ -0,0 +1,26 @@
+package util
+
+import (
+  "fmt"
+  "io/ioutil"
+  "net/http"
+)
+
+func GetIPAdress() (string, error) { // TODO wonering if there is a cleaner way to get public ip.
+  url := "https://api.ipify.org?format=text"	// we are using a pulib IP API, we're using ipify here, below are some others
```
iajrz, line 10: Maybe the provider for public address should be optional? Maybe not now, but in the future - have the public IP provider come from a config file, maybe use this as the default or backup/fallback option.

iajrz (changes requested):
Maybe the enter and exit messages should be TCP? We want to make sure the dispatcher knows who's in and who's out, be aware of the dispatcher not receiving the message and reacting accordingly, too.

cmd/pocket_core/main.go
andrewnguyen22: Nice! Good call. However, I believe OS.Exit() does not execute deffered functions, wondering if there is a better way to exit the program gracefully?

const/dispatcher.go
andrewnguyen22: Yes!

util/ip.go
andrewnguyen22: Interesting, thought. I think that overall, the config file would be more of a headache for the user. However, giving the option may allow for more flexibility, a good future addition!

andrewnguyen22:
> Maybe the enter and exit messages should be TCP? We want to make sure the dispatcher knows who's in and who's out, be aware of the dispatcher not receiving the message and reacting accordingly, too.

Yes! So as I mentioned in #114 I think that TCP is the better option, however a new issue is deemed in the next sprint since this wasn't the original plan

iajrz (dismissed):

PR #122, opened 2019-01-14
Node Liveness Check

andrewnguyen22:
Simple liveness check:
````
For each node within peerlist
{
  Ping(IP + : + Relay port + /v1/)
  if response ! okay {
     remove from peerlist // code actuallly checks twice before removing peer
  }
}
````


iajrz (approved):
👍 adds the capacity to make and respond to live checks~ looks good to me

PR #125, opened 2019-01-14
Updated Readme From Slack to Telegram

andrewnguyen22:
closes #125


iajrz (commented):
It's possible that we want the "How to run it" part to be auto-generated from `help`'s output, so as to never be outdated. Just a thought.

iajrz (approved):

andrewnguyen22:
Good thought! Any ideas on implementation?

iajrz:
I suggest we leave the idea marinate until we have a more complex build system.

One way to do it would be implementing a GitHub Action triggered by merges into staging.

PR #126, opened 2019-01-14
Cross Reference WL Before Operations

### LINKED ISSUE #120: Cross check developer and dispatch white list
Must cross check white list structure before:

(DEVELOPERS) Dispatch Serving, Relaying
(SERVICE NODES) Adding to peerlist

andrewnguyen22:
closes #120


rpc/relay/dispatch.go
```diff
@@ -33,6 +33,13 @@ func DispatchServe(w http.ResponseWriter, r *http.Request, ps httprouter.Params)
 	dispatch := &Dispatch{}
 	result := make(map[string][]string)
 	shared.PopulateModelFromParams(w, r, ps, dispatch)
+	// cross check the WL
+	if !node.GetDeveloperWhiteList().Contains(dispatch.DevID){
```
iajrz: Maybe this should be extracted into a function called something like "ensureDeveloperIsInWhitelist". Then we could make do without the comments around this snippet, because we're choosing names that document our intent.

rpc/relay/relay.go
```diff
@@ -28,6 +29,11 @@ func RelayOptions(w http.ResponseWriter, r *http.Request, ps httprouter.Params)
 func RelayRead(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	relay := &Relay{}                               // create empty relay structure
 	shared.PopulateModelFromParams(w, r, ps, relay) // populate the relay struct from params //TODO handle error for populate model from params (in all cases within codebase!)
+	if !node.GetDeveloperWhiteList().Contains(relay.DevID){
```
iajrz: Maybe this snippet can be extracted into a function too.

rpc/relay/dispatch.go
andrewnguyen22: Haha funny you suggest this! I was thinking the same thing at the time, but I last minute decided to go with  the comments. Will implement now!

rpc/relay/relay.go
andrewnguyen22: Done!

iajrz (dismissed):

PR #136, opened 2019-01-24
Changed default flags

### LINKED ISSUE #134: Default Flag Values
The configuration for the flags should be set up in a way that makes it easy for someone to just download the project and run it with the least possible hassle, then tune and tweak their setup as they learn the ropes.

In particular, the `clientrpc` and `relayrpc` flags would be helpful if they were set to `true`.

andrewnguyen22:
Closes #134


iajrz (commented):
simple, works for me.

iajrz (approved):

PR #137, opened 2019-01-25
Added nested structure & slice parsing for in-client API ref

### LINKED ISSUE #111: Fix RPC Info Struct
Currently, the information structure is not detailed enough for real world usage. 

For instance, the dispatch/serve information only suggests to curl 'blockchains' but does not provide the blockchain structure

andrewnguyen22:
closes #111

### DIFF
--- a/rpc/shared/info.go
+++ b/rpc/shared/info.go
@@ -8,7 +8,6 @@ import (
 )
 
 // "CreateInfoStruct" generates the specific APIReference structure dynamically.
-// TODO adapt this for slice structures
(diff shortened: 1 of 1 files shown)


rpc/shared/info.go
```diff
@@ -10,21 +11,69 @@ import (
 // TODO adapt this for slice structures
```
iajrz: the title of this PR makes me think this TODO is completed.

andrewnguyen22: Indeed it is completed

iajrz (approved):
lgtm

PR #148, opened 2019-01-31
Parameters field now references the type and example field references parameters

andrewnguyen22:
Ignore the commits from 2 days ago, those are stale and already merged into Staging. Sorry about that


iajrz (dismissed):
👍

PR #150, opened 2019-02-01
fixed gets in cmd dispatch message node and session modules, repaired…

### LINKED ISSUE #149: Fix broken tests, remove 'Get<>' according to Effective Go
Note: still haven't found a suitable replacement for config.Get() but working on it

andrewnguyen22:
… broken tests from manual merge (oops), and abstracted the GetRoutes() call to one shared.GetRoutes() call

Note: some of the 'get<>' are necessary because of duplicate names EX: GetRoutes() and Routes() are two different functions at this point.

Closes #149

### DIFF
--- a/tests/rpc/dispatch_test.go
+++ b/tests/rpc/dispatch_test.go
@@ -4,14 +4,15 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/pokt-network/pocket-core/config"
-	"github.com/pokt-network/pocket-core/node"
-	"github.com/pokt-network/pocket-core/rpc/relay"
-	"github.com/pokt-network/pocket-core/rpc/shared"
 	"io/ioutil"
 	"net/http"
 	"reflect"
 	"testing"
+
+	"github.com/pokt-network/pocket-core/config"
+	"github.com/pokt-network/pocket-core/node"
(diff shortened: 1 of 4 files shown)


luyzdeleon (approved):
lgtm

luyzdeleon:
> Note: some of the 'get<>' are necessary because of duplicate names EX: GetRoutes() and Routes() are two different functions at this point.

If we have this issue, then there's a nomenclature problem beyond the `Get` prefix. Please create a separate issue to highlight the instances where this is happening.

iajrz:
> > Note: some of the 'get<>' are necessary because of duplicate names EX: GetRoutes() and Routes() are two different functions at this point.
> 
> If we have this issue, then there's a nomenclature problem beyond the `Get` prefix. Please create a separate issue to highlight the instances where this is happening.

#149 is about this - Andrew already created it.

luyzdeleon:
> > > Note: some of the 'get<>' are necessary because of duplicate names EX: GetRoutes() and Routes() are two different functions at this point.
> > 
> > 
> > If we have this issue, then there's a nomenclature problem beyond the `Get` prefix. Please create a separate issue to highlight the instances where this is happening.
> 
> #149 is about this - Andrew already created it.

Got it! Thanks for the clarification.

PR #151, opened 2019-02-04
Issue #149

andrewnguyen22:
Forgot to add these changes to the last PR sorry guys!

### DIFF
--- a/cmd/pocket_core/main.go
+++ b/cmd/pocket_core/main.go
@@ -36,7 +36,7 @@ func startClient() {
 	// print the configuration the the cmd
 	config.Print()
 	// add peers to dispatch structure
-	node.GetPeerList().CopyToDP()
+	node.PeerList().CopyToDP()
 	// check for hosted chains
 	node.TestChains()
 	// runs the server endpoints for client and relay api
--- a/config/build.go
+++ b/config/build.go
@@ -19,7 +19,7 @@ func Build() {
 // "dataDir" builds the directory for program files.
 func dataDir() {
 	// attempts to make the data directory.
-	if err := os.MkdirAll(Get().DD, os.ModePerm); err != nil {
+	if err := os.MkdirAll(GlobalConfig().DD, os.ModePerm); err != nil {
 		// doesn't use custom logs, because they may or may not be available at this point
 		log.Fatalf(err.Error())
 	}
@@ -28,7 +28,7 @@ func dataDir() {
 // "logsDir" builds the directory for logs.
 func logsDir() {
 	// attempts to make the logs directory
-	if err := os.MkdirAll(Get().DD+_const.FILESEPARATOR+"logs", os.ModePerm); err != nil {
+	if err := os.MkdirAll(GlobalConfig().DD+_const.FILESEPARATOR+"logs", os.ModePerm); err != nil {
 		log.Fatal(err.Error())
 	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -45,7 +45,7 @@ func Init() {
 	// built in function to parse the flags above.
 	flag.Parse()
 	// returns the thread safe c of the client configuration.
-	Get()
+	GlobalConfig()
 }
 
 // "newConfiguration() is a constructor function of the configuration type.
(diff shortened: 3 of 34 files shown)


iajrz (commented):
I have been thinking about this a lot, specially since I'm working with some legacy C++ code and things like the app version information "could be" a singleton. Maybe some of the thought process applies here.

_The class_ is not a singleton; _the object_ is. So you want to make sure everyone who _thinks_ they have certain information, like the configuration, have the same object - but there could be other "configuration" objects which aren't the current live configuration. In the case of the app version - you could have two instances for comparison purposes, but only one of those is the current app version.

Maybe you see where I'm going with this. The whole class need not be a singleton - in some cases _should not be_. But whoever wants a particular instance with a particular meaning, should call a specific method, which is _very clear_ on what instance you're getting. `GlobalConfiguration` or `LiveConfiguration` or `CurrentConfiguration`; `CurrentlyHostedChains`; `LivePeers` <- this one in particular as this list can now more specifically maintain live peers, and many similar instances differently named contain other categories of peers.

Then in the case of, for instance, the configuration, we should be able to mutate it thread-safely when we need to - but that's another issue. What do you think?

andrewnguyen22:
Yeah, so I like 'config.GlobalConfiguration()'. I think it has a nice ring to it and makes it very clear that this is the single shared instance.

andrewnguyen22:
However, I think that brevity is always preffered in golang according to 'Effective Go' so I don't think that LivePeers is necessary considering if the peers are not 'live' they would not be within our memory. 

Just to be clear, the whole point of this PR is to fix according to https://golang.org/doc/effective_go.html#Getters

iajrz:
Brevity of typing is valuable, but clarity if intent and understanding is priceless.

Generic constructors are for generic instances. If an instance's behavior is important to the code, the name helps reflect that. `getInstance` works because the wild popularity of "Design Patterns": readers grasp by context that this is a singleton and there is a reason for that. On the other hand, `Peers`, for instance - there is nothing in the name to let other coders understand that, to quote you:
> if the peers are not 'live' they would not be within our memory.

`LivePeers`, on the other hand, makes you expect that Peers can stop; that somewhere in the code the list is kept up to date; and that in this list you will find only the Peers that are live.

The motivation being about `get`ters is with regards to the "traditional" getters - ie: to get instance members, in the context of being able to control, with `set`ters and `get`ters, the way the value of the member is changed and communicated to the rest of the world respectively.

andrewnguyen22:
Hmm an interesting thought, I'll think about this ^

iajrz (approved):
Meanwhile - the "Get<>" were removed, names are clearer, we can keep thinking about names and improving overall in time.

iajrz:
Don't merge. Tests aren't all running.

```
ok  	github.com/pokt-network/pocket-core/tests/config	0.010s
--- FAIL: TestSessionMessage (2.01s)
    session_test.go:42: No peers within peerlist
FAIL
FAIL	github.com/pokt-network/pocket-core/tests/message	2.025s
ok  	github.com/pokt-network/pocket-core/tests/node	1.674s
Invalid Credentials
--- FAIL: TestDispatchServe (0.00s)
    dispatch_test.go:74: Unable to unmarshall json node response : invalid character 'I' looking for beginning of value
open /Users/iajrz/Library/Pocket/chains.json: no such file or directory
--- FAIL: TestRelay (0.00s)
    relay_test.go:25: open /Users/iajrz/Library/Pocket/chains.json: no such file or directory
Expected Body: "Okay! The node has been successfully reported to our servers and will be reviewed! Thank you!"
Received Body {
	"data": "Okay! The node has been successfully reported to our servers and will be reviewed! Thank you!"
}
2019/02/04 11:04:22 Body is not as expected
FAIL	github.com/pokt-network/pocket-core/tests/rpc	0.020s
ok  	github.com/pokt-network/pocket-core/tests/service	0.021s
```

iajrz (changes requested):
fix tests

iajrz:
The files needed for tests should be fixtures.

andrewnguyen22:
So the unit tests are working properly. The problem is as @iajrz the config files needed for the unit tests are  not fixtures. I will create a separate issue for this and suggest to approve and merge this PR.

iajrz:
Makes sense; I'll create a little script to automate checking out the branch, running the tests, and starting the server - just to have it all running while I read the PR and make sure I'm not missing anything.

iajrz (approved):

andrewnguyen22:
See #152

Issue #155, opened 2019-02-04
Dynamic 'Flags' portion of readme from hosted dispatcher


andrewnguyen22:
-clientrpc
    	whether or not to start the rpc server
  -clientrpcport string
    	specified port to run client rpc
  -datadir string
    	setup the data directory for the DB and keystore
        (default: `%APPDATA%\Pocket` for Windows, `~/.pocket` for Linux, `~/Library/Pocket` for Mac)
  -dwl string
    	specifies the filepath for developer_whitelist.json
  -gid string
    	set the selfNode.GID for pocket core mvp
  -hostedchains string
    	specifies the filepath for hosted chains
  -peerFile string
    	specifies the filepath for peers.json
  -relayrpc
    	whether or not to start the rpc server
  -relayrpcport string
    	specified port to run relay rpc
  -snwl string
    	specifies the filepath for service_whitelist.json

This can be dynamically generated by provining a simple API endpoint

PR #164, opened 2019-02-06
added dynamoDB functionality

### LINKED ISSUE #156: Integrate DynamoDB into PC MVP

### LINKED ISSUE #173: Integrate DynamoDB to AWS cloud

andrewnguyen22:
Note: The DB tests won't run correctly without the database being built. I will look into the best CI configuration to make this work well. I predict that the CI will need aws credentials and an installed version of dynamodb

closes #156 
EDIT:

Now closes #156 and #155 

EDIT2:

Now closes #173

### DIFF
--- a/cmd/db/dynamodb.go
+++ b/cmd/db/dynamodb.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -39,16 +40,20 @@ func DB() *dynamodb.DynamoDB {
 
 func main() {
 	var i int
-	fmt.Print("(1) Create DP Table:\n(2) Delete DP Table: ")
-	_, err := fmt.Scanf("%d", &i)
-	if err != nil {
-		fmt.Println(err)
-	}
-	switch i {
-	case 1:
-		CreateTable()
-	case 2:
-		DeleteTable()
+	for {
+		fmt.Print("(1) Create DP Table:\n(2) Delete DP Table:\n(0) Quit\n")
+		_, err := fmt.Scanf("%d", &i)
+		if err != nil {
+			fmt.Println(err)
+		}
+		switch i {
+		case 1:
+			CreateTable()
+		case 2:
+			DeleteTable()
+		default:
+			os.Exit(0)
+		}
 	}
 
 }
--- a/cmd/db/dynamodb.go
+++ b/cmd/db/dynamodb.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"fmt"
+	"sync"
+	
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/pokt-network/pocket-core/const"
+)
+
+// NOTE: You need to setup your AWS configuration
+// first https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+
+const (
+	REGION   = "us-west-2"
+	ENDPOINT = "http://localhost:8000"
+)
+
+var (
+	c    *aws.Config
+	once sync.Once
+)
+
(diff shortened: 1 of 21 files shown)


cmd/db/dynamodb.go
```diff
@@ -0,0 +1,103 @@
...
+	_, err := fmt.Scanf("%d", &i)
+	if err != nil {
+		fmt.Println(err)
+	}
+	switch i {
+	case 1:
+		CreateTable()
+	case 2:
+		DeleteTable()
+	}
```
iajrz, line 57: This behaves better as part of a while loop, starting at about line 42, and with a default case that would advise the user to choose either 1 or 2 depending on the situation and restart.

rpc/client/live.go
```diff
@@ -14,7 +14,7 @@ import (
 func Register(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	n := &node.Node{}
 	if err := shared.PopModel(w, r, ps, n); err != nil {
-		shared.WriteJSONResponse(w, "500 error: "+err.Error())
+		shared.WriteErrorResponse(w, 500, err.Error())
```
iajrz: I think this is more likely a 400 Bad Request

rpc/client/live.go
```diff
@@ -25,14 +25,14 @@ func Register(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
...
-	shared.WriteJSONResponse(w, "Invalid credentials")
+	shared.WriteErrorResponse(w, 401, "Invalid credentials")
 }
 
 // "Register" handles the localhost:<client-port>/v1/register call.
 func UnRegister(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 	n := &node.Node{}
 	if err := shared.PopModel(w, r, ps, n); err != nil {
-		shared.WriteJSONResponse(w, "500 error: "+err.Error())
+		shared.WriteErrorResponse(w, 500, err.Error())
```
iajrz: I think this is more likely a 400 Bad Request

util/ip.go
```diff
@@ -1,14 +1,12 @@
...
 
 import (
-	"fmt"
 	"io/ioutil"
 	"net/http"
 )
 
 func IP() (string, error) {
 	url := "https://api.ipify.org?format=text"
-	fmt.Printf("Getting IP address from  ipify ...\n")
```
iajrz, line 11: Maybe this could be an Info level log instead of not existing anymore

cmd/db/dynamodb.go
andrewnguyen22: Agreed ^ will fix this

rpc/client/live.go
andrewnguyen22: Yes agreed, a badly formatted request should be the only reason for error. Will fix this

util/ip.go
andrewnguyen22: Yes good thought!

andrewnguyen22:
Requesting a new review! @iajrz @luyzdeleon

cmd/db/dynamodb.go
```diff
@@ -39,16 +40,20 @@ func DB() *dynamodb.DynamoDB {
...
+		_, err := fmt.Scanf("%d", &i)
+		if err != nil {
+			fmt.Println(err)
+		}
+		switch i {
+		case 1:
+			CreateTable()
+		case 2:
+			DeleteTable()
+		default:
```
iajrz: this does essentially the same as not having the loop; I meant to have a loop such that allowed you to "try again" if you made a mistake in the input (say, pressed '12' or '`' by mistake.)

andrewnguyen22: Oh is that what you meant? I don't really see the need for that kind of error checking?

andrewnguyen22: This loop allows you to create and delete the table in one running session.

andrewnguyen22: In any case ^ I have updated the code to reflect the result you desired

iajrz (dismissed):

PR #166, opened 2019-02-06
Shutdown if chains file problem

### LINKED ISSUE #158: Shutdown if no hosted chains file || Hosted chain is unavailable
--chains flag <path to file> if not there... shutdown

andrewnguyen22:
closes #158 

Added `ExitGracefully()` if problem with the chains.json

If you want to operate without hosted chains (development and dispatcher only):

1) Create chains.json in datadirectory
2) Write empty json array `[]` in the file

Should work fine!

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -3,6 +3,7 @@ package node
 import (
 	"github.com/pokt-network/pocket-core/config"
 	"github.com/pokt-network/pocket-core/logs"
+	"github.com/pokt-network/pocket-core/util"
 )
 
 func Files() {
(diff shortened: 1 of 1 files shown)


iajrz (approved):
Simple, straightforward, beautiful. If we find ourselves printing the config in many places at exit, we should consider having a function for "exiting and reporting" at the same time in util, so as not to repeat ourselves. It would also help other coders - instead of having to remember or think about the convention of printing the config File in ExitGracefully, it'd be more communicative.

Food for thoughts about the future.

Issue #113, opened 2019-01-10
Convert Unit Testing To Integration Tests In Playground


andrewnguyen22:
Any unit test that is simulating two different machines should be converted to integration tests, as there is race conditions in multithreaded tests 

Essentially, any unit test with time.Sleep() should be converted as this is not an acceptable solution

iajrz:
I posit we should do mutation testing (to test the tests' relevance to our code), as well as our "regular" tests. This is crucial as we will need to build up _confidence of correctness_. This is a tool for mutation testing in go. https://github.com/zimmski/go-mutesting

andrewnguyen22:
See #185 & #186

PR #167, opened 2019-02-06
Issue #152

### LINKED ISSUE #152: Unit tests need fixture config files to run properly
Create a separate fixtures module within tests module for configuration

andrewnguyen22:
Closes #152 

We have a few tests with "unavoidable" data races -> multithreaded functionality. 

So far the solution is to just simply convert these tests to staging environment when ready. However, if anyone @iajrz or @luyzdeleon has a better suggestion about how to fix these data races. Please let me know. Else, I will eventually phase out the tests with the race conditions once staging is available. 

See https://github.com/pokt-network/pocket-core/issues/113 for more details on current solution

### DIFF
--- a/rpc/relay/relay.go
+++ b/rpc/relay/relay.go
@@ -2,7 +2,7 @@ package relay
 
 import (
 	"net/http"
-
+	
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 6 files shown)


luyzdeleon (approved):
lgtm

PR #203, opened 2019-02-22
Issue #200 (MVP)

### LINKED ISSUE #200: Add dbregion flag to configuration
Add a `--dbregion` flag, specifying where the AWS region the dynamodb instance we're connecting to is running.

### LINKED ISSUE #201: Add `host` attribute to HostedChain struct
HostedChain struct: https://github.com/pokt-network/pocket-core/blob/mvp-staging/node/chain.go#L24

1. Add a `host` attribute to struct.
2. Parse the `host` attribute from chains.json

andrewnguyen22:
Closes #200 
Closes #201


db/config.go
```diff
@@ -30,7 +29,7 @@ func DB() *Database {
 		db = &Database{}
 		var c *aws.Config
 		c = &aws.Config{
-			Region:   aws.String(_const.DBREIGON),
+			Region:   aws.String(config.GlobalConfig().DBRegion),
```
luyzdeleon: Question, why do we pull the GlobalConfig object here, but in the `Endpoint` below we use the `con` instance to pull it? Wouldn't it be better to keep it consistent?

andrewnguyen22: Oops! Good Catch ;)

luyzdeleon (approved):

Issue #222, opened 2019-02-26
Documentation Overhaul


iajrz:
It's important, to avoid confusion, to point people at the [Developer Setup Guide](https://github.com/pokt-network/pocket-core/wiki/Developer-Setup-Guide) earlier in the README.md; reading through the document tempts one to tinker before having all the information; this would be OK if the steps yielded a running app, alas, they don't.

We should consider reorganizing the Developer Setup Guide. The "quick start" commands can be at the beginning, and the details at the end, yielding a natural progression for the curious and an easy startup for those in a hurry.

We should consider organizing the command documentation in a way more reminiscent of what you'd find any day in the command line. Grouping them by function would also help developers map the commands to what they want to do faster.

andrewnguyen22:
I don't have any problem with this. @luyzdeleon initially designed the dev guide so I'd like to hear his input

jeremybeal11:
@iajrz  i can assist in cleaning it up. I'm working on my first setup with PocketCore and i can leave some useful notes to help out.

iajrz:
It was left open on purpose ;) some of this applies to Core. Thanks for keeping things tidy :)

PR #246, opened 2019-03-05
Reorganizes readme to reduce confusion

iajrz:
makes references to latter sections (such as building and the developer guide) show up from earlier in the readme to avoid confusion when trying to setup the project for the first time.
Splits the flags into categories: some are for the dispatchers, some for the service nodes.


iajrz:
This partially addresses #222, although more work on that needs to be done.

andrewnguyen22 (approved):
Looks good, I like the org for MVP only of course

PR #253, opened 2019-03-09
Feature/add integration unit tests

aqt01:
- Add prettier code format for defining args on `cmd.sh`
- Add default values for all `POCKET_CORE` values in case no value is defined
- Add prefix `POCKET_CORE`for DYNAMODB enviroment variables
- Add required `args` for running integration tests

### DIFF
--- a/cmd.sh
+++ b/cmd.sh
@@ -24,15 +24,28 @@ set -o nounset
 # POCKET_CORE_SERVICE_IP = IP of the Pocket Core service node (required)
 # POCKET_CORE_SERVICE_PORT = Port of the Pocket Core service node (required)
 
-
-
 # Start pocket-core
 if [ $POCKET_CORE_NODE_TYPE = "dispatch" ]; then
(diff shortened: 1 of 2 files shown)


andrewnguyen22:
@aqt01 Hmm should the dispatcher IP and port should point to the publicly accessible endpoint? Or is thi specifically for dockerized instances? If this will run in staging only, then it would be useful to point the `--disip dispatch.staging.pokt.network --disrport 80`

aqt01:
> @aqt01 Hmm should the dispatcher IP and port should point to the publicly accessible endpoint? Or is thi specifically for dockerized instances? If this will run in staging only, then it would be useful to point the `--disip dispatch.staging.pokt.network --disrport 80`

Both. I leave the default variables in order to run the entire pocket-core ecosystem locally using docker. If you set env variables for those it will be overriden as you can see in `cmd.sh`

andrewnguyen22 (approved):

PR #254, opened 2019-03-11
fixed localhost bug in integration tests

andrewnguyen22:


plugin/rpc/rpc.go
```diff
@@ -3,19 +3,30 @@ package rpc
...
+	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
 // "ExecuteRequest" takes in the raw json string and forwards it to the port
 func ExecuteRequest(jsonStr []byte, host string, port string) (string, error) {
+	if !strings.Contains(host, "http") {
```
luyzdeleon, line 15: What if it's https?

tests/integration/fixtures/unregister.json
```diff
@@ -8,7 +8,7 @@
   "Blockchains": [
     {
       "Name": "ethereum",
```
luyzdeleon: Wrong blockchain name here.

luyzdeleon (changes requested):

plugin/rpc/rpc.go
andrewnguyen22: The contains function just looks for a substring, thus if the host contains https it will still pass. This is a QOL thing that fixes if someone does `localhost` or ip address without http

tests/integration/fixtures/unregister.json
andrewnguyen22: Good catch! Fixed

luyzdeleon (approved):

PR #265, opened 2019-03-14
Add integration tests for dispatch and service after running server

aqt01:


cmd.sh
```diff
@@ -24,28 +24,65 @@ set -o nounset
...
+		nohup pocket-core --dispatch \
+		  --datadirectory ${POCKET_PATH_DATADIR:-datadir} \
+		  --dbend ${POCKET_CORE_AWS_DYNAMODB_ENDPOINT:-dynamodb.us-east-1.amazonaws.com} \
+		  --dbtable ${POCKET_CORE_AWS_DYNAMODB_TABLE:-dispatch-peers-staging} \
+		  --dbregion ${POCKET_CORE_AWS_DYNAMODB_REGION:-us-east-1} \
+		  --disip ${POCKET_CORE_DISPATCH_IP:-127.0.0.1} \
+		  --disrport ${POCKET_CORE_DISPATCH_PORT:-8081} &
+
+		sleep 10 
+		exec go test ./tests/integration/... --dispatch --dispatchtesturl dispatch.staging.pokt.network --servicetesturl service.staging.pokt.network
```
luyzdeleon: Can we pass the URL's as environment variables?

luyzdeleon (changes requested):

cmd.sh
```diff
@@ -55,7 +55,7 @@ if [ $POCKET_CORE_NODE_TYPE = "dispatch" ]; then
 		  --disrport ${POCKET_CORE_DISPATCH_PORT:-8081} &
 
 		sleep 10 
-		exec go test ./tests/integration/... --dispatch --dispatchtesturl dispatch.staging.pokt.network --servicetesturl service.staging.pokt.network
+		exec go test ./tests/integration/... --dispatch --dispatchtesturl ${POCKET_CORE_DISPATCH_IP:-127.0.0.1} --servicetesturl ${POCKET_CORE_DISPATCH_IP:-127.0.0.1}
```
luyzdeleon: Shouldn't this be it's own env variable? and you used the same in both flags.

cmd.sh
```diff
@@ -81,7 +81,7 @@ elif [ $POCKET_CORE_NODE_TYPE = "service" ]; then
 			--port ${POCKET_CORE_SERVICE_PORT:-8081} &
 		
 		sleep 10 
-		exec go test ./tests/integration/... --dispatchtesturl dispatch.staging.pokt.network --servicetesturl service.staging.pokt.network
+		exec go test ./tests/integration/... --dispatchtesturl ${POCKET_CORE_DISPATCH_IP:-127.0.0.1} --servicetesturl ${POCKET_CORE_DISPATCH_IP:-127.0.0.1}
```
luyzdeleon: see above

luyzdeleon (changes requested):

cmd.sh
aqt01: Oh, sorry. It's Done already

cmd.sh
aqt01: I think we have to tests the same variables that the server is using to connect

luyzdeleon (approved):

PR #277, opened 2019-03-21
Milestone proposal + formatting

andrewnguyen22:
@luyzdeleon @iajrz if you guys notice any changes in terminology it is due to the fact that you cannot nest inside of 'it' statments in Ginkgo and Gomega


tests/bdd/session/session_test.go
```diff
@@ -0,0 +1,190 @@
...
+			Context("Parameters are missing or null", func() {
+
+				It("should error", func() {
+					// Code goes here ...
+				})
+			})
+
+			Context("Devid1 is incorrect...", func() {
+
+				Context("Devid is incorrect format", func() {
```
iajrz, line 35: I would remove this layer of context and go directly to the "it"

tests/bdd/session/session_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+					It("should error", func() {
```
iajrz: This is a good place (as elsewhere in this PR) to say what the error should be. "It should return an 'invalid developer id' error" or something similar. And somewhere we should answer: to whom is that error returned, and how?

iajrz (changes requested):
All in all, I am happy with the happy paths, and not so happy with the unhappy paths. Who would've thought we'd have such symmetry 😝 

The main issue is the description of the errors or failure states, which is mostly a wording thing, but also has implications about how we are designing the process. Thanks for this contribution, keep it up!

tests/bdd/session/session_test.go
luyzdeleon: I agree with Irving on this one, as much as behaviours inform interfaces, they also inform errors and how we design them.

andrewnguyen22: Okay, since the sprint should start today, I think I should go ahead and start developing based on the 'happy paths' we agree on. Then we can meet to discuss this further

tests/bdd/session/session_test.go
luyzdeleon: I actually disagree with this one, it would improve brevity, however it reads better as it is right now.

luyzdeleon (approved):

PR #436, opened 2019-09-03
Issue #416

### LINKED ISSUE #416: Update Session Generation
The legacy code is not up to current specification

Update the legacy session generation with a focus on readability.

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
Needs to pass legacy BDD

andrewnguyen22:
Closes #416


tests/bdd/session/legacySession_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+// ************************************************************************************************************
+// Milestone: Sessions
+//
+// Tentative Timeline (4-6 weeks)
+//
+// Unanswered Questions?
+// - Where is the seed data coming from? 99% sure -> *WORLD STATE*
+// - Is service request validation ddos safe?
+// - Forking behavior
```
luyzdeleon, line 20: You can remove this one, as tendermint won't allow forks.

x/session/session.go
```diff
@@ -0,0 +1,25 @@
...
+
+type Session struct {
+	SessionKey     SessionKey        `json:"sessionkey"`
+	Developer      SessionDeveloper  `json:"developer"`
+	NonNativeChain SessionBlockchain `json:"nonnativechain"`
+	BlockID        SessionBlockID    `json:"latestBlock"`
+	Nodes          SessionNodes      `json:"sessionNodes"`
+}
+
+// Create a new session from seed data
```
luyzdeleon, line 11: We should at least comment the parameters?

luyzdeleon: Also `SessionDeveloper` should be renamed to `SessionApplication`.

tests/bdd/session/session_test.go
```diff
@@ -1,264 +1,82 @@
...
-				Context("No nodes are associated with a blockchain", func() {
-					noNodesSeed := legacy.SessionSeed{DevID: devid, BlockHash: blockhash, RequestedChain: legacy.SHA3FromString("foo"), NodeList: nodelist, Capacity: capacity}
-					It("should return `invalid blockchain` error", func() {
-						_, err := legacy.NewSession(noNodesSeed)
-						Expect(err).To(Equal(legacy.InsufficientNodesError))
-					})
+			Context("All seed data is valid", func() {
+				It("should return nil error", func() {
+					_, err := session.NewSessionKey(validDeveloper, validNonNativeChain, validBlockID)
+					gomega.Expect(err).To(gomega.BeNil())
```
luyzdeleon, line 44: Shouldn't this also test that the new session key is not nil?

tests/bdd/session/session_test.go
```diff
@@ -1,264 +1,82 @@
...
+				It("should return nil error", func() {
+					_, err := session.NewSessionNodes(validNonNativeChain, validSessionKey)
+					gomega.Expect(err).To(gomega.BeNil())
 				})
 			})
 		})
+		Context("Valid seed data for the session", func() {
+			It("should return nil error", func() {
+				_, err := session.NewSession(validDeveloper, validNonNativeChain, validBlockID)
+				gomega.Expect(err).To(gomega.BeNil())
```
luyzdeleon, line 80: Shouldn't this also test that the returned session is not nil?

tests/bdd/session/session_test.go
```diff
@@ -1,264 +1,82 @@
...
+			})
+		})
 	})
+	// abide by the session 'rules'
+	PDescribe("Rules", func() {})
+	// gracefully terminate the session
+	PDescribe("Termination", func() {})
```
luyzdeleon, line 90: I don't think Sessions should ever be "terminated".

x/session/sessionNodes.go
```diff
@@ -0,0 +1,141 @@
...
+	// return NodeDistance array to be ordered
+	xorResult, err = xor(result, sessionKey)
+	if err != nil {
+		return nil, err
+	}
+	// sort the nodes based off of distance
+	result = revSort(xorResult)
+
+	// return the top 5 nodes
+	return result[0:5], nil
```
luyzdeleon: Won't this error out if there are less than 5 nodes? Also the `5` should be made a constant in case it needs to be referenced from different places.

x/session/sessionNodes.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if len(result) == 0 {
+		return nil, InsufficientNodesError
+	}
+	return result, nil
+}
+
+// xor the sessionNodes.publicKey against the sessionKey to find the computationally
+// closest nodes
+func xor(sessionNodes SessionNodes, sessionkey SessionKey) (NodeDistances, error) {
```
luyzdeleon, line 86: Shouldn't it be better for randomness to xor the address? Since it's a hash of the public key?

x/session/sessionNodes.go
```diff
@@ -0,0 +1,141 @@
...
+
+// returns the length of the node pool -> needed for sort.Sort() interface
+func (n NodeDistances) Len() int { return len(n) }
+
+// swaps two elements in the node pool -> needed for sort.Sort() interface
+func (n NodeDistances) Swap(i, j int) { n[i], n[j] = n[j], n[i] }
+
+// returns if node i is less than node j by XOR value
+// it assumes big endian encoding
+func (n NodeDistances) Less(i, j int) bool {
```
luyzdeleon, line 126: Where is this being used?

tests/bdd/session/legacySession_test.go
andrewnguyen22: Yes! it's in legacy so I left it but do you think I should remove it?

tests/bdd/session/session_test.go
andrewnguyen22: Yes, great point

tests/bdd/session/session_test.go
andrewnguyen22: "termination" is what we called it in the full spec

The DOS signal that ends the 'session'

x/session/sessionNodes.go
andrewnguyen22: Yes, and it is a constant! I just didn't use it lol

x/session/sessionNodes.go
andrewnguyen22: I re-hash the public key to ensure that it's the same length as the sessionKey

x/session/sessionNodes.go
andrewnguyen22: It's part of the interface needed for sort.Sort

x/session/session.go
andrewnguyen22: Did we officially decide on the nomenclature Application? What was the reasoning behind this decision?

tests/bdd/session/session_test.go
andrewnguyen22: Yes!

tests/bdd/session/legacySession_test.go
luyzdeleon: Yeah, you're right, since it's legacy we can keep it for future reference.

x/session/session.go
luyzdeleon: The reasoning is that it's more clear to understand the terminology under the following list:

- Applications
- ApplicationClient/User
- Node

`Applications` is what is actually being served by the network.

tests/bdd/session/session_test.go
luyzdeleon: That means expiration not termination. Maybe just change the term to `Expiration`.

x/session/sessionNodes.go
luyzdeleon: Got it, that makes sense.

x/session/sessionNodes.go
luyzdeleon: Got it, a comment about that like you did in the 2 methods above would help identify that.

luyzdeleon (approved):

PR #450, opened 2019-09-20
Issue #446

### LINKED ISSUE #385: Rename Non-Native Chain Structures
Blockchain, HostedChains, and Chain struct are named broadly. Let's change that.

Please fix this relationship between all of these structures, it is hard to read. (Rename, consolidate, and comment)

### LINKED ISSUE #387: Phase out all *List and *Set structures with more standard structures.
It is quite dangerous to create your own mux list when there are already vetted implementations.

Remove custom all `list` and `set` structures and replace with standard structures.

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
Ensure BDD tests don't break with this change.

### LINKED ISSUE #390: Update Relay Batch Interface
Relay Batch needs a complete overhaul to fit the new spec

An increment counter structure needs to be included as well

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
BDD test is needed for IC and Batch.

### LINKED ISSUE #428: Implement AAT token validation
We need a standard validator that can validate different versions of the AAT token according to specification.

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
BDD Test pending for implementation

### LINKED ISSUE #444: Add hex wrapper functions to crypto module
Include hex wrapper functions in crypto module, to handle inputs and convert outputs to hex format.

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
Needs unit test!

### LINKED ISSUE #446: Dispatch legacy code
The dispatch module legacy code is not up to spec

Bring the dispatch legacy code up to the current specification

**(Optional) Include a BDD test that would satisfy the expected behaviour of the new feature**
Requires functioning BDD

andrewnguyen22:
closes #446 
closes #444 
closes #428 
closes #390 
closes #387 
closes #385


types/hostedBlockchain.go
```diff
@@ -0,0 +1,88 @@
...
+	h := hex.EncodeToString(chainHash)
+	res := (*List)(c).Get(h)
+	if res == nil {
+		return HostedBlockchain{}
+	}
+	return res.(HostedBlockchain)
+}
+
+// "HostedChainsFromFile" reads a file into the hosted chains object.
+func HostedChainsFromFile(filepath string) error {
```
luyzdeleon, line 82: I believe this function should not be responsibility of the type, rather than responsibility of the config, which should read the parameters from whichever source and feed it into the type constructor.

types/list.go
```diff
@@ -0,0 +1,51 @@
+package types
+
+import (
+	"sync"
+)
+
+type List struct {
```
luyzdeleon, line 7: Wasn't the point of #387 to phase out these structures?

types/set.go
```diff
@@ -0,0 +1,39 @@
+package types
+
+import (
+	"sync"
+)
+
+type Set struct {
```
luyzdeleon, line 7: Wasn't the point of #387 to phase out these structures?

x/blockchain/blockchain.go
```diff
@@ -0,0 +1,31 @@
+package blockchain
```
luyzdeleon, line 1: Is the objective of this module to wrap the finality storage?

x/dispatch/dispatch.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/pokt-network/pocket-core/x/blockchain"
+	"github.com/pokt-network/pocket-core/x/session"
+)
+
+func DispatchPeers() {
+	// TODO returns tendermint peers cross checked with world state
+}
+
+// essentially a wrapper for session.NewSession()
+func DispatchSession(application session.SessionAppPubKey, nonNativeChain session.SessionBlockchain) (*session.Session, error) {
```
luyzdeleon, line 13: Let's create an issue here to cache sessions so we don't have to generate them

x/session/sessionBlockchain.go
```diff
@@ -7,3 +7,11 @@ import (
 // extension/wrapper of legacy.Blockchain for session
 // TODO non native chains need to be defined by config, for now will be hash
```
luyzdeleon, line 8: Is there an issue for this?

x/session/sessionNodes.go
```diff
@@ -28,15 +29,43 @@ type NodeDistance struct {
...
-	if len(sessionKey) == 0 {
-		return nil, EmptySessionKeyError
+func (sn SessionNodes) Validate() error {
+	if len(sn) < SESSIONNODECOUNT {
+		return InsufficientNodesError
+	}
+	return nil
+}
+
+func (sn SessionNodes) Contains(n types.Node) bool { // todo use a map instead of a slice to save time
```
luyzdeleon, line 39: Create issue.

x/service/const.go
```diff
@@ -1,5 +1,23 @@
...
+
 const (
+	// The http method defaults to POST for Relays
+	// because JSON RPC uses POST for all calls
 	DEFAULTHTTPMETHOD = "POST"
+	// HTTP Service Types
+	RPC ServiceType = iota + 1
+	REST
+
+	// TODO remove and get from world state
```
luyzdeleon, line 15: This should get the app pub key from config, not the world state.

x/service/relay.go
```diff
@@ -1,69 +1,77 @@
...
 
+// a read / write API request from a hosted (non native) blockchain
 type Relay struct {
-	Blockchain       ServiceBlockchain `json:"blockchain"`
-	Payload          ServicePayload    `json:"payload"`
-	ServiceToken     ServiceToken      `json:"token"`
-	IncrementCounter IncrementCounter  `json:"incrementCounter"`
+	Blockchain         ServiceBlockchain  `json:"blockchain"`       // the non-native blockchain needed to service
+	Payload            ServicePayload     `json:"payload"`          // the data payload of the request
+	ServiceCertificate ServiceCertificate `json:"incrementCounter"` // the authentication scheme needed for work
```
luyzdeleon, line 12: Why call it `ServiceCertificate` if it's an increment counter?

x/service/relay.go
```diff
@@ -1,69 +1,77 @@
...
 	}
-	if err := r.ServiceToken.Validate(); err != nil {
-		return errors.New(InvalidTokenError.Error() + " : " + err.Error())
+	// ensure the blockchain is supported
+	if !hostedBlockchains.Contains(hex.EncodeToString(r.Blockchain)) {
+		return UnsupportedBlockchainError
 	}
-	if err := r.IncrementCounter.Validate(r.ServiceToken.AATMessage.ClientPublicKey, r.ServiceToken.Hash()); err != nil {
-		return errors.New(InvalidIncrementCounterError.Error() + " : " + err.Error())
+	// todo check to see if non native blockchain is staked for by the developer
```
luyzdeleon, line 28: Create issues.

x/service/relay.go
```diff
@@ -1,69 +1,77 @@
...
+	if err := r.Validate(hostedBlockchains); err != nil {
 		return "", err
 	}
-	req.Header.Set("Content-Type", "application/json")
-	resp, err := (&http.Client{}).Do(req)
-	if err != nil {
-		return "", err
+	// handle the relay payload based on the type
+	switch r.Payload.Type() {
+	case REST:
```
luyzdeleon: I think these 2 can be encapsulated into HTTP. We need to better define this cause either we need nodes to let Applications know that they are hosting a specific network interface (HTTP vs Websockets vs IPC, etc...).

x/service/relayBatch.go
```diff
@@ -1,17 +1,80 @@
...
+)
+
+func GetGlobalRelayBatches() *RelayBatches {
+	relayBatchOnce.Do(func() {
+		globalRelayBatches = (*RelayBatches)(types.NewList())
+	})
+	return globalRelayBatches
+}
+
+func (rbs *RelayBatches) AddEvidence(authentication ServiceCertificate) error {
```
luyzdeleon, line 29: See my comment on `ServiceCertificate` nomenclature.

x/service/servicePayload.go
```diff
@@ -12,7 +14,16 @@ type HttpServicePayload struct {
...
+type ServiceType int
+
 func (sd ServiceData) Bytes() []byte {
 	// todo
 	return []byte(sd)
 }
+
+func (sp ServicePayload) Type() ServiceType {
+	if reflect.TypeOf(sp.HttpServicePayload) == reflect.TypeOf((HttpServicePayload{})) {
+		return REST
```
luyzdeleon: See my comment regarding service types.

tests/bdd/servicing/service_test.go
```diff
@@ -1,13 +1,41 @@
...
+			Describe("Parsing a json file", func() {
+
+				Context("Able to unmarshal the blockchain list into chain objects", func() {
+
+					It("should return a nil error", func() {
+						Expect(types.HostedChainsFromFile(chainsfile)).To(BeNil())
+					})
+
+					It("should have created a globally accessible list of blockchains", func() {
+						Expect(types.GetHostedChains().Len()).ToNot(BeZero())
```
luyzdeleon, line 26: Shouldn't this test check against the fixture data to make sure the `HostedChain` instances are properly created?

luyzdeleon (changes requested):

types/hostedBlockchain.go
andrewnguyen22: Yeah I agree, but I was being blocked and needed a temporary solution while Lowell handels the config

types/list.go
andrewnguyen22: Originally yes, but I found myself building something very similar and thought that they were the best solution for now

types/set.go
andrewnguyen22: See above

x/blockchain/blockchain.go
andrewnguyen22: Could be, I just needed temporary blockchain functions

x/dispatch/dispatch.go
andrewnguyen22: You had told me that caching was the job of infra. Have you changed your position on this?

x/service/const.go
andrewnguyen22: Nope, app pub key is from the world state

andrewnguyen22: Node public key (self) is from config

x/service/relay.go
andrewnguyen22: It's not an increment counter, it contains an increment counter. The service certificate is proof of work completed on both sides. It has a lot more than just the integer

types/hostedBlockchain.go
luyzdeleon: I see, let's a TODO/Issue to move this to config.

x/dispatch/dispatch.go
luyzdeleon: You are 100% right, what we need to figure out is how the caching will happen.

x/service/relay.go
andrewnguyen22: I disagree, in the future there may be different executions based on RPC or REST

x/service/const.go
luyzdeleon: Got it, thanks for the clarification.

x/service/relay.go
luyzdeleon: How is RPC different from REST when all it does at the networking level is sending a POST request to a single endpoint?

tests/bdd/servicing/service_test.go
andrewnguyen22: I mocked hosted chains using a fixture document that includes correct inputs. Are you suggesting to test the fixture code, cause it has little to do with the servicing functionality?

luyzdeleon: No, what I mean is the design of the test itself, by literally reading the test I get the sense that I could have a list of empty structs and just because the list is full it doesn't mean the list is valid, you see what I'm saying? Behaviour wise we want tests to reflect the expected behaviour literally.

x/dispatch/dispatch.go
andrewnguyen22: Awesome, I'll create the issue

x/service/relay.go
andrewnguyen22: Good point, will consolidate

tests/bdd/servicing/service_test.go
andrewnguyen22: Done

x/service/relay.go
andrewnguyen22: Done

luyzdeleon (approved):

PR #455, opened 2019-10-01
Adds graceful exit

### LINKED ISSUE #397: CLI should shutdown/startup node in a standard
The current shutdown process is not standardized and the software is not aware of the shutdown signal incoming from docker 

A standardized shutdown/startup way in order to work with docker and other OS in a native way

aqt01:
Closes https://github.com/pokt-network/pocket-core/issues/397

### DIFF
--- a/tests/unit/cmd_test.go
+++ b/tests/unit/cmd_test.go
@@ -10,14 +10,30 @@ import (
 	"time"
 )
 
+func TestBuildClient(t *testing.T) {
+	// Build the pocket_core binary in workdir
+	build := exec.Command("go", "build", "../../cmd/pocket_core")
+	err := build.Run()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


tests/unit/cmd_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"bufio"
+	"fmt"
+	"os/exec"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestBuildClient(t *testing.T) {
```
luyzdeleon, line 13: This shouldn't just be a test, but also a step in the CI config.

luyzdeleon (changes requested):

luyzdeleon (approved):

PR #462, opened 2019-10-11
Adds logrotation and logSize, logAge, logBackups, logCompress flags

### LINKED ISSUE #415: Provide configurable log rotation
Log rotation is not configurable via flags via CLI

Provide flags for rotation based on time or size of the log file, or a combination of the both

aqt01:
Closes https://github.com/pokt-network/pocket-core/issues/415

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -548,14 +548,6 @@
   version = "v1.4.7"
 
 [[projects]]
-  digest = "1:80c34337e8a734e190f2d1b716cae774cca74db98315166f92074434e9af0227"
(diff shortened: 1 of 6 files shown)


andrewnguyen22:
So I noticed you introduced `lumberjack "gopkg.in/natefinch/lumberjack.v2"` dependency. 
Before I knew it would have another dep, I was pro application log rotation. 

So, with that being said... Do you think it would be better to allow infra to handle log rotation?

andrewnguyen22 (approved):

PR #471, opened 2019-10-25
Feature/go modules

edjroz:
- Use multi stage docker builds to reduce image size from GB to MB
- always have `go.sum` on git as per the modules [docs](https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file)


andrewnguyen22:
Hey Eduardo! Thanks for catching this, great work man. I'll go ahead and merge this. Can you create an issue next time and call the feature branch the name of the issue as seen in the development guidelines?

andrewnguyen22 (approved):

Issue #696, opened 2020-03-02
Supply endpoint not working in RPC


andrewnguyen22:
Returns empty response

andrewnguyen22:
closed with #710

Issue #503, opened 2020-01-04
RPC Documentation


luyzdeleon:
**Describe the solution you'd like**
We need either a Swagger spec or a simple Markdown document documenting the Pocket Core RPC API.

**Additional context**
Please see here for all the routes and handler functions for these API endpoints: https://github.com/pokt-network/pocket-core/blob/staging/app/cmd/rpc/server.go#L48

PR #697, opened 2020-03-03
Issue #691

### LINKED ISSUE #691: Change camelCase to snake_case on RPC responses

oten91:
closes #691


doc/rpc-spec.md
```diff
@@ -22,9 +22,109 @@ ALl RPC calls in Pocket Core will be HTTP `POST`
 ### Client Namespace
 The default namespace contains functions that are pertinent to the execution of the Pocket Node.
 
-// todo
+- /v1/client/dispatch
+> sends a dispatch request to the network and get the nodes that will be servicing your requests for the session.
```
andrewnguyen22, line 26: This spec has very minimal information on how to execute .

I think we need examples, the actual structure.

See: https://github.com/ethereum/wiki/wiki/JSON-RPC#web3_clientversion

For example

andrewnguyen22: I don't see any standards being set with this doc. 
Remember that client side and future rpc development should conform to this spec. But this isn't really much of a spec

andrewnguyen22 (changes requested):

doc/rpc-spec.md
andrewnguyen22: Also, you may want to look into #696 before finishing the spec as it affects your work

andrewnguyen22: After call with @oten91 it seems as if this commit was on the wrong branch, and does not satisfy #503

Issue #693, opened 2020-03-02
Pocket Core CLI version command


luyzdeleon:
**Describe the solution you'd like**
Add a `version` command to the CLI. That outputs the current version of Pocket Core, e.g.:

`$  pocket version`
`RC-0.1.0`

andrewnguyen22:
This is tricky because it would just return the version of the CLI and not the App running on tendermint

andrewnguyen22:
Perhaps a better solution is to update the rpc endpoint to provide us with comprehensive version information
@luyzdeleon

luyzdeleon:
I believe still this would be good enough. When dealing with installing binaries, the easier way to confirm that you have the right binary installed is to check the version. In the future if we could display more info on this command then perfect. We could make the output explicity by saying:

```
$ pocket version
CLI Version: RC-0.1.0
```

andrewnguyen22:
Alright, will likely implement both

PR #713, opened 2020-03-10
add version CLI command

edjroz:
Solves #693

### DIFF
--- a/app/cmd/cli/query.go
+++ b/app/cmd/cli/query.go
@@ -2,12 +2,13 @@ package cli
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
+
 	"github.com/pokt-network/pocket-core/app"
 	appTypes "github.com/pokt-network/pocket-core/x/apps/types"
 	nodeTypes "github.com/pokt-network/pocket-core/x/nodes/types"
 	"github.com/spf13/cobra"
-	"strconv"
-	"strings"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


app/cmd/cli/query.go
```diff
@@ -94,6 +96,14 @@ var queryHeight = &cobra.Command{
 		fmt.Printf("Block Height: %d\n", res)
 	},
 }
+var version = &cobra.Command{
```
oten91: Command is under query not root 

```
oten91@OMP % ./pocket_core version
Error: unknown command "version" for "pocket"
Run 'pocket --help' for usage.
unknown command "version" for "pocket"
oten91@OMP pocket_core % ./pocket_core query version
Version: RC-0.2.0
oten91@OMP pocket_core % 
```

is this intended?

app/cmd/rpc/server.go
```diff
@@ -2,17 +2,18 @@ package rpc
...
-	"github.com/julienschmidt/httprouter"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
-)
 
-const (
-	APIVersion = "0.2.0"
```
oten91, line 13: what happens to releases, do we change the app.Tag value to "" each time?

oten91 (changes requested):

app/cmd/cli/query.go
edjroz: It is intended, however we can move this to root if everyone agrees.

app/cmd/rpc/server.go
edjroz: We change the version as needed, and the tag on a git tag basis

oten91: what do you think about merging app.Tag and app.Version in 1 variable?

edjroz: the usage of `app.Tag` allows for fine grained behavior for whenever we release `stable` tags which should be separate of `RC`

oten91 (approved):

PR #744, opened 2020-03-30
Removing Empty Proof Object from RequestHash,

### LINKED ISSUE #742: Empty proof object in request hash
Fix here:

```
relay := Relay{
func (r Relay) RequestHash() []byte {
		Payload: r.Payload,
		Meta:    r.Meta,
	}

```

oten91:
Changed from Relay to an anonymous struct to remove the empty proof object being generated

Closes #742


edjroz (dismissed):
LGTM!

andrewnguyen22 (changes requested):
I notice that the payload bytes() function has an empty headers object, can you use omit instead?

```
func (p Payload) Bytes() []byte {
	bz, err := json.Marshal(Payload{Data: p.Data, Method: p.Method, Path: p.Path})
	if err != nil {
		panic(fmt.Sprintf("an error occured converting the payload to bytes:\n%v", err))
	}
	return bz
}

```

oten91:
> I notice that the payload bytes() function has an empty headers object, can you use omit instead?
> 
> ```
> func (p Payload) Bytes() []byte {
> 	bz, err := json.Marshal(Payload{Data: p.Data, Method: p.Method, Path: p.Path})
> 	if err != nil {
> 		panic(fmt.Sprintf("an error occured converting the payload to bytes:\n%v", err))
> 	}
> 	return bz
> }
> ```
Added omitempty!

oten91 (commented):
@andrewnguyen22 Changes made

andrewnguyen22 (approved):

PR #766, opened 2020-04-09
Issue #754

### LINKED ISSUE #754: New transaction query endpoints
In order for the participants of the network to be able to query transactions in a certain block height and by a particular sender, 2 new endpoints must be created:

- `/v1/query/blocktxs`
- Request body example:
```json
{"height": "1"}
```

- `/v1/query/accounttxs`
- Request body example:
```json
{"address": "abcd..."}
```

luyzdeleon:
Closes #754

### DIFF
--- a/app/cmd/cli/query.go
+++ b/app/cmd/cli/query.go
@@ -87,8 +87,8 @@ var queryTx = &cobra.Command{
 }
 
 func validatePagePerPageArgs(args []string) (page int, perPage int) {
-	page := 0
-	perPage := 0
+	page = 0
+	perPage = 0
 	if len(args) == 2 {
 		parsedPage, err := strconv.Atoi(args[1])
 		if err != nil {
--- a/app/cmd/cli/query.go
+++ b/app/cmd/cli/query.go
@@ -86,6 +86,24 @@ var queryTx = &cobra.Command{
 	},
 }
 
+func validatePagePerPageArgs(args []string) (page int, perPage int) {
+	page := 0
+	perPage := 0
+	if len(args) == 2 {
+		parsedPage, err := strconv.Atoi(args[1])
+		if err != nil {
+			page = parsedPage
+		}
+	}
+	if len(args) == 3 {
+		parsedPerPage, err := strconv.Atoi(args[2])
+		if err != nil {
+			perPage = parsedPerPage
+		}
+	}
+	return page, perPage
+}
+
 var queryAccountTxs = &cobra.Command{
 	Use:   "account-txs <address> <page> <per_page>",
 	Short: "Get the transactions sent by the address, paginated by page and per_page",
@@ -93,20 +111,7 @@ var queryAccountTxs = &cobra.Command{
 	Long:  `Returns the transactions sent by the address`,
 	Run: func(cmd *cobra.Command, args []string) {
 		app.SetTMNode(tmNode)
-		page := 1
-		perPage := 30
-		if len(args) == 2 {
-			parsedPage, err := strconv.ParseInt(args[1], 10, 64)
-			if err != nil {
-				page = int(parsedPage)
-			}
-		}
-		if len(args) == 3 {
-			parsedPerPage, err := strconv.ParseInt(args[2], 10, 64)
-			if err != nil {
-				perPage = int(parsedPerPage)
-			}
-		}
+		page, perPage := validatePagePerPageArgs(args)
 		res, err := app.QueryAccountTxs(args[0], page, perPage)
 		if err != nil {
 			fmt.Println(err)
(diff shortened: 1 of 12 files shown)


app/cmd/cli/query.go
```diff
@@ -84,6 +86,71 @@ var queryTx = &cobra.Command{
...
+	Use:   "account-txs <address> <page> <per_page>",
+	Short: "Get the transactions sent by the address, paginated by page and per_page",
+	Args:  cobra.RangeArgs(1, 3),
+	Long:  `Returns the transactions sent by the address`,
+	Run: func(cmd *cobra.Command, args []string) {
+		app.SetTMNode(tmNode)
+		page := 1
+		perPage := 30
+		if len(args) == 2 {
+			parsedPage, err := strconv.ParseInt(args[1], 10, 64)
```
andrewnguyen22: Use strconv.Atoi() instead, no need to convert to int64

andrewnguyen22: Do this for all parseInt()

app/cmd/cli/query.go
```diff
@@ -84,6 +86,71 @@ var queryTx = &cobra.Command{
...
 }
 
+var queryAccountTxs = &cobra.Command{
+	Use:   "account-txs <address> <page> <per_page>",
+	Short: "Get the transactions sent by the address, paginated by page and per_page",
+	Args:  cobra.RangeArgs(1, 3),
+	Long:  `Returns the transactions sent by the address`,
+	Run: func(cmd *cobra.Command, args []string) {
+		app.SetTMNode(tmNode)
+		page := 1
```
andrewnguyen22: These can be zero, because defaults are in query

x/nodes/query.go
```diff
@@ -203,6 +208,42 @@ func QueryTransaction(tmNode rpcclient.Client, hash string) (*ctypes.ResultTx, e
 	return tx, nil
 }
 
+func validatePageAndPerPage(page, perPage int) (resPage, resPerPage int) {
+	resPage = defaultPage
+	resPerPage = defaultPerPage
+	if page > 0 {
```
andrewnguyen22, line 217: These conditionals do not seem right

x/nodes/query.go
```diff
@@ -203,6 +208,42 @@ func QueryTransaction(tmNode rpcclient.Client, hash string) (*ctypes.ResultTx, e
...
+	}
+	return resPage, resPerPage
+}
+
+func QueryAccountTransactions(tmNode rpcclient.Client, addr string, page, perPage int) (*ctypes.ResultTxSearch, error) {
+	_, err := hex.DecodeString(addr)
+	if err != nil {
+		return nil, err
+	}
+	query := fmt.Sprintf("message.sender='%s'", addr)
```
andrewnguyen22: Make this a constant

x/nodes/query.go
```diff
@@ -203,6 +208,42 @@ func QueryTransaction(tmNode rpcclient.Client, hash string) (*ctypes.ResultTx, e
...
+	page, perPage = validatePageAndPerPage(page, perPage)
+	result, err := tmNode.TxSearch(query, false, page, perPage)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func QueryBlockTransactions(tmNode rpcclient.Client, height int64, page, perPage int) (*ctypes.ResultTxSearch, error) {
+	query := fmt.Sprintf("tx.height=%d", height)
```
andrewnguyen22: Make this a constant

andrewnguyen22 (changes requested):

app/cmd/cli/query.go
luyzdeleon: You're right for all, except the block height which needs to be an int64.

app/cmd/cli/query.go
luyzdeleon: Correct!

x/nodes/query.go
luyzdeleon: The argument cannot be 0, as it causes the endpoint to error out. So as a nicety if you pass 0 it will use defaults.

x/nodes/query.go
luyzdeleon: Applied not only to these, but also to other queries in the same file.

x/nodes/query.go
luyzdeleon: Applied not only to these, but also to other queries in the same file.

andrewnguyen22:
```
# github.com/pokt-network/pocket-core/app/cmd/cli
app/cmd/cli/query.go:90:7: no new variables on left side of :=
app/cmd/cli/query.go:91:10: no new variables on left side of :=
```

andrewnguyen22 (approved):

PR #792, opened 2020-04-21
Fix for reset cmd on new configuration model

### LINKED ISSUE #789: Update Reset Command for new config
on _latest_ the reset command is not working

oten91:
closes #789

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -681,6 +681,7 @@ func createDummyACL(kp crypto.PublicKey) govTypes.ACL {
 // XXX: this is totally unsafe.
 // it's only suitable for testnets.
 func ResetWorldState(cmd *cobra.Command, args []string) {
+	loadGlobaConfigForReset(cmd)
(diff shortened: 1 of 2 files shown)


app/config.go
```diff
@@ -681,14 +681,30 @@ func createDummyACL(kp crypto.PublicKey) govTypes.ACL {
...
+		if err != nil {
+			panic("could not get home directory for data dir creation: " + err.Error())
+		}
+		datadir = home + FS + DefaultDDName
+	} else {
+		datadir = cmd.Flag("datadir").Value.String()
+	}
+	c := DefaultConfig(datadir)
+	GlobalConfig = c
+}
```
edjroz: This looks very specific to the reset CLI, can we just put this inside the Reset WorldState ?

edjroz (dismissed):

app/config.go
oten91: sure

edjroz (approved):

PR #926, opened 2020-05-07
Fix Inclusive >= for maxRelays value

### LINKED ISSUE #881: Inclusive >= for maxRelays value
**Describe the bug**

should be >= instead of >
```
// if the total relays is greater than the resulting integer, this is the pseudorandom chosen proof
if totalRelays > maxValue {
```
pocketcore/keeper/proof.go

oten91:
Closes #881


edjroz (approved):

edjroz (changes requested):
Changelog should reflect a change on the pseudorandom algorithm

PR #928, opened 2020-05-07
Swapped ExecuteProof and SetReceipt

### LINKED ISSUE #879: Swap ExecuteProof and SetReceipt
**Describe the bug**
The proper business logic is to `execute` then set the `receipt`, not the other way around! 

```
// set the claim in the world state
	er := k.SetReceipt(ctx, addr, types.Receipt{
		SessionHeader:   claim.SessionHeader,
		Total:           claim.TotalProofs,
		ServicerAddress: addr.String(),
		EvidenceType:    proof.Leaf.EvidenceType(),
	})
	if er != nil {
		return sdk.ErrInternal(er.Error()).Result()
	}
	// valid claim message so execute according to type
	err = k.ExecuteProof(ctx, proof, claim)
	if err != nil {
		return err.Result()
	}
```
x/pocketcore/handler.go

oten91:
closes #879

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -22,7 +22,8 @@
 - Catch error writing json for RPC endpoints
 - log mint errors
 - Changed SendClaimTx to use sessionContext for supportedBlockchains
-- Refactor ABCI BeginBlock for app module, avoid looping twice on mature applications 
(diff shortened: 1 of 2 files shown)


edjroz (dismissed):
Changelog should be updated

edjroz (dismissed):

edjroz (approved):

PR #948, opened 2020-05-18
Removing Perpetual Jailed Validators

### LINKED ISSUE #891: Remove perpetually jailed nodes
Currently, if nodes don't unjail they will exist in the world state forever with their coins locked up. We should burn after x number of blocks. Parameterize x as MaxJailedBlocks. And remove them from the store

oten91:
Closes #891


edjroz (dismissed):
LGTM

PR #972, opened 2020-05-25
Faster relay times

andrewnguyen22:
Flush to db, specific encoding for bloom filter


app/config.go
```diff
@@ -676,6 +680,22 @@ func newDefaultGenesisState() []byte {
 	})
 	res := Codec().MustMarshalJSON(accountGenesis)
 	defaultGenesis[auth.ModuleName] = res
+	// set address as application too
+	rawApps := defaultGenesis[appsTypes.ModuleName]
+	var appsGenesis appsTypes.GenesisState
+	types.ModuleCdc.MustUnmarshalJSON(rawApps, &appsGenesis)
+	appsGenesis.Applications = append(appsGenesis.Applications, appsTypes.Application{
```
luyzdeleon, line 687: Should this be in the default genesis?

x/pocketcore/types/codec.go
```diff
@@ -23,6 +24,7 @@ func RegisterCodec(cdc *codec.Codec) {
 	cdc.RegisterInterface((*Proof)(nil), nil)
 	cdc.RegisterConcrete(RelayProof{}, "pocketcore/relay_proof", nil)
 	cdc.RegisterConcrete(ChallengeProofInvalidData{}, "pocketcore/challenge_proof_invalid_data", nil)
+	cdc.RegisterConcrete(evidence{}, "pocketcore/evidence_persisted", nil)
```
luyzdeleon, line 27: Is this a private type?

luyzdeleon (changes requested):

x/pocketcore/types/codec.go
andrewnguyen22: Yes

app/config.go
andrewnguyen22: I think so, people would likely want to do exactly what I did to test for relays

luyzdeleon (approved):

PR #1058, opened 2020-07-26
http profile memory

edjroz:
expose pprof endpoints to get retrieve cpu & memory data

enable & disable with `--profileApp <true|false>`


edjroz:
This branch is made with testing purposes only. **Do not** merge!

PR #1067, opened 2020-08-05
Added Configurable Timeouts to HTTP Server/Client

### LINKED ISSUE #1066: Add custom timeout to Http server / Http Client
We should add custom timeouts to the http server on RPC and the client to better manage server resources, enhance performance in case of heavy load. This timeout should be configurable under config.json and have a default value to make it seamless with existing configs.

oten91:
Adding rpc_timeout Parameter under config.json. 

It should be an integer between 1 and 1,000,000 denoting the amount of ms (milliseconds) for request to wait before timeout for the RPC layer and http client. 

If the number is lower or higher than the value specified we use a default value of 3000 ms which should be enough for most use cases.

Closes #1066

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	AppVersion = "RC-0.5.0"
+	AppVersion = "RC-0.5.0-T"
 )
 
 // NewPocketCoreApp is a constructor function for PocketCoreApp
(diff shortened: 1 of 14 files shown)


oten91:
Tests done with the load tool have shown more consistent relay times when a reasonable timeout is used on a resource limited environment as the system is able to fail fast if it was not able to give a response to a request on the time frame defined by the timeout. Using a large timeout gives the usual behavior that mostly depends on the systems resources, with the slight difference that max value allowed for the param is smaller than the default Go value which is quite large.

andrewnguyen22 (approved):

edjroz (approved):
LGTM

Issue #1082, opened 2020-10-01
Add context relevant info to error messages in  pocket-core/app/cmd/cli/


BenVanGithub:
**Is your feature request related to a problem? Please describe.**
Error message return values are a bit cryptic and lack context which is obvious in the code, but invisible to the user.

Example 1.)
command:    pocket query node-params a
returns:        strconv.Atoi: parsing "a": invalid syntax

But should return something like:  Invalid Height: "a"  (followed by the more detailed error info)

Example 2.) 
command:    pocket query node-params 100000000
returns:  http://localhost:8082/v1/query/nodeparams
the http status code was not okay: 400, and the status was: 400 Bad Request, with a response of {"code":400,"message":"failed to get Store: no data"}

But should return something like  Invalid Height: "10000000"  (followed by the more detailed error info)

I propose to go through each command in:
pocket-core/app/cmd/cli/accounts.go
pocket-core/app/cmd/cli/app.go
pocket-core/app/cmd/cli/gov.go
pocket-core/app/cmd/cli/node.go
pocket-core/app/cmd/cli/query.go
test each command against invalid arguments and augment the existing error returns with context info:

For example: this code inside pocket query balance:
 {
			var err error
			height, err = strconv.Atoi(args[1])
			if err != nil {
				**fmt.Println(err)**
				return
			}
		}
Would be modified to read:
 {
			var err error
			height, err = strconv.Atoi(args[1])
			if err != nil {
				**fmt.Println("Invalid Height: ",args[1],err)**
				return
			}
		}

andrewnguyen22:
Good call out @BenVanGithub would you mind tackling this issue?

BenVanGithub:
I’d be happy to.  Please assign it to me.

PR #1086, opened 2020-10-03
Patch 3

BenVanGithub:


BenVanGithub:
Added context info to every command where appropriate,  
Tested every command by hand when done.
=======  4 NOTES FOR REVIEW PLEASE ===
1.)  I built a little struct to unmarshal height (see line 17) because I'm still having difficulty finding the correct structures in the repository.  It is only used one time.  Please replace it, or point me to where the proper struct is.
2.)  Modified the query block command to return the current block if no height is provided (lines 63-71)
3.)  Also in the query block command.. there was a work around for a typecasting problem which I changed to use strconv.ParseInt(args[0],10,64).. It just feels cleaner (Line 74)
4.)  Opened a separate issue about the command query daoOwner (it doesn't work)

This PR partially addresses #1082 but does not close it.

andrewnguyen22:
Hey @BenVanGithub 

Thanks for this.

Before merging:

1) [Here's the height param structure you were asking to replace](https://github.com/pokt-network/pocket-core/blob/5a26ae0f2fc69b523581e1e3ccb30b04d3ad062c/app/cmd/rpc/query.go#L23)
2) Random `res = res` on L74
3) We want to avoid the CLI doing two requests in one. We should modify the server code where if height == 0 use latest height instead of doing two requests:
```
res, err := QueryRPC(GetHeightPath, []byte{})
			if err != nil {
				fmt.Println(err)
				return
			}
			json.Unmarshal([]byte(res), &currentHeight)
			height = currentHeight.Height
```

BenVanGithub:
>     3. We want to avoid the CLI doing two requests in one. We should modify the server code where if height == 0 use latest height instead of doing two requests:
> 
Totally agree.  It was kinda hacky to include that change on the same PR with the rest of that anyway... I'm still getting the hang of github... have a few questions for the Thursday meeining.

andrewnguyen22:
Awesome, I'll keep an eye on this for edits

andrewnguyen22:
@BenVanGithub will resubmit this PR

Issue #1102, opened 2020-10-07
FAQ: What's the correct way to shut down Pocket Core and avoid Data/WAL corruption with shutdown?


andrewnguyen22:
### PURPOSE
This issue is meant to serve as a guide to how to avoid Data/WAL corruption with shutdown.
### GUIDELINES
- If FAQ does not comprehensively cover the issue, community comments and contributions are welcome
- Although the issue is closed, FAQ issues are meant to be maintained and edited over time

## What's the correct way to shut down Pocket Core and avoid Data/WAL corruption with shutdown?
Answer the question as comprehensively as possible using bulleted lists format:
- The best way to shutdown your node is to run: `kill -SIGTERM <PID>` 

### NOTES
Anecdotal reports of shutdown corruption could signal possible shutdown issues with data corruption, especially during periods of syncing

AusIV:
I'll note that I've observed this in our syncing process, though I had come to believe that I was not shutting down properly.

We have a periodic backup process that starts up, syncs up to current, then sends a SIGHUP to the process to start a graceful shutdown, then backs up the ~/.pocket/data folder. We recently realized that SIGHUP was not triggering a shutdown, and switched to SIGTERM and explicitly started waiting for the shutdown. We haven't encountered this since, but it also hasn't been long enough to be sure.

andrewnguyen22:
Thanks @AusIV for the information! This is helpful

andrewnguyen22:
Internal PNI infrastructure Dev Ops Engineers reported:
`we never experience that by resetting or providing more resources to the pocket validators` ( using proper SIGTERM )
`we have the graceful termination period for our pods`

as @AusIV indicated ^

It's possible this issue is related to improper signalling through default K8s built in solution shuttown

PR #1119, opened 2020-10-12
Client nodes should validate transactions before submitting

### LINKED ISSUE #1094: Claim Resulted In an Invalid Merkle Proof: Code 66
**Describe the bug**
I see too many code 66 claim failures to believe there is no bug in the merkle validation/generation system

See mainnet block `6666` as a case study. 

Investigate 
```
[{\"msg_index\":0,\"success\":false,\"log\":\"{\\\"codespace\\\":\\\"pocketcore\\\",\\\"code\\\":66,
\\\"message\\\":\\\"claim resulted in an invalid merkle Proof\\\"}\",\"events\":[{\"type\":\"message\",
\"attributes\":[{\"key\":\"action\",\"value\":\"proof\"}]}]},{\"msg_index\":0,\"success\":true,\"log\":\"
```

AusIV:
Out of the last 100 blocks, about 80% of the proofs submitted are generating code 66, invalid merkle proof. While I think invalid proofs are relatively uncommon, valid proofs are never resubmitted after they confirm, while invalid proofs are resubmitted each block. This is increasing the number of transactions nodes must validate considerably, and raising the CPU and RAM requirements for block validation.

This commit doesn't solve the underlying problem - I haven't figured out why invalid proofs are being generated in the first place. It simply has the client check its own proof before submitting, so they don't have to pay the fee for an invalid transaction, and the rest of the network doesn't have to process the invalid proof. If widely adopted, I believe this will significantly reduce the CPU requirements for running a node.

This doesn't fully resolve #1094, but it should reduce the impact on the network, so I would advocate adopting it quickly.

### DIFF
--- a/x/pocketcore/keeper/proof.go
+++ b/x/pocketcore/keeper/proof.go
@@ -51,6 +51,10 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, proofTx func(cliCtx ut
 		}
 		// get the merkle proof object for the pseudorandom index
 		mProof, leaf := evidence.GenerateMerkleProof(int(index))
+		if !mProof.Validate(claim.MerkleRoot, leaf, claim.TotalProofs) {
+			ctx.Logger().Error(fmt.Sprintf("produced invalid proof for pending claim for app: %s, at sessionHeight: %d", claim.ApplicationPubKey, claim.SessionBlockHeight))
+			continue
+		}
 		// generate the auto txbuilder and clictx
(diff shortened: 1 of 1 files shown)


AusIV:
I can't see how, but this is causing a consensus issue on the node I'm testing it on. I'll continue working on it, but don't merge as is.

AusIV:
Strike that, I just happened to run into #1102 as I was testing this. Starting with a known good backup this works fine. Reopening.

andrewnguyen22:
Partially fixes #1129 

Rest will be fixed in https://github.com/pokt-network/pocket-core/projects/36

Thank you for the contribution!

PR #1140, opened 2020-10-14
Added stop command.

rbrogan-git:
Just a quick add for symmetry.  You have a start why not a stop?  Not sure if you want to add some error handling or more output but I thought this was a quick fix.

### DIFF
--- a/app/cmd/cli/root.go
+++ b/app/cmd/cli/root.go
@@ -112,6 +112,14 @@ var resetCmd = &cobra.Command{
 	Run:   app.ResetWorldState,
 }
 
+// stopCmd represents the stop command
+var stopCmd = &cobra.Command{
+	Use:   "stop",
+	Short: "Stop pocket-core",
+	Long:  `Stop the Pocket node daemon`,
+	Run:   app.ShutdownPocketCore,
+}
+
 var version = &cobra.Command{
 	Use:   "version",
 	Short: "Get current version",
--- a/app/cmd/cli/root.go
+++ b/app/cmd/cli/root.go
@@ -22,6 +22,7 @@ var (
 	keybase         bool
 	mainnet         bool
 	testnet         bool
+	signalChannel   chan os.Signal
 )
 
 var CLIVersion = app.AppVersion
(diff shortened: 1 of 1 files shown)


andrewnguyen22:
Interesting idea @rbrogan-git 

I like the idea and would merge this.

The problem is that the function: 
```
func ShutdownPocketCore() {
	types.FlushSessionCache()
	types.StopServiceMetrics()
}
```
Only does those two things ^ and does not actually kill pocket-core or the Tendermint process and all of its threads.

I would try adding shutdown signals like:
```
// trap kill signals (2,3,15,9)
		signalChannel := make(chan os.Signal, 1)
		signal.Notify(signalChannel,
			syscall.SIGTERM,
			syscall.SIGINT,
			syscall.SIGQUIT,
			os.Kill, //nolint
			os.Interrupt)

		defer func() {
			sig := <-signalChannel
			app.ShutdownPocketCore()
			err := tmNode.Stop()
			if err != nil {
				fmt.Println(err)
				return
			}
			message := fmt.Sprintf("Exit signal %s received\n", sig)
			fmt.Println(message)
			os.Exit(3)
		}()
```

Youll also need to access the tendermint instance to call the stop operation

rbrogan-git:
Ah, knew it was too simple to be true!  I saw that code in the startCmd and thought it might be needed.  I would not want to duplicate it.  Could we refactor lines 92-103 and extract that method so we could call that do do all the work from either the signal capture or the stop command?

andrewnguyen22:
Yep totally, upon second inspection you'll just need trigger any of these signals:

```
			syscall.SIGTERM,
			syscall.SIGINT,
			syscall.SIGQUIT,
			os.Kill,
			os.Interrupt)
```
Because the channel is already listening for those and will notify the process to stop

rbrogan-git:
What about something like this?
```
cmd := exec.Command("kill", "-SIGTERM", os.Getpid()) // not sure this will get the correct pid
err := cmd.Start()
```
Thoughts?

rbrogan-git:
I just raised the scope of the channel and called the os.Interrupt.  How about this?

It is only 11 new lines of code.

andrewnguyen22:
Hey @rbrogan-git a few issues:

1) The stop command isn't attached anywhere so it's just a floating variable and won't be accessed via the CLI, please test the code before making a PR as seen in the [contribution guidelines](contributing guidelines)
2) The notify method isn't actually going to signal os.Interrupt, you'll need to do a syscall
3) Upon second thought, I think this command isn't useful because pocket start runs a daemon so this command would essentially be doing ctrl-c but in another window (cause you can't enter commands once the daemon is started)

With all that said I am going to close this

I appreciate the contribution and look forward to others in the future.

PR #1141, opened 2020-10-14
Only load block header for prevCtx calls

### LINKED ISSUE #1122: Load ONLY the header for prevCtx and not the entire block
PrevCtx() call loads the entirety of the block into memory. Let's just load the header instead

andrewnguyen22:
closes #1122


oten91 (approved):

edjroz (approved):
Great Contribution !!

PR #1146, opened 2020-10-15
Cache PrevCtx using Global Cache Storage

### LINKED ISSUE #1125: Cache Ctx for PrevCtx Call
`Small` LRU Cache for Contexts will greatly help performance in servicing, dispatching, and block processing

edjroz:
Close #1125


types/context.go
```diff
@@ -142,7 +152,23 @@ func (c Context) MustGetPrevCtx(height int64) Context {
...
+}
+func (c Context) addToCache(key string, i interface{}) (evicted bool) {
+	if c.CachedStore == nil {
+		return false
+	}
+	return c.CachedStore.Add(key, i)
+}
+
 func (c Context) PrevCtx(height int64) (Context, error) {
+	if cachedCtx, ok := c.getFromCache(fmt.Sprintf("PrevCtx%v", height)); ok {
```
andrewnguyen22: Since we are trying to shave every bit off of memory usage, let's remove `PrevCtx` from the key string and just use the height... Could even use bytes if that's better

types/context.go
```diff
@@ -190,7 +216,9 @@ func (c Context) PrevCtx(height int64) (Context, error) {
 		EvidenceHash:       meta.Header.EvidenceHash,
 		ProposerAddress:    meta.Header.ProposerAddress,
 	}
-	return NewContext((*ms).(MultiStore), header, false, c.logger).WithAppVersion(c.appVersion).WithBlockStore(c.blockstore).WithConsensusParams(c.consParams), nil
+	newCtx := NewContext((*ms).(MultiStore), header, false, c.logger).WithAppVersion(c.appVersion).WithBlockStore(c.blockstore).WithConsensusParams(c.consParams)
```
andrewnguyen22, line 221: Hmmm this is interesting... Probably don't need to, but wouldn't hurt to add the pointer to the cache in the prevCtx

NewContext((*ms).(MultiStore), header, false, c.logger).WithAppVersion(c.appVersion).WithBlockStore(c.blockstore).WithConsensusParams(c.consParams).WithCache(ctx.Cache)

types/lru.go
```diff
@@ -0,0 +1,120 @@
+package types
+
+import (
+	"github.com/hashicorp/golang-lru/simplelru"
+)
+
+var CacheStore *Cache = NewCache(20)
```
andrewnguyen22, line 7: Need configuration for this size

app/pocket.go
```diff
@@ -56,11 +58,16 @@ func NewPocketBaseApp(logger log.Logger, db db.DB, options ...func(*bam.BaseApp)
 	tkeys := sdk.NewTransientStoreKeys(nodesTypes.TStoreKey, appsTypes.TStoreKey, pocketTypes.TStoreKey, gov.TStoreKey)
 	// add params Keys too
 	// Create the application
+
+	// setup a global cache system
+	globalCache := sdk.NewCache(10)
```
andrewnguyen22: Need configuration

app/pocket.go
```diff
@@ -87,12 +94,12 @@ func (app *PocketCoreApp) InitChainerWithGenesis(ctx sdk.Ctx, req abci.RequestIn
 
 // setups all of the begin blockers for each module
 func (app *PocketCoreApp) BeginBlocker(ctx sdk.Ctx, req abci.RequestBeginBlock) abci.ResponseBeginBlock {
-	return app.mm.BeginBlock(ctx, req)
+	return app.mm.BeginBlock(ctx.WithCache(app.GlobalCacheStore), req)
```
andrewnguyen22: I'm worried about the ctx coming from other places (Queries and Transactions)

Setting .WithCache every-time seems error prone... Maybe we can add the cache anytime you create NewContext()?

andrewnguyen22 (changes requested):

types/context.go
edjroz: simple-lru only supports string keys we could just use the height as key

app/pocket.go
edjroz: resolved by pushing the cache storage to the context level in its entirety

app/pocket.go
edjroz: added a flag @ CLI level to allow node-runners to increase and decrease cache size

app/cmd/cli/root.go
```diff
@@ -78,7 +80,7 @@ var startCmd = &cobra.Command{
 		if testnet {
 			genesisType = app.TestnetGenesisType
 		}
-		tmNode := app.InitApp(datadir, tmNode, persistentPeers, seeds, remoteCLIURL, keybase, genesisType)
+		tmNode := app.InitApp(datadir, tmNode, persistentPeers, seeds, remoteCLIURL, keybase, genesisType, cacheSize)
```
andrewnguyen22: This cache-size doesn't merit a flag.

Also, this is just ctxCache... Let's be specific here with naming

types/context.go
```diff
@@ -116,10 +127,16 @@ func (c Context) BlockHeader() abci.Header {
...
 }
+func InitCtxCache(size int) {
+	GlobalCtxCache = NewCache(size)
+}
 
 // create a new context
 func NewContext(ms MultiStore, header abci.Header, isCheckTx bool, logger log.Logger) Context {
 	// https://github.com/gogo/protobuf/issues/519
+	if GlobalCtxCache == nil {
+		GlobalCtxCache = NewCache(20)
```
andrewnguyen22: Let's make this (20) a constant in the PocketConfig in config.go

types/context.go
```diff
@@ -190,7 +224,9 @@ func (c Context) PrevCtx(height int64) (Context, error) {
 		EvidenceHash:       meta.Header.EvidenceHash,
 		ProposerAddress:    meta.Header.ProposerAddress,
 	}
-	return NewContext((*ms).(MultiStore), header, false, c.logger).WithAppVersion(c.appVersion).WithBlockStore(c.blockstore).WithConsensusParams(c.consParams), nil
+	newCtx := NewContext((*ms).(MultiStore), header, false, c.logger).WithAppVersion(c.appVersion).WithBlockStore(c.blockstore).WithConsensusParams(c.consParams).WithCache(c.CachedStore)
+	_ = c.addToCache(fmt.Sprintf("%v", height), newCtx)
```
andrewnguyen22: Should use `%d`

types/lru.go
andrewnguyen22: I think this needs to be removed

types/context.go
```diff
@@ -106,6 +112,11 @@ func (c Context) IsCheckTx() bool             { return c.checkTx }
 func (c Context) MinGasPrices() DecCoins      { return c.minGasPrice }
 func (c Context) EventManager() *EventManager { return c.eventManager }
 func (c Context) AppVersion() string          { return c.appVersion }
+func (c Context) WithCache(cache *Cache) Context {
```
andrewnguyen22: I don't see any usages of this function... We should remove and remove the tests too

types/context.go
```diff
@@ -41,6 +44,7 @@ type Context struct {
 	consParams    *abci.ConsensusParams
 	eventManager  *EventManager
 	appVersion    string
+	CachedStore   *Cache
```
andrewnguyen22: Does this need public access?

app/cmd/cli/root.go
andrewnguyen22: After reviewing the entire code, I realized you never made a config option for this in the file. Please do that instead of the flag

andrewnguyen22 (changes requested):

andrewnguyen22:
One more thing. Please refactor `x/pocketcore/types/lru` and to global types to not duplicate code

types/context.go
edjroz: removed

types/lru.go
edjroz: removed

x/pocketcore/keeper/common_test.go
```diff
@@ -1025,3 +1025,20 @@ func (_m *Ctx) WithVoteInfos(voteInfo []abcitypes.VoteInfo) pocketTypes.Context
 
 	return r0
 }
+
+// WithCache provides a mock function with given fields: cache
```
andrewnguyen22: Should delete this function

andrewnguyen22 (changes requested):

andrewnguyen22:
@edjroz Flip these two blocks of code for safety:

```
       if cachedCtx, ok := c.getFromCache(fmt.Sprintf("%d", height)); ok {
		return cachedCtx.(Context), nil
	}
       // flip these two ^ 
	if height == c.BlockHeight() {
		header := c.BlockHeader()
		if header.LastBlockId.Hash == nil {
			header.LastBlockId.Hash = c.BlockHeader().ConsensusHash
		}
		return c.WithBlockHeader(header), nil
	}
```

andrewnguyen22 (approved):

PR #1182, opened 2020-11-11
Issue #1077

### LINKED ISSUE #1077: Use protobuf encoding instead of amino
Currently, pocket core and tendermint use amino encoding which has proven to be less than optimal.
Switch the encoding to protobuf in a backwards compatible manner

andrewnguyen22:
closes #1077


.circleci/config.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       # Install dependencies
       - run: ./deps.sh
       # static analyzer
-      - run: go vet ./...
+      # - run: go vet ./...
```
edjroz, line 23: Why remove it? unlike the linter go vet performs basic analysis of the code, which can be useful to find errors the compiler won't

oten91 (approved):

edjroz (approved):

.circleci/config.yml
andrewnguyen22: It blocks me from building if I have a MarshalJSON() function that doesn't match the standard method (like our codec does)

PR #1219, opened 2021-01-19
Fixed var=testnetGenesis

### LINKED ISSUE #1218: Wrong evidence/max_age param in testnetGenesis var
**Describe the bug**

When the user uses the flag to generate its own genesis, it was noticed on testnet that the following error was appearing:
```
Error reading genesisdoc - EvidenceParams.MaxAge must be greater than 0.
```

This was not happening on mainnet tho.

After investigating, it was found that on the apps/genesis.go the `evidence` param was wrong, as follows:
```
"evidence": {
            "max_age_num_blocks": "10000",
            "max_age_duration": "172800000000000"
        }
```

The right one should be: (as in https://github.com/pokt-network/pocket-network-genesis/blob/master/testnet/genesis.json)
```
"evidence": {
            "max_age": "120000000000"
        },
```
**To Reproduce**
1. Once having RC-0.5.2.9 running locally.
2. Do `pocket start --testnet`
3. Notice the possible error.

**Expected behavior**
The local node should start syncing normally.

emedrano2112:
Solves #1218

### DIFF
--- a/app/genesis.go
+++ b/app/genesis.go
@@ -14867,8 +14867,7 @@ var testnetGenesis = `{
             "time_iota_ms": "1"
         },
         "evidence": {
-            "max_age_num_blocks": "10000",
-            "max_age_duration": "172800000000000"
+            "max_age": "120000000000"
         },
         "validator": {
             "pub_key_types": [


edjroz (approved):
Great catch

oten91 (approved):

PR #1247, opened 2021-02-21
[Fix][Pocket-core] Wait for the child process to terminate before the parent (two processes launch mode)

eddyzags:
The aim of this PR is to gracefully shutdown the pocket core program when started with two processes. This reinforces the resilience of the program following the stop signals (SIGTERM, SIGINT,SIGQUIT,SIGKILL) sent by the kernel.

### DIFF
--- a/app/cmd/cli/root.go
+++ b/app/cmd/cli/root.go
@@ -68,16 +68,6 @@ func init() {
 	rootCmd.AddCommand(stopCmd)
 }
 
-// stopSignals defines the signals to gracefully shut down
-// the pocket core deamon (2,3,15,9)
-var stopSignals = []os.Signal{
-	syscall.SIGTERM,
-	syscall.SIGINT,
-	syscall.SIGQUIT,
-	os.Kill,
-	os.Interrupt,
-}
-
 // startCmd represents the start command
 var startCmd = &cobra.Command{
 	Use:   "start --keybase=<true or false>",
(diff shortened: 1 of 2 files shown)


app/cmd/cli/root.go
```diff
@@ -129,7 +145,7 @@ func start(cmd *cobra.Command, args []string) {
 		}
 		message := fmt.Sprintf("Exit signal %s received\n", sig)
 		fmt.Println(message)
-		os.Exit(3)
+		os.Exit(0)
```
eddyzags, line 146: When we are sending the `os.Exit(3)`, the main process interprets child process exit as an error. (This result to the cmod.Wait() returning an error line 113).
Don't think it really fits in that scenario as we are properly shutting down the service. wdyt?

app/cmd/cli/root.go
```diff
@@ -91,11 +91,32 @@ var startCmd = &cobra.Command{
 			comd.Env = env
 			comd.Stdin = os.Stdin
 			comd.Stdout = os.Stdout
-			_ = comd.Run()
+
```
oten91, line 94: Curious about this, comd.Run() seems to be the same underneath as comd.Start() + comd.Wait() .

oten91:
Hi @eddyzags thanks for your contribution!, 

I'll like to know a bit more about the scenario this tries to solve, it appears to target only the ctrl + C scenario ( which sends the SIGINT to the process group) but seems to fail for scenarios where pocket is running in background or when the signal is send directly to the child process ( the parent remains running).

andrewnguyen22:
Marking stale until we get a response from author @eddyzags 

> Hi @eddyzags thanks for your contribution!,
> 
> I'll like to know a bit more about the scenario this tries to solve, it appears to target only the ctrl + C scenario ( which sends the SIGINT to the process group) but seems to fail for scenarios where pocket is running in background or when the signal is send directly to the child process ( the parent remains running).

andrewnguyen22:
Closed due to inactivity

eddyzags:
Hey @andrewnguyen22 and @oten91 ! Sorry for the late response. I'll try to get back to you with more informations in the coming days

nymd:
Bump for @eddyzags  -- I was going to take a look at this issue but this looks like a good start. Any further news?

arnaud-skillz:
Hello @nymd, 

@eddyzags is no longer working on Pocket, i'll redirect your ping to (hopefully) someone that can give you more information. But not sure we continue the work on this topic since then.

arnaud-skillz:
To date, we forked pocket-core to have our custom overrides, so it’s OK we deal with it and implemented a workaround on our end. 
MXS answered to you @nymd in #pokt-swap i think about that

PR #1251, opened 2021-02-25
drop tags from versions

### LINKED ISSUE #1250: Split version Tags from version
Currently when an upgrade is issued

We perform a string comparassion of versions such that 
`RC-0.5.2.9 < RC-0.6.0`

Different tags might yield to erroneous comparison as seen [here](https://play.golang.org/p/vdi0PUfXZb3)

edjroz:
Close #1250

### DIFF
--- a/app/cmd/cli/gov.go
+++ b/app/cmd/cli/gov.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


andrewnguyen22:
@edjroz this is not the correct implementation. This is consensus level, it should be done at the CLI level

andrewnguyen22 (approved):

PR #1304, opened 2021-04-27
custom tx indexer

### LINKED ISSUE #1187: Add option to remove events from accounttxs and blocktxs
Remove all events from account and block txs with a parameter

### LINKED ISSUE #1191: Shift Tx Indexer to pocket core
Currently by using tendermint's `txIndexer` can lead to miss indexing of a tx as evidenced in #1188. This happens because the event does not necessarily match the signer.

A custom `txIndexer` that allows us to specify how we wanna index and store tx's

**Describe alternatives you've considered**
Modifying the tx Indexer on tendermint was considered however it would've led to breaking the ABCI standard in order to be able to pass down signer for comparassion of event & signer

### LINKED ISSUE #1231: account-txs cli command ignores the <page> parameter
The account-txs cli command seems to ignore the <page> parameter. Running RC-0.5.2.9

For example execute the following to return 10 transactions

```pocket query account-txs 42a8ab1fcdb335da4d5fd0e8ebaa4486bce12acc 1 10 false false```

Reduce the number of results to **1** and set the page parameter to **1** and note the first transaction has a hash of "3DD6721419C5A809D977DB7757B4F3B254C5BED4AAF985C62785C166D2A647F7"

```pocket query account-txs 42a8ab1fcdb335da4d5fd0e8ebaa4486bce12acc 1 1 false false```

Now set the page parameter to **2** and note the transaction returned is still "3DD6721419C5A809D977DB7757B4F3B254C5BED4AAF985C62785C166D2A647F7"

```pocket query account-txs 42a8ab1fcdb335da4d5fd0e8ebaa4486bce12acc 2 1 false false```

I would expect that with the second command as I am only returning 1 transaction at a time and asking for the 2nd page it should be returning the second transaction.

### LINKED ISSUE #1301: RPC Spec Blocktxs Missing Pagination
**Describe the bug**
Blocktxs params specification in rpcspec.yaml is missing the pagination usage and example

### LINKED ISSUE #1302: TxIndexer Pagination Broken
**Describe the bug**

> curl  --header "Content-Type: application/json" -X POST http://localhost:8081/v1/query/blocktxs --data '{"page":1,"per_page":1000,"order":"asc","height":6146}'
> 
> Even though I request 1000 records per page, the service is capping it to 100 records 
> {"total_count":100,
> 
> This is problematic as there is no pagination information in the response so I use the total_count as a guide to indicate that this is the last page. i.e. total_count < records requested means it is the last page.
> Additionally the page parameter seems to be being ignored
> curl  --header "Content-Type: application/json" -X POST http://localhost:8081/v1/query/blocktxs --data '{"page":1,"per_page":10,"order":"asc","height":6146}'
> 
> and 
> 
> curl  --header "Content-Type: application/json" -X POST http://localhost:8081/v1/query/blocktxs --data '{"page":2,"per_page":10,"order":"asc","height":6146}'
> 
> returns the same results
>

### PR DESCRIPTION
**Pull request by andrewnguyen22**
closes #1191 
closes #1187
closes #1302
closes #1301 
closes #1231

**Custom Indexer**
Since LevelDB comparators order lexicongraphically, the implementation uses ELEN to encode numbers to ensure alphanumerical
ordering at insertion time. https://www.zanopha.com/docs/elen.pdf
Since the keys are sorted alphanumerically from the start, we don't have to:
    - Load all results to memory 
    - Paginate and sort transactions after
This indexer inserts in sorted order so it can paginate and return based on the db iterator resulting in a significant 
reduction in resource consumption

The custom pocket core transaction indexer also reduces the scope of the Search() functionality to optimize strictly for 
the following use cases:
    - BlockTxs (Get transactions at a certain height)
    - AccountTxs (Get transactions for a certain account (sent and received))
    
The custom pocket core transaction indexer also injects the message_type into the struct to provide an easier method of 
parsing the transactions. `json:"message_type"`

### DIFF
--- a/types/indexer.go
+++ b/types/indexer.go
@@ -1,0 +1,292 @@
+package types
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"fmt"
+	"github.com/jordanorelli/lexnum"
+	"github.com/pkg/errors"
+	"github.com/tendermint/tendermint/libs/pubsub/query"
+	"github.com/tendermint/tendermint/state/txindex"
+	"github.com/tendermint/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+	"math"
+)
+
+var (
+	_ txindex.TxIndexer = &TransactionIndexer{}
+	e                   = lexnum.NewEncoder('=', '-')
+)
+
+const (
+	TxHeightKey         = "tx.height"
+	TxSignerKey         = "tx.signer"
+	TxRecipientKey      = "tx.recipient"
+	TxHashKey           = "tx.hash"
+	SortAscending       = "asc"
+	SortDescending      = "desc"
+	AuthCodespace       = "auth"
+	sep                 = "/"
+	maxPerPage          = 1000
+	AnteHandlerMaxError = 10
+)
+
+type TransactionIndexer struct {
+	store dbm.DB
+}
+
+func NewTransactionIndexer(store dbm.DB) *TransactionIndexer {
+	return &TransactionIndexer{store: store}
+}
+
+func (t *TransactionIndexer) AddBatch(b *txindex.Batch) error {
+	storeBatch := t.store.NewBatch()
+	defer storeBatch.Close()
+
+	for _, result := range b.Ops { // iterate through all the transaction results
+		if result.Result.Codespace == AuthCodespace && result.Result.Code < AnteHandlerMaxError {
+			continue // don't index any ante handler level errors
+		}
+		hash := result.Tx.Hash()
+
+		// index tx by sender
+		if result.Result.Signer != nil {
+			storeBatch.Set(keyForSigner(result), hash)
+		}
+
+		// index tx by height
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: types/indexer.go [hunk 1]
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/tendermint/tendermint/libs/pubsub/query"
+	"github.com/tendermint/tendermint/state/txindex"
+	"github.com/tendermint/tendermint/types"
+	dbm "github.com/tendermint/tm-db"
+	"math"
+)
+
+var (
+	_ txindex.TxIndexer = &TransactionIndexer{}
+	e                   = lexnum.NewEncoder('=', '-')
```
edjroz: I think `e` is probably not a good way to name this variable since its global.

edjroz (dismissed):
I think this issue might require an explanation of the design choices.

types/indexer.go
andrewnguyen22: Agreed, will update

PR #1354, opened 2021-10-11
Beta 0.6.3.8

iajrz:
Includes the in-memory cache implementation to reduce CPU resource usage.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -180,7 +180,7 @@ func UpdateConfig(datadir string) {
 	sdk.DefaultPocketConsensusConfig(GlobalConfig.TendermintConfig.Consensus)
 	GlobalConfig.TendermintConfig.P2P.AllowDuplicateIP = true
 	GlobalConfig.TendermintConfig.P2P.AddrBookStrict = false
-	GlobalConfig.TendermintConfig.P2P.MaxNumInboundPeers = 21
+	GlobalConfig.TendermintConfig.P2P.MaxNumInboundPeers = 14
 	GlobalConfig.TendermintConfig.P2P.MaxNumOutboundPeers = 7
 	GlobalConfig.TendermintConfig.RPC.GRPCMaxOpenConnections = 2500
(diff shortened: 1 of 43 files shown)


types/config.go
```diff
@@ -42,6 +42,7 @@ type PocketConfig struct {
 	ABCILogging              bool   `json:"abci_logging"`
 	RelayErrors              bool   `json:"show_relay_errors"`
 	DisableTxEvents          bool   `json:"disable_tx_events"`
+	Cache                    bool
```
oten91: Missing metadata for json `json:"cache"`, if not present is presented as  "Cache": false on the config.json.

app/config.go
```diff
@@ -71,9 +71,10 @@ const (
 	DefaultGenesisType
 )
 
-func InitApp(datadir, tmNode, persistentPeers, seeds, remoteCLIURL string, keybase bool, genesisType GenesisType) *node.Node {
+func InitApp(datadir, tmNode, persistentPeers, seeds, remoteCLIURL string, keybase bool, genesisType GenesisType, useCache bool) *node.Node {
 	// init config
 	InitConfig(datadir, tmNode, persistentPeers, seeds, remoteCLIURL)
+	GlobalConfig.PocketConfig.Cache = useCache
```
oten91, line 77: Move this to InitConfig to keep it working like other flags that override, currently only allows to toggle cache with the flag but is present on the config.

oten91 (changes requested):
Also please check CI some tests are failing with invalid or nil pointer references

This could be due to the test setup but let's double-check
```

--- FAIL: TestQueryRelay (3.10s)
    --- FAIL: TestQueryRelay/query_relay_amino (3.10s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xcdd793]
```

oten91 (approved):

Issue #1351, opened 2021-09-12
not informative error log


mostmojo:
i stopped the pocket start process to launch it as a bg process but i got this
 
 ```
 panic: resource temporarily unavailable

sync.(*Once).doSlow(0x1c64ff8, 0xc0000c78a8)
    /usr/lib/go-1.16/src/sync/once.go:68 +0xec
sync.(*Once).Do(...)
    /usr/lib/go-1.16/src/sync/once.go:59
github.com/pokt-network/pocket-core/x/pocketcore/types.InitConfig(0xc00013d880, 0x1537668, 0xc000544340, 0x0, 0x0, 0xc000529fa0, 0x19, 0xc00000d908, 0x15, 0xc00056f630, ...)
    /home//go/src/github.com/pokt-network/pocket-core/x/pocketcore/types/config.go:25 +0x126
github.com/pokt-network/pocket-core/app.InitPocketCoreConfig(0xc00013d880, 0x1537668, 0xc000544340)
    /home//go/src/github.com/pokt-network/pocket-core/app/config.go:374 +0x98
github.com/pokt-network/pocket-core/app.InitApp(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    /home//go/src/github.com/pokt-network/pocket-core/app/config.go:86 +0x10b
github.com/pokt-network/pocket-core/app/cmd/cli.start(0x1c24420, 0x1c64f38, 0x0, 0x0)
    /home//go/src/github.com/pokt-network/pocket-core/app/cmd/cli/root.go:96 +0x105
github.com/pokt-network/pocket-core/app/cmd/cli.glob..func53(0x1c24420, 0x1c64f38, 0x0, 0x0)
    /home//go/src/github.com/pokt-network/pocket-core/app/cmd/cli/root.go:80 +0xc5
github.com/spf13/cobra.(*Command).execute(0x1c24420, 0x1c64f38, 0x0, 0x0, 0x1c24420, 0x1c64f38)
    /root/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x1c24180, 0x407f65, 0xc00010e058, 0x401320)
    /root/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
[...]

PR #1357, opened 2021-10-14
closes 1351 with a clear error message when db can't be opened.

iajrz:


iajrz:
this would close #1351

oten91 (approved):

PR #1383, opened 2022-01-04
Explicitly cast Stdin file descriptors to int for Windows

msmania:
On Windows, Stdin needs to be casted into int as explained in https://github.com/cloudflare/gokey/pull/23.

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -4,6 +4,15 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


iajrz:
This will take a little bit of work -- running stuff in windows and checking regressions in unix like systems; will let you know how this goes. As mentioned in the other, simpler PR, this is going to take a bit but is already in the radar.

oten91 (approved):

Issue #1391, opened 2022-02-08
Cannot load a read-only chains.json file


lander2k2:
**Describe the bug**
Cannot open a chains.json file that is on a read-only filesystem.

I'm running a pocket node on Kubernetes with the `chains.json` file in a read-only volume mounted from a K8s secret since there are relay chain node passwords in there.

Error:
```
2022/02/08 21:26:24 invalid chains.json: open /root/.pocket/config/chains/chains.json: read-only file system
```

**To Reproduce**
1. Put `chains.json` on a read-only file system
2. Update `pocket_config.chains_name` in `config.json` to point to the correct location for the chains file
3. Run `pocket start`

**Expected behavior**
I expect to be able to load a read-only chains file.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - Linux

**Smartphone (please complete the following information):**
N/A

**Additional context**
It's a one-line fix in the code.  I forked and tested.  Will submit a PR.

lander2k2:
This problem also applies to the config.json file.  When mounting this to a Kubernetes pod with a configmap it fails to load for the same reason: read only filesystem.

oten91:
Merged on RC-0.8.3

PR #1392, opened 2022-02-08
Allow read-only chains.json files

lander2k2:
Issue: https://github.com/pokt-network/pocket-core/issues/1391


lander2k2:
@iajrz This PR is related but different to #1396 - this one is for `chains.json` whereas the other is for `config.json`.

iajrz:
You are, indeed, correct; perhaps this change could be tacked onto the other one, especially since they have a similar intent "allow to provide configuration from read-only filesystems"

lander2k2:
@iajrz Can do.  Will add a commit to the other PR.

PR #1393, opened 2022-02-10
SGS

oten91:
-

### DIFF
--- a/x/pocketcore/types/session.go
+++ b/x/pocketcore/types/session.go
@@ -110,16 +110,28 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
 	}
 	sessionNodes = make(SessionNodes, sessionNodesCount)
 	var node exported.ValidatorI
+	//unique address map
+	m := make(map[string]sdk.Address)
 	// only select the nodesAddrs if not jailed
 	for i, numOfNodes := 0, 0; ; i++ {
+		//if this is true we already checked all nodes we got on getValidatorsBychain
+		if len(m) >= totalNodes {
+			return nil, NewInsufficientNodesError(ModuleName)
+		}
 		// generate the random index
 		index := PseudorandomSelection(sdk.NewInt(int64(totalNodes)), sessionKey)
 		// merkleHash the session key to provide new entropy
 		sessionKey = Hash(sessionKey)
 		// get the node from the array
 		n := nodesAddrs[index.Int64()]
+		//if we already have seen this address we continue as it's either on the list or discarded
+		if _, ok := m[n.String()]; ok {
+			continue
+		}
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
+		//add the node address to the list
+		m[n.String()] = n
 		// if not found or jailed, don't add to session and continue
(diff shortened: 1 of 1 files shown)


x/pocketcore/types/session.go
```diff
@@ -110,14 +110,27 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
 	}
 	sessionNodes = make(SessionNodes, sessionNodesCount)
 	var node exported.ValidatorI
+	//unique address map
+	m := make(map[string]sdk.Address)
```
andrewnguyen22: Lets use a low memory map

`map[string]struct{}{}`

x/pocketcore/types/session.go
```diff
@@ -110,14 +110,27 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
+			return nil, NewInsufficientNodesError(ModuleName)
+		}
 		// generate the random index
 		index := PseudorandomSelection(sdk.NewInt(int64(totalNodes)), sessionKey)
 		// merkleHash the session key to provide new entropy
 		sessionKey = Hash(sessionKey)
 		// get the node from the array
 		n := nodesAddrs[index.Int64()]
+		//if we already have seen this address we continue as it's either on the list or discarded
+		if _, ok := m[n.String()]; ok {
```
andrewnguyen22, line 128: Though this DOES prevent the infinite loop...

It requires you to exhaust ALL applicable nodes within the chain set.

This is a very resource intensive operation which will require (at a minimum) looping over totalNodes

HOWEVER, since this error only happens when active service nodes is less than the minimum (24), this is an acceptable intermediate solution - as it's bounded by a constant

andrewnguyen22 (dismissed):

x/pocketcore/types/session.go
```diff
@@ -110,14 +110,27 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
 	}
 	sessionNodes = make(SessionNodes, sessionNodesCount)
 	var node exported.ValidatorI
+	//unique address map
```
Olshansk: NIT: update comment, "unique address map to avoid re-checking a pseudorandomly selected servicer"

x/pocketcore/types/session.go
```diff
@@ -110,14 +110,27 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
 		// generate the random index
 		index := PseudorandomSelection(sdk.NewInt(int64(totalNodes)), sessionKey)
 		// merkleHash the session key to provide new entropy
 		sessionKey = Hash(sessionKey)
 		// get the node from the array
 		n := nodesAddrs[index.Int64()]
+		//if we already have seen this address we continue as it's either on the list or discarded
+		if _, ok := m[n.String()]; ok {
+			continue
+		} else {
```
Olshansk: NIT (style): Technically, we don't even need the else. We can continue or add to the map in the normal code flow.

x/pocketcore/types/session.go
oten91: done

x/pocketcore/types/session.go
oten91: done

PR #1394, opened 2022-02-14
Adding Dockerfile support for pocket-core

jrynkiew:
Hi,

Adding pocket-core Dockefile as a template for users to develop further as needed


Olshansk:
Thanks for the PR! However, the go version is quite old and most of the core infrastructure is available at [pokt-network/pocket-core-deployments](https://github.com/pokt-network/pocket-core-deployments), so we'll be closing out this PR in the meantime.

PR #1395, opened 2022-02-15
Wrong parameter name for jail status in rpc-spec.yaml

msmania:


Olshansk (approved):
Thanks for the changes

PR #1396, opened 2022-02-16
Allow read-only config.json, chains.json

lander2k2:
Issue: #1391


iajrz:
Hey, just a heads up: this is in my radar; there's something to check with regards to the generation of empty `chains.json` when there are no chains -- a corner case, really; this is in my radar, though. Thanks!

iajrz:
This is the same as #1392 ; I will close that one.

lander2k2:
@iajrz commit added for read-only chains.json.  I think we're good to go now.

oten91:
Hey @lander2k2 can you check the conflict?

lander2k2:
@oten91 conflict resolved

oten91 (approved):

PR #1399, opened 2022-02-19
 Fix pocket node staking command

nodiesBlade:
no spaces in between relay_chains ids otherwise cli will read them as multiple arguments

port is required now with service_uri

true boolean is no longer required


Olshansk (approved):
Thanks for the changes!

PR #1423, opened 2022-05-03
Persistence replacement

iajrz:
This PR includes a partial implementation of the fastarchival store, as well as a remix of code to simplify changing and debugging in the path of the `start` command.

More changes are forthcoming, but this is already a good chunk that can be tested and incorporated with no ill effect to the functions of the system while keeping the maintainability wins.


oten91:
I would love to keep staging as "working and tested 100%" branch. Haven't completed the review but would prefer to have it merged to a feature, pre-release or pre-staging branch as some changes are not related to the feature this PR is named after

PR #1430, opened 2022-05-14
show staking amount warning for less than 15100 POKTs

immunity20:


iajrz (dismissed):
Nice catch; I might've decided to change the method to `LT` instead of changing the amount, but it works either way. Thanks for the contribution!

iajrz:
This will take a little bit of time to merge in due to work we're doing around the code check-in pipeline; thanks again for the contribution!

immunity20:
I changed to LT function

app/cmd/cli/stake.go
```diff
@@ -109,7 +109,7 @@ The signer may be the operator or the output address.`,
 			return
 		}
 		am := types.NewInt(int64(amount))
-		if am.LTE(types.NewInt(15100000000)) {
+		if am.LTE(types.NewInt(15099999999)) {
```
andrewnguyen22: Why LTE here but LT above?

immunity20: Initially I had .LTE 15099999999 in both places but as per iajrz comment I update LTE to LT but forgot to do the same for the second place. I have updated

oten91 (approved):

Issue #1432, opened 2022-05-24
http status code was not okay: 400, Unable to stake !


abdulgig:
Dears,

I am new here and I am trying to seek assistance. Following is the issue, I am unable get the current block. My node is syncing and the service is running,

```
root@pokt-n1:~/.pocket# pocket query block                                                                                                                   │·····················
2022/05/24 10:38:53 Initializing Pocket Datadir                                                                                                              │·····················
2022/05/24 10:38:53 datadir = /root/.pocket                                                                                                                  │·····················
http://127.0.0.1:8082/v1/query/block                                                                                                                         │·····················
the http status code was not okay: 400, and the status was: 400 Bad Request, with a response of {"code":400,"message":"height must be greater than 0, but got│·····················
 0"}              
```
Pocket is running , what is wrong then ?

```
root@pokt-n1:~/.pocket# netstat -lntpe | grep pocket                                                                                                         │·····················
tcp        0      0 127.0.0.1:26657         0.0.0.0:*               LISTEN      0          416164     56452/pocket                                           │·····················
tcp6       0      0 :::26656                :::*                    LISTEN      0          417091     56452/pocket                                           │·····················
tcp6       0      0 :::8082                 :::*                    LISTEN      0          413095     56452/pocket                                           │·····················
tcp6       0      0 :::8083                 :::*                    LISTEN      0          416146     56452/pocket 
```

Can connect also,

```
root@pokt-n1:~/.pocket# telnet localhost 8082                                                                                                                │·····················
Trying 127.0.0.1...                                                                                                                                          │·····················
Connected to localhost.                                                                                                                                      │·····················
Escape character is '^]'.                 

```

Other commands work,

```
root@pokt-n1:~/.pocket# pocket stop                                                                                                                          │·····················
2022/05/24 11:32:48 Initializing Pocket Datadir                                                                                                              │·····················
2022/05/24 11:32:48 datadir = /root/.pocket                                                                                                                  │·····················
http://127.0.0.1:8082/v1/private/stop                                                                                                                        │·····················
Post "http://127.0.0.1:8082/v1/private/stop?authtoken=bX26mH4WaIo6E3IBCU7pSsFrx": EOF   

```

abdulgig:
@Olshansk @msmania @luyzdeleon @AusIV , can anybody help please ?

msmania:
[...]

PR #1433, opened 2022-05-26
Use the current height when /query/block is invoked without height

msmania:
Fix for #1432


Olshansk:
This looks good to me but CCing @oten91 to make sure there's no potential side effects.

oten91 (approved):

oten91:
LGTM. Thank you very much for your contributions @msmania

PR #1436, opened 2022-06-06
Correct the apps command

jbroudou:
The namespace for "apps" is incorrectly documented as "app"


oten91 (approved):
Thanks For your contribution.

Issue #1448, opened 2022-07-06
[CONTRIBUTION PROPOSAL] PIP-22: Stake Weighted Servicer Rewards


Andrew-Pohl:
**Please explain your change in detail.**
This change is to enable [PIP22](https://forum.pokt.network/t/pip-22-stake-weighted-servicer-rewards/3101). PIP22 allows for vertical node stacking, it does this by scaling the rewards paid to servicers based on the amount (bin) staked by the servicer.

The changes proposed are as follows:

1. Add new DAO ACL parameters used to control PIP 22

- > **ValidatorStakeFloorMultiplier** - This corresponds to the bin widths, i.e. the size of the stake increments, or number of extra tokens required to increase the stake;

- > **ValidatorStakeWeightMultiplier** - This is the maximum multiple/divisor allowed to scale the rewards;

- > **ServicerStakeWeightCeiling** - This sets the upper limit for consolidation;

- > **ValidatorStakeFloorMultiplierExponent** - Defines the linearity of the stake-weighting curve, with 1 being linear and <1 non-linear.

2. Add two new feature Flags 

- > **RSCAL** (Reward Scaling): This flag would tell the protocol that reward scaling is active when on and to use the formulas proposed. When off, this would go back to the original formula of coins = RelaysToTokensMultiplier * Relays. The idea behind this feature flag is that if any issues arise with the mechanism overall, the whole mechanism can be disabled and the network can fallback into the already known behavior as a default in case of crises and/or unknown unknowns.

- > **VEDIT** (Validate Edit Stake): If RSCAL is true, this flag would activate the edit stake validation feature described below, meaning that edit stake transactions would be invalid if they do not cause the stake size to reach the next bin. This will give the DAO an emergency fallback to counteract edit stake spamming.

3. Modify [RewardForRelays](https://github.com/pokt-network/pocket-core/blob/8ad860a86be1cb9b891c1b6f244f3511d56a9949/x/nodes/keeper/reward.go#L11)() to read the servicers stake, floor it the nearest bin (ValidatorStakeFloorMultiplier) and weight the rewards based on the bin

```
IF RSCAL == TRUE
    // Calculate bin
    flooredStake = MIN(nodeStake - (nodeStake % ValidatorStakeFloorMultiplier), ServicerStakeWeightCeiling)

    // Calculate weight
[...]

PR #1449, opened 2022-07-11
PIP22 Stake Weighted Servicer Rewards

Andrew-Pohl:
https://github.com/pokt-network/pocket-core/issues/1448

### DIFF
--- a/x/gov/module.go
+++ b/x/gov/module.go
@@ -125,6 +125,17 @@ func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
 		//update params
 		am.keeper.SetParams(ctx, gParams)
 	}
+	
+	if am.keeper.GetCodec().IsOnNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey) {
+		params := am.keeper.GetParams(ctx)
+		//on the height we get the ACL and insert the key
+		params.ACL.SetOwner("pos/ServicerStakeFloorMultiplier", am.keeper.GetDAOOwner(ctx))
+		params.ACL.SetOwner("pos/ServicerStakeWeightMultiplier", am.keeper.GetDAOOwner(ctx))
+		params.ACL.SetOwner("pos/ServicerStakeWeightCeiling", am.keeper.GetDAOOwner(ctx))
+		params.ACL.SetOwner("pos/ServicerStakeFloorMultiplierExponent", am.keeper.GetDAOOwner(ctx))
+		//update params
+		am.keeper.SetParams(ctx, params)
+	}
 
 	u := am.keeper.GetUpgrade(ctx)
 	if ctx.AppVersion() < u.Version && ctx.BlockHeight() >= u.UpgradeHeight() && ctx.BlockHeight() != 0 {
--- a/x/nodes/keeper/reward.go
+++ b/x/nodes/keeper/reward.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"fmt"
+	"github.com/pokt-network/pocket-core/codec"
 	sdk "github.com/pokt-network/pocket-core/types"
 	govTypes "github.com/pokt-network/pocket-core/x/gov/types"
 	"github.com/pokt-network/pocket-core/x/nodes/types"
@@ -17,7 +18,32 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 			return sdk.ZeroInt()
 		}
 	}
-	coins := k.RelaysToTokensMultiplier(ctx).Mul(relays)
+
+	var coins sdk.BigInt
+
+	//check if PIP22 is enabled, if so scale the rewards
+	if k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey) {
+		//grab stake
+		validator, found := k.GetValidator(ctx, address)
+		if !found {
+			ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
+			return sdk.ZeroInt()
+		}
+
+		stake := validator.GetTokens()
+		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), (k.ServicerStakeWeightCeiling(ctx).Sub(k.ServicerStakeWeightCeiling(ctx).Mod(k.ServicerStakeFloorMultiplier(ctx)))))
+		//Convert from tokens to a BIN number
+		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
+		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
+		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), PIP_22_EXPONENT_DENOMINATOR).Quo(k.ServicerStakeWeightMultiplier(ctx))
+		coinsDecimal := k.RelaysToTokensMultiplier(ctx).ToDec().Mul(relays.ToDec()).Mul(weight)
+		//truncate back to int
+		coins = coinsDecimal.TruncateInt()
+	} else {
+		coins = k.RelaysToTokensMultiplier(ctx).Mul(relays)
+	}
+
 	toNode, toFeeCollector := k.NodeReward(ctx, coins)
 	if toNode.IsPositive() {
 		k.mint(ctx, toNode, address)
--- a/x/nodes/keeper/slash.go
+++ b/x/nodes/keeper/slash.go
@@ -8,12 +8,34 @@ import (
 	"github.com/pokt-network/pocket-core/x/nodes/types"
 	"github.com/tendermint/tendermint/crypto"
 
+	"github.com/pokt-network/pocket-core/codec"
 	sdk "github.com/pokt-network/pocket-core/types"
 )
 
 // BurnForChallenge - Tries to remove coins from account & supply for a challenged validator
 func (k Keeper) BurnForChallenge(ctx sdk.Ctx, challenges sdk.BigInt, address sdk.Address) {
-	coins := k.RelaysToTokensMultiplier(ctx).Mul(challenges)
+	var coins sdk.BigInt
+	//check if PIP22 is enabled, if so scale the rewards
+	if k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey) {
+		//grab stake
+		validator, found := k.GetValidator(ctx, address)
+		if !found {
+			ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
+			return
+		}
+
+		stake := validator.GetTokens()
+		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), (k.ServicerStakeWeightCeiling(ctx).Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx)))))
+		//Convert from tokens to a BIN number
+		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
+		//calculate the weight value
+		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), PIP_22_EXPONENT_DENOMINATOR).Quo(k.ServicerStakeWeightMultiplier(ctx))
+		coinsDecimal := k.RelaysToTokensMultiplier(ctx).ToDec().Mul(challenges.ToDec()).Mul(weight)
+		coins = coinsDecimal.TruncateInt()
+	} else {
+		coins = k.RelaysToTokensMultiplier(ctx).Mul(challenges)
+	}
 	k.simpleSlash(ctx, address, coins)
 }
 
--- a/x/nodes/keeper/valStateChanges.go
+++ b/x/nodes/keeper/valStateChanges.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pokt-network/pocket-core/crypto"
 	abci "github.com/tendermint/tendermint/abci/types"
 
+	"github.com/pokt-network/pocket-core/codec"
 	sdk "github.com/pokt-network/pocket-core/types"
 	"github.com/pokt-network/pocket-core/x/nodes/types"
 )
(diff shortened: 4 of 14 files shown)


Andrew-Pohl:
Again @luyzdeleon FYI :) Let me know if you want me to squash the commits?

x/nodes/keeper/reward.go
```diff
@@ -17,7 +18,32 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
...
+		if !found {
+			ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
+			return sdk.ZeroInt()
+		}
+
+		stake := validator.GetTokens()
+		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), (k.ServicerStakeWeightCeiling(ctx).Sub(k.ServicerStakeWeightCeiling(ctx).Mod(k.ServicerStakeFloorMultiplier(ctx)))))
+		//Convert from tokens to a BIN number
+		Bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
```
luyzdeleon: Misnamed variable, should be called `bin`

x/nodes/keeper/reward.go
```diff
@@ -17,7 +18,32 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
...
+		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
+		weight := Bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), 100).Quo(k.ServicerStakeWeightMultiplier(ctx))
```
luyzdeleon: Should the hardcoded `100` argument here be a parameter as well?

luyzdeleon: If not, it should still be a constant declared globally with enough documentation to be able to go back to why the value was hard coded in the first place.

x/nodes/keeper/slash.go
```diff
@@ -8,12 +8,34 @@ import (
...
+		if !found {
+			ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
+			return
+		}
+
+		stake := validator.GetTokens()
+		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), (k.ServicerStakeWeightCeiling(ctx).Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx)))))
+		//Convert from tokens to a BIN number
+		Bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
```
luyzdeleon: Misnamed variable, should be called `bin`.

x/nodes/keeper/slash.go
```diff
@@ -8,12 +8,34 @@ import (
...
+		//calculate the weight value
+		weight := Bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), 100).Quo(k.ServicerStakeWeightMultiplier(ctx))
```
luyzdeleon: See comment about hardcoded value here: `x/nodes/keeper/reward.go`.

x/nodes/keeper/valStateChanges.go
```diff
@@ -200,6 +201,18 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentValidator, newValidtor typ
 	if diff.IsNegative() {
 		return types.ErrMinimumEditStake(k.codespace)
 	}
+
+	if k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey) && k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.VEDITKey) {
+		//check that we are bumping up into a new bin or are above the cieling, if not return an error
+		if amount.LT(k.ServicerStakeWeightCeiling(ctx)) {
+			//grab the bin and check if we are in a new bin
+			flooredNewStake := amount.Sub(amount.Mod(k.ServicerStakeFloorMultiplier(ctx)))
+			if flooredNewStake.LTE(currentValidator.StakedTokens) {
```
luyzdeleon, line 210: I believe this should be `LT` not `LTE` as allowing "the max value" should be allowed according to the spec: https://forum.pokt.network/t/pip-22-stake-weighted-servicer-rewards/3101#validate-edit-stake-8

x/nodes/types/errors.go
```diff
@@ -95,6 +95,10 @@ func ErrMinimumEditStake(codespace sdk.CodespaceType) sdk.Error {
 	return sdk.NewError(codespace, CodeMinimumEditStake, "validator must edit stake with a stake greater than or equal to current stake")
 }
 
+func ErrSameBinEditStake(codespace sdk.CodespaceType) sdk.Error {
+	return sdk.NewError(codespace, CodeMinimumEditStake, "validator must edit stake to a greater stake Bin tha the current stake")
```
luyzdeleon: There's a typo in the error message: `Bin tha the` should be `Bin than the`

x/nodes/types/params.go
```diff
@@ -11,65 +11,77 @@ import (
...
+	DefaultMaxEvidenceAge                     = 60 * 2 * time.Second
+	DefaultSignedBlocksWindow                 = int64(100)
+	DefaultDowntimeJailDuration               = 60 * 10 * time.Second
+	DefaultSessionBlocktime                   = 25
+	DefaultProposerAllocation                 = 1
+	DefaultDAOAllocation                      = 10
+	DefaultMaxChains                          = 15
+	DefaultMaxJailedBlocks                    = 1000
+	DefaultServicerStakeFloorMultiplier int64 = 15000000000
+	DefaultServicerStakeWeightCeiling   int64 = 60000000000
```
luyzdeleon: Shouldn't the default values be equivalent to whatever makes the application behave like it does today?

luyzdeleon (changes requested):
There are very little syntax and structural changes needed for this PR, however there's one mechanism missing from the PR and it's understandable as we've never had to add a parameter before: we need code to process to actually write the params to the state. Luckily we were working in a PR as well to add another parameter, so we will be providing a reference implementation for this later today or tomorrow at the latest.

x/nodes/keeper/valStateChanges.go
Andrew-Pohl: This is caught on line 207 (if above the ceiling this code isn't hit) needs to be less than or equal to avoid bumping stake in the same bin.

x/nodes/keeper/reward.go
Andrew-Pohl: Yes this shouldn't be configurable, so declared globally. It is used to avoid overflows when finding the nth root of a decimal number. having 100 ensures that the denominator of the decimal number can never be smaller than 100th's. 

Were would be best to declare this?, personally don't think it needs global scope as it is local to both reward and slashing, so only needs to be available to nodes keeper.

x/nodes/types/params.go
Andrew-Pohl: > There are very little syntax and structural changes needed for this PR, however there's one mechanism missing from the PR and it's understandable as we've never had to add a parameter before: we need code to process to actually write the params to the state. Luckily we were working in a PR as well to add another parameter, so we will be providing a reference implementation for this later today or tomorrow at the latest.

Cool, hit me up when you have an example/ mechanism for this :)

Andrew-Pohl: Addressed in [d540a6f](https://github.com/pokt-network/pocket-core/pull/1449/commits/d540a6fdb0bdd6a472d096ac425636f767acf515)

x/nodes/keeper/slash.go
Andrew-Pohl: addressed in [7f6fbed](https://github.com/pokt-network/pocket-core/pull/1449/commits/7f6fbed326cae72229041a71f1735055a82f425f)

x/nodes/keeper/reward.go
Andrew-Pohl: addressed in [7f6fbed](https://github.com/pokt-network/pocket-core/pull/1449/commits/7f6fbed326cae72229041a71f1735055a82f425f)

x/nodes/keeper/reward.go
Andrew-Pohl: fixed in 26ccf1c7c1fb26c47ea42acedf946984b7241b33

x/nodes/keeper/slash.go
Andrew-Pohl: fixed in 26ccf1c7c1fb26c47ea42acedf946984b7241b33

x/nodes/types/errors.go
Andrew-Pohl: fixed in 26ccf1c7c1fb26c47ea42acedf946984b7241b33

x/nodes/keeper/valStateChanges.go
luyzdeleon: Bumping stake in the same bin is a desired behaviour, because a node might be getting slashed and they might be wanting to top up that stake to avoid falling to the lower bin.

x/gov/module.go
```diff
@@ -112,6 +112,18 @@ func (am AppModule) ExportGenesis(ctx sdk.Ctx) json.RawMessage {
 
 // module begin-block
 func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
+	
+	if am.keeper.GetCodec().IsOnNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey) {
+		gParams := am.keeper.GetParams(ctx)
```
luyzdeleon: This variable should be named `params` to mantain consistent nomenclature with other places where `GetParams` is called.

x/nodes/keeper/valStateChanges.go
Andrew-Pohl: The purpose of VEDIT is to prevent continually topping up stake resulting in bloating so this would break that use case. Could add a guard to allow it if within 10pokt of dropping to the next bin. 

If it was LT only here then this would already be covered as you can't lower stake anyway.

i.e something like 

if flooredNewStake.LTE(currentValidator.StakedTokens) && (amount.Sub(flooredStake) > 10 && amount.Sub(flooredStake) > 0 )

luyzdeleon (changes requested):
Small nomenclature change and clarified formula for edit stake.

x/nodes/keeper/valStateChanges.go
luyzdeleon: You are correct! I amend my assessment and consider this resolved.

Andrew-Pohl: Cool want me to add a guard if within X pokt of lower bound then allow an update?

x/gov/module.go
Andrew-Pohl: addressed in https://github.com/pokt-network/pocket-core/pull/1449/commits/8f4f0acaac4df17338bd685c442744fb2488a2c1

luyzdeleon (approved):

Andrew-Pohl:
@luyzdeleon rebased and forced pushed :)

x/nodes/types/params.go
```diff
@@ -91,28 +103,36 @@ func (p *Params) ParamSetPairs() sdk.ParamSetPairs {
...
+		DowntimeJailDuration:                 DefaultDowntimeJailDuration,
+		SlashFractionDoubleSign:              DefaultSlashFractionDoubleSign,
+		SlashFractionDowntime:                DefaultSlashFractionDowntime,
+		SessionBlockFrequency:                DefaultSessionBlocktime,
+		DAOAllocation:                        DefaultDAOAllocation,
+		ProposerAllocation:                   DefaultProposerAllocation,
+		RelaysToTokensMultiplier:             DefaultRelaysToTokensMultiplier,
+		MaximumChains:                        DefaultMaxChains,
+		MaxJailedBlocks:                      DefaultMaxJailedBlocks,
+		ServicerStakeFloorMultiplier:         DefaultServicerStakeFloorMultiplier,
```
oten91, line 128: Remove 
```
		ServicerStakeFloorMultiplier:         DefaultServicerStakeFloorMultiplier,
		ServicerStakeWeightMultiplier:        DefaultServicerStakeWeightMultiplier,
		ServicerStakeWeightCeiling:           DefaultServicerStakeWeightCeiling,
		ServicerStakeFloorMultiplierExponent: DefaultServicerStakeFloorMultiplierExponent,
``` 
From default parameters, as these parameters are not part of the original parameter set, this breaks compatibility when calling initgenesis.  Test broken : TestDefaultParams line 12 x/nodes/types/params_test.go

x/nodes/types/params.go
```diff
@@ -11,65 +11,77 @@ import (
...
+	KeyRelaysToTokensMultiplier                 = []byte("RelaysToTokensMultiplier")
+	KeySessionBlock                             = []byte("BlocksPerSession")
+	KeyDAOAllocation                            = []byte("DAOAllocation")
+	KeyProposerAllocation                       = []byte("ProposerPercentage")
+	KeyMaxChains                                = []byte("MaximumChains")
+	KeyMaxJailedBlocks                          = []byte("MaxJailedBlocks")
+	KeyServicerStakeFloorMultiplier             = []byte("ServicerStakeFloorMultiplier")
+	KeyServicerStakeWeightMultiplier            = []byte("ServicerStakeWeightMultiplier")
+	KeyServicerStakeWeightCeiling               = []byte("ServicerStakeWeightCeiling")
+	KeyServicerStakeFloorMultiplierExponent     = []byte("ServicerStakeFloorMultiplerExponent")
```
oten91: typo should be : []byte("ServicerStakeFloorMultiplierExponent"

oten91 (changes requested):
Hey @Andrew-Pohl left 2 small comments. Also if you can, please squash all the commits when you have it ready. 

Great job!

x/nodes/types/params.go
oten91: Keep in mind that changing this may break other test that relied on default values as they all must be 0 until the feature is turned on

x/nodes/types/params.go
Andrew-Pohl: Good spot! fixed in the squashed commit

x/nodes/types/params.go
Andrew-Pohl: Ahh ok makes sense! have removed the entries in the squashed commit

Andrew-Pohl:
Thanks @oten91 made the changes you suggested and squashed all my commits and forced pushed too [2a30eba](https://github.com/pokt-network/pocket-core/commit/2a30eba1d9e5e251401e49d9fcf05397b83a8c0e)

oten91 (approved):

POKT-Discourse:

https://forum.pokt.network/t/pep-40-stake-weight-servicer-rewards-pip-22-development-reimbursement/3389/1

POKT-Discourse:

https://forum.pokt.network/t/pip-38-re-electing-michael-orourke-to-the-foundation/5292/1

POKT-Discourse:

https://forum.pokt.network/t/pip-38-proposal-for-pocket-network-governance-and-economic-reform-revised/5327/1

Issue #1474, opened 2022-09-12
Beta-0.9.1 Feature Flag


jessicadaugherty:
# Objective
Allow pocket-core to reproduce the behavior of omitting rewards for a subset of nodes (from this point on "the undesired behavior") when the non-custodial flag was turned on, while also allowing pocket-core to have the behavior we want for "Non-Custodial".

There are several ways to achieve this; letting the original flag preserve its undesired behavior and creating a new flag for the desired behavior is a mechanism that allows preserving the history of the blockchain and allows for quick testing of both the undesired and desired behaviors which would be needed to synch the blockchain.

# Goals / Deliverables
- [ ] Allowing the undesired behavior to happen when the existing flag is turned on
- [ ] Allowing the desired behavior to happen when a new flag is turned on
- [ ] Notify @emedrano2112  to re-test

# Non-goals
- Disallowing the feature flag that causes the undesired behavior

# Testing Methodology
- Confirm that turning on the original flag causes the undesired behavior
- Confirm that turning off the original flag stops the undesired behavior
- Confirm that turning on the new flag causes non-custodial to kick in
- Confirm that turning off the new flag causes non-custodial to stop

Creator: @jessicadaugherty

iajrz:
Fleshing out the alternatives we have for the historical replay of the NC issue (part of this is documented [here](https://github.com/pokt-network/pocket-core/pull/1476#issuecomment-1247376270), I'm aiming to distill pros and cons here).

The alternatives to compare are these:
- We can do as it's done in #1476 and we've done it before (from now on **Classical Style**)
- We can create a flag for the desired behavior and allow the undesired behavior to have its own flag (from now on **Proposed Style**)

**Classical Style Pros**
- We have done this before, which implies
  - It's tried and true
  - Cases are handled homogeneously

**Classical Style Cons**
- We increase the amount of blocks we need to process in an isolated cluster in order to test the flag behaviors by 30%
- We're overloading behaviors, which means activating the flag in an isolated cluster has two different behaviors which are height dependent.
- It might require an increasing the load of the `codec.TestMode` variable to allow unit testing to verify the behavior in isolation

**Proposed Style Pros**
- We make it possible to test the NC issue behavior and the desired behavior at any height
  - it is self contained; no need to overload the `codec.TestMode` variable
- We isolate behaviors, which means that activating the flag in an isolated cluster has a specific behavior regardless of height, increasing predictability
- We know how to set it up: we've added flags before

[...]

PR #1476, opened 2022-09-12
allowed 'historical replay' of the noncustodial issue

andrewnguyen22:
- [ ] allows for historical replay of the noncust issue

### DIFF
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -36,18 +36,19 @@ var (
 )
 
 const (
-	UpgradeCodecHeight      = int64(30024)
-	CodecChainHaltHeight    = int64(30334)
-	ValidatorSplitHeight    = int64(45353)
-	UpgradeCodecUpdateKey   = "CODEC"
-	ValidatorSplitUpdateKey = "SPLIT"
-	NonCustodialUpdateKey   = "NCUST"
-	TxCacheEnhancementKey   = "REDUP"
-	MaxRelayProtKey         = "MREL"
-	ReplayBurnKey           = "REPBR"
-	BlockSizeModifyKey      = "BLOCK"
-	RSCALKey                = "RSCAL"
-	VEDITKey                = "VEDIT"
+	UpgradeCodecHeight          = int64(30024)
+	CodecChainHaltHeight        = int64(30334)
+	ValidatorSplitHeight        = int64(45353)
+	NonCustodial1RollbackHeight = int64(69583)
+	UpgradeCodecUpdateKey       = "CODEC"
+	ValidatorSplitUpdateKey     = "SPLIT"
+	NonCustodialUpdateKey       = "NCUST"
+	TxCacheEnhancementKey       = "REDUP"
+	MaxRelayProtKey             = "MREL"
+	ReplayBurnKey               = "REPBR"
+	BlockSizeModifyKey          = "BLOCK"
+	RSCALKey                    = "RSCAL"
+	VEDITKey                    = "VEDIT"
 )
 
 func GetCodecUpgradeHeight() int64 {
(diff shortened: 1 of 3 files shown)


andrewnguyen22:
closes #1474

oten91 (commented):
@andrewnguyen22 can you double check the "NonCustodial1RollbackHeight", as we are using <= while comparing to set "isDuringNonCustodialIssue". Everything seems in order but heavily dependent on this height.

An alternative strategy like using today's current height (or an arbitrary height we know the feature is disabled after the rollback) as a safe height to activate the new behaviour may be other option.

iajrz:
@oten91 we're looking into the possibility of doing this fix differently from what we've done before, also without the `<=` vs `<` risk. Jessica created an issue for it in #1474; core idea is to not overload the network parameter and reap the benefits of that both in transparent behavior and easier/faster testing of the state machine permutations.

andrewnguyen22:
What risk is there with `<` vs `<=` - either way, the flag was already rolled back so we are safe

iajrz:
You are correct, I ran with Otto's idea without giving it much thought: in this particular case we're safe. I'm sure his observation is tied to past events around the codec change height and synchronization, so it's more an observation about safety in the general case rather than the specifics of this incident.

oten91:
@iajrz Seem I'm missing context as the idea expressed on the issue is not super clear to me after reading your comment. 

Also I was under the impression that this was the related PR, and it seemed like a simple and good enough solution to tackle the sync from scratch scenario.

iajrz:
Yes, this is the PR referred to from the issue. There seems to have been a miscommunication when the sync from scratch scenario was identified and solutions evaluated. We have alternatives to explore.

The core idea behind my proposal is this: as long as we add hardcoded heights to the executable we keep increasing the height a localnet needs to grow to in order to be able to fully test the scenarios.

The effect of this is that even if we were to have a localnet going through 100 heights per second a single scenario related to this would take over 10 minutes to test, an increase of over 30% of time with regards to the last height used to gate behavior.

There is another argument to be made: using the same flag for more than one functionality unfavorably shapes testing parameter combinations.

I think we can discuss this when you get back. Get better.

codec/codec.go
```diff
@@ -36,18 +36,19 @@ var (
...
-	TxCacheEnhancementKey   = "REDUP"
-	MaxRelayProtKey         = "MREL"
-	ReplayBurnKey           = "REPBR"
-	BlockSizeModifyKey      = "BLOCK"
-	RSCALKey                = "RSCAL"
-	VEDITKey                = "VEDIT"
+	UpgradeCodecHeight          = int64(30024)
+	CodecChainHaltHeight        = int64(30334)
+	ValidatorSplitHeight        = int64(45353)
+	NonCustodial1RollbackHeight = int64(69583)
```
Olshansk, line 42: What is the purpose of the `1` in `NonCustodial1RollbackHeight`?

x/nodes/keeper/reward.go
```diff
@@ -10,21 +10,38 @@ import (
...
 		return sdk.ZeroInt()
 	}
 
-	if k.Cdc.IsAfterNonCustodialUpgrade(ctx.BlockHeight()) {
+	if isAfterNonCustodial {
 		address = k.GetOutputAddressFromValidator(validator)
 	}
 
+	// This simulates the issue that happened between the original non-custodial rollout and the rollback height
+	// This code is needed to allow a 'history replay' of the issue during sync-from-scratch
```
Olshansk, line 32: Not actionable: good learning lesson for myself on the importance of blockchain history & immutability.

Olshansk (approved):
Code LGTM. Left one comment but would also want an approval from @oten91 before merge.

oten91 (approved):

oten91:
For context , documentation and answering my original ask of double checking the height, I double checked it and height 69583 should work without problems:

This is the rollback tx https://pokt.watch/tx/BD2AB44335EE3355EDECDAA10613602D0CEAD2A5D57B518E5ECA1F2678C022CC .

PR #1483, opened 2022-10-12
Added /v1/health endpoint

### LINKED ISSUE #1456: Health endpoint
Add and "status" or "health" endpoint that provide fast results of the state of the node, like for example if is catching up or not, if stake or not, etc.
In this particular case I accept any kind of suggestion or advice, I was thinking to add status information that is not sensitive from tendermint layer that help a node runner or an automation tool to see/check if a node is running well.

**Please provide a justification for your change.**
There is no way currently to check the health of the node.
There is no way using pocket rpc to see if the node is catching up, but is if you know and use tendermint rpc, but people should be able to only point and use pocket node for anything around their nodes.

**Timeline**
After any discussion and ideas of what information this could expose safety, should be a trivial task of 1 day or 2.

jorgecuesta:
Fix #1456

Request:
`curl http://localhost:8081/v1/health`

Response:
```json
{
  "height": 8,
  "is_catching_up": false,
  "is_starting": false,
  "version": "BETA-0.9.2"
}
```

### DIFF
--- a/app/cmd/rpc/query.go
+++ b/app/cmd/rpc/query.go
@@ -26,6 +26,16 @@ func Version(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	WriteResponse(w, APIVersion, r.URL.Path, r.Host)
 }
 
+func Health(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	res := app.PCA.QueryHealth(APIVersion)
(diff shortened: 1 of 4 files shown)


oten91:
Hey @jorgecuesta can you rebase this ?

oten91 (approved):
Looks good. Simple and non invasive change. approved. 

Thanks @jorgecuesta

PR #1485, opened 2022-10-12
Fix 2 issues and minor enhancement.

### LINKED ISSUE #1457: [BUG REPORT] Updating Evidences doesn't guarantee an atomic write
**Describe the bug**
With the current implementation of [SetProof](https://github.com/pokt-network/pocket-core/blob/017ae6e86a3b092acd82018b7022f2fb5c3f4624/x/pocketcore/types/cache.go#L405), this results in a race condition where evidences are not written into the cache storage atomically. This can result in evidences overwriting each other whenever concurrent relays are handled, and the node not being paid for the right amount of work when submitting a claim.

Consider this scenario:
1. Two relays are sent to the node concurrently, labeled respectively A and B relay.
2. Node validates both A and B relays
3. Node retrieves the evidence to update relays A's proof, then subsequently grabs the evidence to update relays B proofs.
4. The node at this time currently has a local copy of evidence for Relay A and another copy for Relay B. (two separate evidences in memory at this point!)
5. The local copy of evidence A is updated with the proof, then updated into cache storage.
6. The local copy of evidence B is updated with the proof, then updated into cache storage - overwriting the previous written evidence.
7. Evidence storage now only contains 1 proof, where the expected amount is to be two.

Proof of concept test [here](https://github.com/PoktBlade/pocket-core/commit/414f1329723146bdd2f9a3bf1c3c910f89c568d7) I wrote it really quickly, might've missed something

**Expected behavior**
We need a way to ensure that proofs are written to the evidence storage atomically. There's multiple ways to do this, you can add a lock within the SetProof function, and then every call to SetProof can be completed in a different go routine. By putting it in a different go routine, it won't block the response back to the caller due to lock contention (i.e under high load)

Or the more proper way is to set up a queue using channels that will buffer relay proofs to a worker that will update the evidences one by one, similarly to the consumers/producer problem.

### LINKED ISSUE #1484: Avoid http.Client recreation on every relay
Create a single HTTP client that will be used to call chains, using the best tunning performance settings for it based on a bit of research (google it) about how to speed up and reduce open files in Golang http.Client

**Please provide a justification for your change.**
Every time a relay is received on the background of that process, a new http.Client creates and this is not the best way to do, it because it will not reuse connections, it will open more and more file descriptors and that could reach system limits or even grows up the needed RAM to process bigger relay bursts.

jorgecuesta:
Fix #1457; 
Fix #1484; 
Fix a high memory consumption that also is part of issue #1457.
Under the high load of requests (1000/rps or more), the RAM got crazy and scaled up to 40GB or close to that.
Now after the fix of #1457 with the worker pool, the node remains under 14GB of ram in my local tests.

### DIFF
--- a/x/pocketcore/types/service.go
+++ b/x/pocketcore/types/service.go
@@ -30,7 +30,11 @@ type Relay struct {
 
 func GetChainsClient() *http.Client {
 	if chainHttpClient == nil {
-		InitHttpClient(1000, 1000, 1000)
+		InitHttpClient(
+			GlobalPocketConfig.RPCMaxIdleConns,
+			GlobalPocketConfig.RPCMaxConnsPerHost,
+			GlobalPocketConfig.RPCMaxIdleConnsPerHost,
+		)
 	}
 
 	return chainHttpClient
--- a/x/pocketcore/types/service.go
+++ b/x/pocketcore/types/service.go
@@ -17,11 +17,45 @@ import (
 
 const DEFAULTHTTPMETHOD = "POST"
 
+var (
+	chainHttpClient *http.Client
+)
+
 // "Relay" - A read / write API request from a hosted (non native) external blockchain
 type Relay struct {
 	Payload Payload    `json:"payload"` // the data payload of the request
 	Meta    RelayMeta  `json:"meta"`    // metadata for the relay request
 	Proof   RelayProof `json:"proof"`   // the authentication scheme needed for work
+}
+
+func GetChainsClient() *http.Client {
+	if chainHttpClient == nil {
+		InitHttpClient(1000, 1000, 1000)
+	}
+
+	return chainHttpClient
+}
+
+func InitHttpClient(maxIdleConns, maxConnsPerHost, maxIdleConnsPerHost int) {
+	var chainsTransport *http.Transport
+
+	t, ok := http.DefaultTransport.(*http.Transport)
+	if ok {
+		chainsTransport = t.Clone()
+		// this params may could be handled by config.json, but how much people know about this?
+		// tbd: figure out the right values to this, rn the priority is stop recreating new http client and connections.
+		chainsTransport.MaxIdleConns = maxIdleConns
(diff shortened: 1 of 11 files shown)


oten91:
Hey @jorgecuesta please rebase.

x/pocketcore/types/service.go
```diff
@@ -17,13 +17,47 @@ import (
...
 // "Relay" - A read / write API request from a hosted (non native) external blockchain
 type Relay struct {
 	Payload Payload    `json:"payload"` // the data payload of the request
 	Meta    RelayMeta  `json:"meta"`    // metadata for the relay request
 	Proof   RelayProof `json:"proof"`   // the authentication scheme needed for work
 }
 
+func GetChainsClient() *http.Client {
+	if chainHttpClient == nil {
+		InitHttpClient(1000, 1000, 1000)
```
oten91: can we remove the hardcoded values?

oten91: they seem to be defined and used on other calls
 	DefaultRPCMaxIdleConns             = 1000
	DefaultRPCMaxConnsPerHost          = 1000
	DefaultRPCMaxIdleConnsPerHost      = 1000

oten91 (changes requested):

oten91:
@jorgecuesta  looks good, left you a comment

x/pocketcore/types/service.go
jorgecuesta: My concern about using the config approach is that this function is inside of `x/pocketcore/types` and the config is part of the `app` package. My understanding here is that the `app` package should use `x/pocketcore/types` and not in a reverse way.

oten91: Yeah, a cycle dependency should be avoided , look for the usage of globalRPCTimeout as an example

jorgecuesta: I update it to use the GlobalPocketConfig load from InitConfig on `x/pocketcore/types/config.go`

oten91: Thanks!

PR #1489, opened 2022-10-25
Technology sea

Olshansk:


x/nodes/keeper/reward.go
```diff
@@ -25,7 +25,11 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	}
 
 	if isAfterNonCustodial {
-		address = k.GetOutputAddressFromValidator(validator)
+		// if during non-custodial rollback issue on height 69583, use the output address
+		// else only use the output address if after noncustodial allowance height 74622
+		if isDuringNonCustodialIssue || ctx.BlockHeight() >= codec.NonCustodial2AllowanceHeight {
```
Olshansk: Rename `isDuringNonCustodialIssue` to `IsDuringFirstNonCustodialIssue`

x/nodes/keeper/reward.go
```diff
@@ -25,7 +25,11 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	}
 
 	if isAfterNonCustodial {
```
Olshansk: This outer if structure can be removed

x/nodes/keeper/reward.go
Olshansk: Makes me think that the behaviour is unexpected between the second activation height `74620` and the chain halt height `74622`.

andrewnguyen22: The behavior the validators accidentally chose (due to the cache) is to pull a `stale` validator state effectively ignoring the non-cust upgrade on height `74620` - this means that the expected behavior is to ignore non-cust in this part of the code until `74622`

x/nodes/keeper/reward.go
andrewnguyen22: This can be rolled back with the feature map

andrewnguyen22: So we can't remove it without losing that protection

x/nodes/keeper/reward.go
Olshansk: Acknowledged.

Documenting that even though the world state has it enabled at `74620`, since the hardcoded value is `74622`, it's actually disabled between `74620` and `74622`.

PR #1506, opened 2022-11-28
Geo-Mesh

jorgecuesta:
POKTscan Geo-Mesh feature. 

TLDR: https://forum.pokt.network/t/poktscan-s-geo-mesh/3585


Olshansk:
Closing this out as geo-mesh will not be merged (until a further discussion) into `pocket-core` but is still seen as one of the other biggest clients on the network and shoutout to the poktscan team @jorgecuesta for keeping up with upstream!

PR #1512, opened 2022-12-06
Fix bad and outdated links

bmmpxf:
A few documentation links were pointing to the old structure, and a few were 404ing. This PR fixes them.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -26,7 +26,7 @@ visit [pokt.network](https://pokt.network).
 
 [Protocol Forum](https://forum.pokt.network/)
 
-[Protocol Wiki](https://docs.pokt.network/home/main-concepts/protocol)
+[Protocol Wiki](https://docs.pokt.network/learn/protocol)
 
 [Community Discord](https://discord.com/invite/KRrqfd3tAK)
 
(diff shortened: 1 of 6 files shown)


bmmpxf:
FYI, I'm seeing that there's a GitHub action that builds the docs on GitBook. We're no longer using GitBook, so those actions can be turned off. (Not sure who handles that.)

image

jessicadaugherty (approved):
I have tested all updated URLs to confirm they resolve to the intended host. Will check on disabling the Gitbook hooks.

jessicadaugherty:
Update on Gitbook integration: disabling requires account owner to 2FA so will need to figure that out before removing it from this workflow @bmmpxf

bmmpxf:
> Update on Gitbook integration: disabling requires account owner to 2FA so will need to figure that out before removing it from this workflow @bmmpxf

Thanks! I'd guess the person who has the 2FA is @jacklaing .

It's not a blocker for this particular PR, but just something we don't need to have anymore.

bmmpxf:
@jessicadaugherty If you have the ability to merge this, please do whenever you get the chance, as I don't appear to have access (which is fine).

jessicadaugherty:
So sorry @bmmpxf thought you could do the honors yourself. Done!

bmmpxf:
Thank you @jessicadaugherty !

PR #1514, opened 2022-12-12
Adds the e2e test suite framework and a bit over 80 test scenarios

### LINKED ISSUE #1482: Automated e2e Testing
This proposal suggests adding an in-repo automated suite of e2e tests that can be expanded upon and help reduce the effort needed to increase the confidence of a specific release.

**Goals**
- Support the tests from [the functional tests repo](https://github.com/pokt-network/pocket-core-func-tests).
- Enable the creation of new tests
- Reduce the time necessary to run tests in the release cycle

**Non-goals**
- Increase coverage from that of existing tests
- Fix any defects revealed by the tests
- Mimic the tests' exact behavior if it would require changing pocket-core code to enable interfacing properly

**Components:**
- Network Launcher
  - Pocket Client
  - Pocket Server
  - Network Configurations
- Tests
  - cucumber style feature files, with scenarios written in gherkin
  - step implementation for scenario creation and execution
  - new **external dependency**: [godog](https://github.com/cucumber/godog), to parse feature files and run scenarios

**Details**

The code to be introduced would depend on code from pocket-core, but would not affect any builds.
The complexity introduced would be self-contained in its packages, which would live under the e2e folder at the top level.

The Network Launcher compiles pocket-core, establishes datadirs for all the defined nodes, and starts the network. It contains in runtime the information defined in the network configuration.

Network configurations allow defining a genesis file, node configurations, and hold the private keys for the actors of the network.

The Pocket Client and Pocket Server can be run with any command-line arguments. We get to read their full stdout and stderr, which allows for inspecting state transitions as evidenced by, for example, the debug log.

The tests are defined in a twofold process:
- Design and implementation of the scenarios, choosing or defining the "steps" required to verify that the system is behaving the way we want. This should be biased towards reusing existing steps. This is done in gherkin.
- Design and implementation of steps so they work together properly. This includes designing the state that should be kept from one step to the other and how it will be made available to a future step. This is done in go.

iajrz:
This PR has both Go code and a cucumber test suite in gherkin; I'd expect @Olshansk to look at the go side of the matter and @emedrano2112 and @AneudyM to look at the gherkin side of the matter.

Fixes #1482 

Co-authored-by: Daniel Olshansky <olshansky@pokt.network>


e2e/launcher/configuration.go
```diff
@@ -0,0 +1,52 @@
...
+	GenesisPath        string               `json:"genesis_path"`
+	NodeConfigurations []*NodeConfiguration `json:"node_configurations"`
+}
+
+type NodeConfiguration struct {
+	PrivateKey string `json:"private_key"`
+	ConfigPath string `json:"config_path"`
+}
+
+const networkConfigurationFileName = "network.json"
```
Olshansk: NIT: `defaultNetworkConfigurationFileName`?

e2e/launcher/configuration.go
```diff
@@ -0,0 +1,52 @@
...
+
+func loadNetworkConfiguration(networkConfigDirectory string) (*NetworkConfiguration, error) {
```
Olshansk:
```suggestion
func loadNetworkConfiguration(networkConfigDirectory string) (networkConfiguration *NetworkConfiguration, err error) {
```

Given that you have to declare these individually, consider using named return values

e2e/launcher/configuration.go
```diff
@@ -0,0 +1,52 @@
...
+	var err error
+	networkConfigDirectory, err = filepath.Abs(networkConfigDirectory)
+
```
Olshansk, line 23:
```suggestion
```

e2e/launcher/network.go
```diff
@@ -0,0 +1,43 @@
...
+	Nodes                []*Node
+	NetworkConfiguration *NetworkConfiguration
+}
+
+func LaunchNetwork(networkConfigDirectory string, executablePath string) (network Network, err error) {
+	fmt.Println("./launcher/network_configs/" + networkConfigDirectory)
+	networkConfiguration, err := loadNetworkConfiguration("../launcher/network_configs/" + networkConfigDirectory)
+	if err != nil {
+		return
+	}
```
Olshansk, line 24: You have pretty nice spaces after err if structures `e2e/launcher/configuration.go` but not here.

I personally prefer the former, but w/e you choose, please stay consistent.

Screenshot 2022-12-13 at 9 43 53 PM

e2e/launcher/network.go
```diff
@@ -0,0 +1,43 @@
...
+	networkRootDirectory, err := os.MkdirTemp("", networkConfiguration.NetworkId+"-")
+	if err != nil {
+		return
+	}
+	log.Printf("root network directory: %v", networkRootDirectory)
+	var nodes []*Node
+	for _, nodeConfiguration := range networkConfiguration.NodeConfigurations {
+		var node *Node
+		node, err = newNode(nodeConfiguration, networkRootDirectory, networkConfiguration.GenesisPath, executablePath)
```
Olshansk:
```suggestion
		node, err := newNode(nodeConfiguration, networkRootDirectory, networkConfiguration.GenesisPath, executablePath)
```

e2e/launcher/network.go
```diff
@@ -0,0 +1,43 @@
...
+	fmt.Println("./launcher/network_configs/" + networkConfigDirectory)
+	networkConfiguration, err := loadNetworkConfiguration("../launcher/network_configs/" + networkConfigDirectory)
+	if err != nil {
+		return
+	}
+	networkRootDirectory, err := os.MkdirTemp("", networkConfiguration.NetworkId+"-")
+	if err != nil {
+		return
+	}
+	log.Printf("root network directory: %v", networkRootDirectory)
```
Olshansk: You use `log.Println("string" + var)` above, but `log.Printf` here. 

What's your choice on when to use which?

e2e/launcher/network.go
```diff
@@ -0,0 +1,43 @@
...
+		if err != nil {
+			return
+		}
+		nodes = append(nodes, node)
+	}
+	network = Network{
+		Nodes:                nodes,
+		NetworkConfiguration: networkConfiguration,
+	}
+	return
```
Olshansk:
```suggestion
	return &Network{
		Nodes:                nodes,
		NetworkConfiguration: networkConfiguration,
	}, nil
```

e2e/launcher/network.go
```diff
@@ -0,0 +1,43 @@
...
+	"log"
+	"os"
+)
+
+type Network struct {
+	Nodes                []*Node
+	NetworkConfiguration *NetworkConfiguration
+}
+
+func LaunchNetwork(networkConfigDirectory string, executablePath string) (network Network, err error) {
```
Olshansk: Why not return a `*Network`?

e2e/launcher/node.go
```diff
@@ -0,0 +1,147 @@
...
+	if err = writeNodeKeyFile(privateKey, nodeDataDir, pocketCoreConfig.TendermintConfig.NodeKey); err != nil {
+		return
+	}
+
+	if err = writePrivValState(nodeDataDir, pocketCoreConfig.TendermintConfig.PrivValidatorState); err != nil {
+		return
+	}
+
+	// initialize returnable node
+	node = &Node{
```
Olshansk: In these types of situations, consider `return &Node{...}, nil`

e2e/launcher/node.go
```diff
@@ -0,0 +1,147 @@
...
+	pocketCoreConfig.TendermintConfig.P2P.RootDir = nodeDataDir
+	pocketCoreConfig.TendermintConfig.Mempool.RootDir = nodeDataDir
+	pocketCoreConfig.TendermintConfig.Consensus.RootDir = nodeDataDir
+	pocketCoreConfig.PocketConfig.DataDir = nodeDataDir
+
+	// write config.json & genesis.json
+	if err = os.Mkdir(filepath.Join(nodeDataDir, "config"), os.ModePerm); err != nil {
+		return
+	}
+
```
Olshansk, line 67: Personal preference (feel free to ignore).

I would've probably declared interim variables for readability and used in place.

```go
configFilePath := filepath.Join(nodeDataDir, "config", "config.json")
genesisFilePath := filepath.Join(nodeDataDir, "config", "genesis.json")
```

e2e/launcher/node.go
```diff
@@ -0,0 +1,147 @@
...
+		configuration: pocketCoreConfig,
+		privateKey:    privateKey,
+		PocketServer:  NewPocketServer(executableLocation),
+	}
+	return
+}
+
+func writePrivValState(datadir, privValState string) error {
+	cdc := app.Codec()
+	pvkBz, err := cdc.MarshalJSONIndent(privval.FilePVLastSignState{}, "", "  ")
```
Olshansk: Doesn't cost much to inline here. DItto below.

```suggestion
	pvkBz, err := app.Codec().MarshalJSONIndent(privval.FilePVLastSignState{}, "", "  ")
```

e2e/launcher/pocketclient.go
```diff
@@ -0,0 +1,55 @@
...
+
+type CommandResult struct {
+	Stdout string
+	Stderr string
+	Err    error
+}
+
+type writerByteArray []byte
+
+func (so *writerByteArray) Write(p []byte) (n int, err error) {
```
Olshansk: s/so/ba so its general?

e2e/launcher/pocketclient.go
```diff
@@ -0,0 +1,55 @@
...
+	*so = append(*so, p...)
+	return len(p), nil
+}
+
+type PocketClient interface {
+	RunCommand(...string) (*CommandResult, error)
+}
+
+func NewPocketClient(executableLocation string, verbose bool) PocketClient {
```
Olshansk: NIT: put factory function below type defitnion.

e2e/launcher/pocketclient.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+type pocketClient struct {
+	executableLocation string
+	verbose            bool
+}
+
+func (pc *pocketClient) RunCommand(commandAndArgs ...string) (*CommandResult, error) {
+	if pc.verbose {
+		fmt.Printf("Running Command: %v\n", commandAndArgs)
```
Olshansk: Why are we using `fmt` instead of `log` everywhere?

e2e/launcher/pocketclient.go
```diff
@@ -0,0 +1,55 @@
...
+
+	so := &writerByteArray{}
+	se := &writerByteArray{}
+
+	cmd.Stdout = so
+	cmd.Stderr = se
+	err := cmd.Run()
+
+	return &CommandResult{
+		Stdout: string(*so),
```
Olshansk: Since we're using the byte array to keep appending and convert it to a string in the end, thoughts on using the strings builder in golang? It implements `Write` by default: https://pkg.go.dev/strings#Builder.Write which is the interface `Stdout`  above needs.

e2e/launcher/pocketserver.go
```diff
@@ -0,0 +1,124 @@
...
+		ps.stdOutPipeline.patternActors = append(ps.stdOutPipeline.patternActors, patternActor)
+	case StdErr:
+		ps.stdErrPipeline.patternActors = append(ps.stdOutPipeline.patternActors, patternActor)
+	default:
+		return errors.New("invalid stream")
+	}
+	return nil
+}
+
+type PatternActor interface {
```
Olshansk: DISCUSS: I've always been a fan of declaring all the types in one place or at the top of the file.

I've never gotten accustomed to having types be defined midfile, but is this a common golang pattern?

e2e/launcher/pocketserver.go
```diff
@@ -0,0 +1,124 @@
...
+
+func NewPocketServer(executableLocation string) PocketServer {
+	return &pocketServer{
+		executableLocation: executableLocation,
+		pocketInstance:     nil,
+	}
+}
+
+type pocketServer struct {
+	executableLocation string
```
Olshansk: By location, do we mean `Path`? If so, could/should we rename it?

e2e/launcher/pocketserver.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+	"sync"
+)
+
+type PocketServer interface {
+	Start(...string) error
+	Kill() error
+	RegisterPatternActor(patternActor PatternActor, stream Stream) error
```
Olshansk: DISCUSS: I'm familiar with the "Actor Model" from prior work but this is the first I've heard of "Pattern Actor". 

Can you link to any other work that references this naming convention?

Olshansk (changes requested):
@iajrz I've still got a long way to go (see screenshot) but figured I'd publish what I have so far so we can start working in parallel as I continue.

Screenshot 2022-12-13 at 10 10 45 PM

e2e/launcher/network.go
iajrz: This is a no-no: in the worst case scenario changing code to be like this would shadow the return value. In this particular case, it would merely cause a compilation error.
Here, two examples:
- [here](https://go.dev/play/p/399zRdoXppd) for the shadowing case (run and look at the logs).
- [here](https://go.dev/play/p/86efKsyifB7) for the compilation error case

e2e/launcher/pocketclient.go
iajrz: Been looking into this; considered a byteBuffer, too. Keeping it in consideration.

e2e/launcher/pocketserver.go
iajrz: yes to all

e2e/launcher/pocketserver.go
iajrz: My pattern is shaped by years of Java development (declaration and definition are together, so types end up mid-file). The pattern you prefer reminds me of C and C++ development, where declaration and definition are separate. 

Navigation between files is usually done by symbol cross reference, so we don't waste much time finding things because they're far from the top; furthermore, having all of the types at the top will yield a group of functions unrelated to each other hanging out together at the end of the file.

I had way more text, but decided to go to `io/io.go` in the standard library for guidance. This is how that file is ordered:
- Package level constants and variables
- Interfaces
- Functions with no receivers (operate on interfaces)
- all the structs with their related data around:
  - struct initializer (NewXxYy and similar)
  - struct
  - functions with the struct or pointer to the struct as a receiver

structs are ordered so that the ones below depend on the ones above.

I think this works because of this:
- First you see package level things you might need to customize or refer to
- Then interfaces you might want to implement
- Then functions you might need to use on the interfaces
- Then structs and related implementation nearby
- In the only case you'd read from top to bottom (generally when studying a module to learn it/from it) the fact that types go from more general to more specific means that you get a sense for the architecture / rationale.

I think we should emulate this pattern in our codebases.

e2e/launcher/pocketserver.go
iajrz: Reference to the pattern-action paradigm as seen in awk and similar tools, given the similar usage: if the pattern is matched, we do stuff.
Open to renaming.

e2e/launcher/network.go
iajrz: My personal preferences is the use of spaces -- in fact, blank line before an `else` too when I'm being especially mindful.
Usage in golang seems to go against whitespace. In fact, standard library has no breathing room _within_ functions, although it has plenty _between_ functions.

I see you like the spaces, too. Bad thing about spaces is you need to think of where to put them and configure the editor so you never forget. If you never have space you never have to do that.

How're you doing it in v1 code? How'd you like it to be in the future? I could do either and I'm going to configure the editor to stick to the official option as closely as possible.

e2e/launcher/network.go
iajrz: The code was not written sequentially; I did not stop to think very closely about fmt vs log, and `f` vs `ln`. I will think about this during the revision to comply with the review, though.

e2e/launcher/network.go
iajrz: Decision was possibly made because members of the struct already point at things. I will revisit this line of thought and possibly be better able to articulate why I decide whatever I end up with.

e2e/launcher/network.go
iajrz: Will think about this and address :) both the named returns and the matter of the pointer

e2e/launcher/node.go
iajrz: I have to think about named vs unnamed returns, which is related to this.

e2e/launcher/node.go
iajrz: I generally don't do that except for things that are repeated. Usually 3rd repetition is when I name the thing. Bearing in mind.

e2e/launcher/node.go
iajrz: Indeed, this was a mistake borne from rewriting. Possibly the value of `app.Codec` should live in a package level variable.

e2e/launcher/pocketclient.go
iajrz: NOT (see what I did there?): this is in the style of the go standard library. I say more about that [in this comment](https://github.com/pokt-network/pocket-core/pull/1514#discussion_r1048961345) -- which if you're reading top to bottom you already saw.

e2e/launcher/pocketclient.go
iajrz: Looks like it should be like that; I think this was lost in translation from where the type had another name.

e2e/launcher/pocketclient.go
iajrz: It's a mistake, as mentioned elsewhere; will think through & normalize. In general I think "useful when using" vs "useful when debugging" should guide that decision, but that might not be what I'm actually doing here.

e2e/launcher/configuration.go
Olshansk: I see a 😄 emoji but no comments, resolution or change. Will you make the change?

e2e/launcher/configuration.go
Olshansk: I see an 👀 , but no comments/resolution/change. What does this mean?

e2e/launcher/network.go
Olshansk: Use the spaces approach. Apply everywhere in this PR. It's what we're aiming for in v1 as well.

e2e/launcher/pocketserver.go
Olshansk: Please update

e2e/launcher/pocketserver.go
Olshansk: Let's rename to `patternAction`

e2e/launcher/pocketclient.go
Olshansk: Let's change to `log`.

e2e/launcher/pocketserver.go
Olshansk: Sounds good to me. Leave as is.

e2e/launcher/pocketclient.go
Olshansk: In the future, no need to comment on stuff like this. Just make the change and `Resolve conversation`.

e2e/launcher/node.go
Olshansk: No need to leave comments on stuff like this moving forward, if you're okay with the change -> make change -> `Resolve conversation` -> continue

Olshansk:
@iajrz I'll continue reviewing once the existing comments are resolved or Once the existing comments are resolved or tended

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query an existing account
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query account {{address}}" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/account"
+	And the user should be able to see standard output containing "address"
+
+  Scenario: To query an existing account, wrong address
+	Given the user is running the network "single_node_network"
```
AneudyM, line 18: I recommend the use of hooks here to take advantage of reusability. Lots of repeated steps. Background would work for this case.

AneudyM (changes requested):

e2e/tests/accounts.feature
```diff
@@ -0,0 +1,78 @@
...
+	  | "accounts export thisisdefinitelyawrongaccount --pwd-decrypt=badpassword"     | "Address Error"                |
+	  | "accounts export-raw thisisdefinitelyawrongaccount --pwd-decrypt=badpassword" | "Address Error"                |
+	  | "accounts sign address testing"                                               | "invalid byte"                 |
+	  | "accounts create-multi-public thisisdefinitelyawrongaccount"                  | "error in public key creation" |
+
+  Scenario: Account Sign With Non-Existing Account
+	Given the user has a pocket client
+	When the user runs the command "accounts sign 7ab712998671b09e1a266ce6901000acb657833b 'example_message'"
+	Then the user should be able to see standard output containing "invalid byte"
+	And the pocket client should have exited without error
```
AneudyM, line 65: Ideally, you'd want Then to be the last step, since its purpose is to observe outcomes. Here, each test is validating that the client exited without errors. That should be an individual scenario, not one to do in every scenario. For example:

Given the user has a pocket client
When the user runs the command ...
Then the user should see ...       <-- this is the outcome
---END---

e2e/tests/query.feature
iajrz: I agree that some scenarios could have fewer visible steps by using Background; as mentioned in the PR's initial comment, though, we're looking for accuracy in the gherkin side, so I think this can stay as is.

e2e/tests/accounts.feature
iajrz: I understand where you're coming from; in this case -- a command line interface's API is comprised of its stdout, stderr and exit status. In particular, scripts often use the exit status to know whether the script can continue or needs to abort.

With regards to the `Then...` `And...`, the preference is aesthetic in this case; we're checking for two different outcomes, and it reads easier than `Then...` `Then...`.

e2e/launcher/configuration.go
iajrz: The emoji was because this literally made me smile and to remind self that I'd seen it. In this case: network configuration file name is not configurable, so it's not "default", but the "actual" name. It doesn't gel to make it configurable, either, so I won't change it.

Olshansk:
@iajrz I saw that you pushed some updates. Please re-request a review (see screenshot) when it's ready for another round

Screenshot 2022-12-20 at 3 50 38 PM

e2e/tests/accounts.feature
```diff
@@ -0,0 +1,78 @@
...
+	  | "accounts send-tx"                                       | "accounts send-tx"           |
+	  | "accounts show"                                          | "accounts show"              |
+	  | "accounts sign"                                          | "accounts sign"              |
+	  | "accounts sign 7ab712998671b09e1a266ce6901000acb657833b" | "accounts sign"              |
+	  | "accounts sign message"                                  | "accounts sign"              |
+	  | "accounts update-passphrase"                             | "accounts update-passphrase" |
+
+  Scenario Outline: Running Commands With Correct Syntax and Bad Data
+	Given the user has a pocket client
+	Given the user runs the command <bad_command>
```
emedrano2112: Change given, by When. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
+Feature: Pocket Query Commands
+
+  Scenario: To show existing commands within the pocket query section
+	Given the user has a pocket client
+	When the user runs the command "query"
+	Then the user should be able to see standard output containing "Usage:"
+
+  Scenario: To query an existing account
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query account {{address}}" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/account"
+	And the user should be able to see standard output containing "address"
+
+  Scenario: To query an existing account, wrong address
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query account {{address}} 0" against validator -1
+	Then the user should be able to see standard output containing "encoding/hex: invalid byte:"
+
+  Scenario: Pocket Query Account address is invalid But Has valid Bytes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query account 2BBCA5DC9792C72AC3A2616910C4AAAA 0"
+	Then the user should be able to see standard output containing "Incorrect address length"
+
+  Scenario: To query an existing account, wrong height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query account {{address}} 5600000" against validator 0
+	Then the user should be able to see standard output containing "wanted to load target 5600000 but only found up to"
+
+  Scenario: To query an non-existing Account
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "null"
+
+  Scenario: To query an existing account, incomplete command
+	Given the user has a pocket client
+	When the user runs the command "query account"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query an existing account txs
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query account-txs c7b6a62385f8999c0cb63ad9cb464ee730c597b1 0 1 0 0" against validator 0
+	Then the user should be able to see standard output containing "page_count"
+	And the user should be able to see standard output containing "total_txs"
+
+  Scenario: To query an existing account txs, wrong address
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "encoding/hex: invalid byte:"
+
+  Scenario: To query an existing account txs, incomplete command
+	Given the user has a pocket client
+	When the user runs the command "query account-txs"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query current acl
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query acl"
+	Then the user should be able to see standard output containing "/v1/query/acl"
+	And the user should be able to see standard output containing "gov/non_map_acl"
+
+  Scenario Outline: Unsuccessful Query of an App
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "Incorrect address length"
+
+  Scenario: To query an existing app from address, incomplete command
+	Given the user has a pocket client
+	When the user runs the command "query app"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query an existing app parameters
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query app-params"
+	Then the user should be able to see standard output containing "/v1/query/appparams"
+	And the user should be able to see standard output containing "base_relays_per_pokt"
+
+  Scenario Outline: Querying App Parameters, Invalid Height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command <command_with_bad_height>
+	Then the user should be able to see standard output containing <error_msg>
+	Examples:
+	  | command_with_bad_height  | error_msg             |
+	  | "query app-params 56000" | "error loading store" |
+	  | "query app-params ^"     | "invalid syntax"      |
+
+  Scenario: To query the list of existing apps
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query apps"
+	Then the user should be able to see standard output containing "/v1/query/apps"
+	And the user should be able to see standard output containing "result"
+
+  Scenario: To query an address balance in the network
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query balance {{address}} 0" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/balance"
+
+  Scenario: To query an address balance in the network, wrong address
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query balance {{address}} 0" against validator -1
+	Then the user should be able to see standard output containing "invalid byte"
+
+  Scenario: To query an address balance in the network, wrong height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard error containing "unknown shorthand flag: '3' in -3"
+
+  Scenario: To query an address balance in the network, incomplete command
+	Given the user has a pocket client
+	When the user runs the command "query balance"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query a block at height, wrong height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query block 0"
+	Then the user should be able to see standard output containing "height must be greater than 0"
+
+  Scenario: To query a block's at height, given an invalid number height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query block 1000"
+	Then the user should be able to see standard output containing "must be less than or equal to the current blockchain height"
+
+  Scenario: To query a block's at height, given an invalid char in height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query block 0-0"
+	Then the user should be able to see standard output containing "invalid syntax"
+
+  Scenario: To query block-txs at height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query block-txs 0 1 1 1"
+	Then the user should be able to see standard output containing "/v1/query/blocktxs"
+	And the user should be able to see standard output containing "total_txs"
+
+  Scenario: To query a block-txs at height, incomplete
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query dao owner
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query daoOwner" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/daoowner"
+
+  Scenario: To query the chains height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query height"
+	Then the user should be able to see standard output containing "height"
+
+  Scenario: To query an existing node from address
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	And the user should be able to see standard output containing "jailed"
+
+  Scenario: To query an existing node from address, wrong address
+	Given the user has a pocket client
+	When the user runs the command "query node tsst"
+	Then the user should be able to see standard output containing "invalid byte"
+
+  Scenario: To query an existing node from address, wrong address with only valid bytes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query an existing node claim, incomplete
+	Given the user has a pocket client
+	When the user runs the command "query node-claim"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query an existing node claims
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query node-claims {{address}} 0" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/nodeclaims"
+	And the user should be able to see standard output containing "page"
+	And the user should be able to see standard output containing "result"
+	And the user should be able to see standard output containing "total_pages"
+
+  Scenario: To query an existing node parameters
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query node-claims" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/nodeclaims"
+	Then the user should be able to see standard output containing "result"
+
+  Scenario: To query an existing node parameters, wrong address
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query node-claims {{address}}" against validator -1
+	Then the user should be able to see standard output containing "invalid byte"
+
+  Scenario: To query the list of existing nodes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query nodes" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/nodes"
+	Then the user should be able to see standard output containing "result"
+
+  Scenario: To query the list of existing nodes, Returns error when given wrong height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query nodes 60000"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To query the list of existing nodes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query nodes" against validator 0
+	Then the user should be able to see standard output containing "result"
+	Then the user should be able to see standard output containing "address"
+	Then the user should be able to see standard output containing "jailed"
+	Then the user should be able to see standard output containing "chains"
+	Then the user should be able to see standard output containing "service_url"
+
+  Scenario: To query single param
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query param pocketcore/MinimumNumberOfProofs 0" against validator 0
+	Then the user should be able to see standard output containing "param_key"
+	Then the user should be able to see standard output containing "pocketcore/MinimumNumberOfProofs"
+
+  Scenario: To query the list of params
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query params"
+	Then the user should be able to see standard output containing "param_key"
+	Then the user should be able to see standard output containing "param_value"
+
+  Scenario: To query the list of pocketcore params
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query pocket-params"
+	Then the user should be able to see standard output containing "claim_expiration"
+
+  Scenario: To query the list of pocketcore params, Return error with incorrect height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query pocket-params 9999999"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To query a specific node's signing info
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query signing-info {{address}} 0" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/signinginfo"
+	Then the user should be able to see standard output containing "result"
+
+  Scenario: To get supplies
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query supply"
+	Then the user should be able to see standard output containing "/v1/query/supply"
+	Then the user should be able to see standard output containing "app_staked"
+
+  Scenario: To get supplies, Returns error code with wrong height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query supply 999999"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To get existing supported networks
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query supported-networks"
```
emedrano2112, line 348: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "/v1/query/supportedchains"
+	Then the user should be able to see standard output containing "0011"
+
+  Scenario: To get existing supported networks, Returns an error code with invalid height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query supported-networks 999999"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario Outline: To get transactions based on the hash, wrong hash
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	  | "query tx 23197E4D4)(&09879F28F978A90627C7DFEAB64B4777AFCC24E2B9C3D72B4DADA22" | "invalid byte"          |
+
+  Scenario: To get transactions based on the hash, incomplete commands
+	Given the user has a pocket client
+	When the user runs the command "query tx"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query the latest upgrade
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query upgrade"
+	Then the user should be able to see standard output containing "/v1/query/upgrade"
+	Then the user should be able to see standard output containing "Height"
+
+  Scenario: To query the latest upgrade, invalid height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
emedrano2112: Change given, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	And the user should be able to see standard output containing "page"
+	And the user should be able to see standard output containing "result"
+	And the user should be able to see standard output containing "total_pages"
+
+  Scenario: To query an existing node parameters
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query node-claims" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/nodeclaims"
+	Then the user should be able to see standard output containing "result"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Given the user has a pocket client
+	When the user runs the command "query node-claims {{address}}" against validator -1
+	Then the user should be able to see standard output containing "invalid byte"
+
+  Scenario: To query the list of existing nodes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query nodes" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/nodes"
+	Then the user should be able to see standard output containing "result"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Given the user has a pocket client
+	When the user runs the command "query nodes 60000"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To query the list of existing nodes
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query nodes" against validator 0
+	Then the user should be able to see standard output containing "result"
+	Then the user should be able to see standard output containing "address"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "jailed"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "chains"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Then the user should be able to see standard output containing "service_url"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query single param
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query param pocketcore/MinimumNumberOfProofs 0" against validator 0
+	Then the user should be able to see standard output containing "param_key"
+	Then the user should be able to see standard output containing "pocketcore/MinimumNumberOfProofs"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query the list of params
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query params"
+	Then the user should be able to see standard output containing "param_key"
+	Then the user should be able to see standard output containing "param_value"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Given the user has a pocket client
+	When the user runs the command "query pocket-params 9999999"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To query a specific node's signing info
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query signing-info {{address}} 0" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/signinginfo"
+	Then the user should be able to see standard output containing "result"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To get supplies
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query supply"
+	Then the user should be able to see standard output containing "/v1/query/supply"
+	Then the user should be able to see standard output containing "app_staked"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Given the user has a pocket client
+	When the user runs the command "query supply 999999"
+	Then the user should be able to see standard output containing "error loading store"
+
+  Scenario: To get existing supported networks
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query supported-networks"
+	Then the user should be able to see standard output containing "/v1/query/supportedchains"
+	Then the user should be able to see standard output containing "0011"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	Given the user has a pocket client
+	When the user runs the command "query tx"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query the latest upgrade
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query upgrade"
+	Then the user should be able to see standard output containing "/v1/query/upgrade"
+	Then the user should be able to see standard output containing "Height"
```
emedrano2112: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query the latest upgrade, invalid height
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query upgrade 50000"
+	Then the user should be able to see standard output containing "error loading store"
```
emedrano2112, line 390: Change Then, by And. Returns the same. Refs: https://www.loom.com/share/81f7d41bcfca480b9f6e9b2ca5ca0651
https://specflow.org/gherkin/five-ways-to-make-gherkin-interesting/
https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
```
emedrano2112: Missing “exit clause” in all scenarios here (query.feature).

e2e/tests/accounts.feature
iajrz: These checks don't help accuracy. I will go through and revise them, but the accuracy of the test suite won't change any. The gherkin revision is meant to help determine whether the tests are accurate -- ie: sufficient to guarantee that the behavior they seem to be testing is actually correct.

e2e/tests/query.feature
iajrz: Great catch! Adding those.

e2e/tests/accounts.feature
ivan-pokt: seconded! We want to have no more than one Given statement for every scenario

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+	When the user runs the command "query app {{address}} 0" against validator <index>
+	Then the user should be able to see standard output containing <message>
+	Examples:
+	  | index | message                      |
+	  | -1    | "invalid byte"               |
+	  | 0     | "application does not exist" |
+
+  Scenario: To query a non-existing app from address, address has incorrect length but only valid bytes
+	Given the user has a pocket client
+	When the user runs the command "query app 4920ce1d787123456aeff366c79e8aa2 0"
```
ivan-pokt, line 93: Can we possibly make the app ID invalid as well (like giving it a value of 0 or 1 or -1, something impossible to be generated)? We don't want the possibility of a conflict is an app with the same id is created which would make the test fail (given that "4920ce1d787123456aeff366c79e8aa2" refers to the app ID"). Probability of that is low but with more usage, it is more likely - just to be safe.

e2e/tests/query.feature
```diff
@@ -0,0 +1,341 @@
...
+
+  Scenario: To query a block-txs at height, incomplete
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
+	When the user runs the command "query block-txs"
+	Then the user should be able to see standard error containing "Usage:"
+
+  Scenario: To query dao owner
+	Given the user is running the network "single_node_network"
+	Given the user has a pocket client
```
ivan-pokt: Let's try to use one Given statement per scenario. We can combine both statements into one Given statement for future usage but ideal to have one Given statement.

ivan-pokt (changes requested):
I think we need to readjust this to have singular Given statements per scenario and should not have a Then statements immediately following another Then statement. See line comments.

e2e/tests/query.feature
ivan-pokt: Seconded! We should not have multiple Then statements following each other.

e2e/tests/query.feature
```diff
@@ -0,0 +1,391 @@
...
+	Then the user should be able to see standard output containing "Usage:"
+	And the pocket client should have exited without error
+
+  Scenario: To query an existing account
+	Given the user is running the network "single_node_network"
+	And the user has a pocket client
+	When the user runs the command "query account {{address}}" against validator 0
+	Then the user should be able to see standard output containing "/v1/query/account"
+	And the user should be able to see standard output containing "address"
+	And the pocket client should have exited without error
```
AneudyM: The last And keyword could be replaced with "But should not exit with an error"

e2e/tests/query.feature
```diff
@@ -0,0 +1,391 @@
...
+
+  Scenario: Pocket Query Account address is invalid But Has valid Bytes
+	Given the user is running the network "single_node_network"
+	And the user has a pocket client
+	When the user runs the command "query account 2BBCA5DC9792C72AC3A2616910C4AAAA 0"
+	Then the user should be able to see standard output containing "Incorrect address length"
+	And the pocket client should have exited without error
+
+  Scenario: To query an existing account, wrong height
```
AneudyM, line 31: Scenario titles could be simplified to drive more accuracy and readability. Strive to be consistent with what you have in accounts.feature. It reads more naturally since you have a colon in between the Scenario keyword and the title. For example:

Instead of "Scenario to query an existing account", we could use "Scenario: Query an existing account with wrong height".

e2e/tests/query.feature
```diff
@@ -0,0 +1,391 @@
...
+	Then the user should be able to see standard error containing "Usage:"
+	And the pocket client should have exited with error
+
+  Scenario: To query current acl
+	Given the user is running the network "single_node_network"
+	And the user has a pocket client
+	When the user runs the command "query acl"
+	Then the user should be able to see standard output containing "/v1/query/acl"
+	And the user should be able to see standard output containing "gov/non_map_acl"
+	And the pocket client should have exited without error
```
AneudyM, line 78: NIT: We can also take advantage of indentation to improve readability. For example, the above scenario could be indented as follows:
```
Scenario: To query current acl
	Given the user is running the network "single_node_network"
	   And the user has a pocket client
	  When the user runs the command "query acl"
	  Then the user should be able to see standard output containing "/v1/query/acl"
	   And the user should be able to see standard output containing "gov/non_map_acl"
	   And the pocket client should have exited without error
```

e2e/tests/query.feature
```diff
@@ -0,0 +1,391 @@
...
+
+  Scenario: To get existing supported networks
+	Given the user is running the network "single_node_network"
+	And the user has a pocket client
+	When the user runs the command "query supported-networks"
+	Then the user should be able to see standard output containing "/v1/query/supportedchains"
+	And the user should be able to see standard output containing "0011"
+	And the pocket client should have exited without error
+
+  Scenario: To get existing supported networks, Returns an error code with invalid height
```
AneudyM, line 353: Confusing scenario. What is the acceptance criterion here?

AneudyM (changes requested):
The whole query.feature file needs to be refined. It is also too big. Consider using broader-concept steps to avoid repetitions or split them into smaller files if you are unwilling to use background.

e2e/tests/query.feature
Olshansk: @AneudyM I completely with you that this scenario, and it's title are confusing. When reviewing code, I usually try to use Github's "suggestion" feature so the reviewee can just accept/commit it or modify it if need be.; would appreciate your teams help in distributing the workload on this.

For example (I have no clue if this is clearer):
```suggestion
  Scenario: To verify that an invalid height cannot be quried from the list of supported-networks
```

Re the `scenario` - 
1. Are we verifying we can't query an invalid height OR we can't query an unsupported network?

Re the `Then` - `error loading store` makes me think the network is not supported.
2. Is that correct? 
3. If so, is there any way to make the source code (without comments) make that more explicit?

Re the `And`
4. Does this exit successfully because the client worked but the backing data is invalid?

e2e/tests/query.feature
iajrz:
```suggestion
	But the pocket client should have exited without error
```

e2e/tests/query.feature
iajrz: The style of indentation I am using is consistent with the style used by cucumber.io in the [Gherkin language reference ](https://cucumber.io/docs/gherkin/reference/)

e2e/tests/query.feature
iajrz: In this scenario the address has an incorrect length, so it's impossible to conflict with a correct app id.

e2e/tests/query.feature
iajrz:
```suggestion
	But the pocket client should have exited without error
```

e2e/launcher/node.go
iajrz: Decided against -- the first element would be a multi-line initialization, while the second element (the nil) would get lost lines below the return.

e2e/launcher/network.go
iajrz: I don't fancy the multi-line first value and single-line second value. I adopted the pointer, though. Might want to do away with the named returns. Have to think more about _that_.

e2e/tests/query.feature
iajrz: Titles lifted in the style of the [func tests](https://github.com/pokt-network/pocket-core-func-tests/blob/65fc1cc7a9093b0fcbaab0f52ed3f345827c4de8/tests/cli-commands/pocket-query-commands.feature#L51); will take a look at all scenario titles, thanks for pointing it out.

iajrz: As mentioned elsewhere -- the style was lifted from the original [functional test suite](https://github.com/pokt-network/pocket-core-func-tests/blob/65fc1cc7a9093b0fcbaab0f52ed3f345827c4de8/tests/cli-commands/pocket-query-commands.feature#L51). Will look at all scenario titles.

e2e/tests/query.feature
AneudyM: Yeah. Those are good points. Mmm..., looking at the validation of the scenario it might be trying to query a network with an invalid height. In that case, I think I would rewrite it as follows:

````
Scenario: Return error when querying supported networks with invalid height number
  Given a “single_node_network”
    And a valid pocket client
  When querying a supported network with invalid height “99999” number
  Then an “error loading store” message should be returned
    But the pocket client should not exit with an error status
````

Re point 4: 
That's why I am unsure of including the exit status steps here. What does the specification say about these exit statuses.

Olshansk:
@iajrz I updated the documentation, added a few comments and refactored a bit of the code. I'm in favour of adopting a similar pattern (with a simpler initial implementation) in V1 using gherkin.

Lmk if you see issue with any of my changes. If not, I can go ahead and merge it in.

cc @jessicadaugherty for visibility

Olshansk (approved):
Got an off-line 👍 from @iajrz on the changes I made so going to go ahead and merge this in:

Screenshot 2023-01-25 at 12 05 47 PM

PR #1525, opened 2023-02-18
minor: Use height in query account txs

alleneubank:
This PR Adds a new query parameter to the accounttxs endpoint to allow querying for account (send/received) txs starting at a certain height.

This is not a breaking change, as the new parameter is optional. If it is not provided, the endpoint will behave as it did before.

It is a performance improvement, as it allows the client to specify a height to start querying from, instead of having to query all txs and then filter them out.

You can try and verify by running the following commands:

**CURRENT_POKT_RPC_URL**= # replace with current rpc url
**POKT_RPC_WITH_HEIGHT_INDEX_URL**= # replace with new rpc url built from this branch

Tested using the SendNodes POPs address.

# pull hash and height of first and last tx

```sh
curl --request POST \
 --url http://CURRENT_POKT_RPC_URL/v1/query/accounttxs \
 --header 'Content-Type: application/json' \
 --data '{
"address": "cb6ff4204f8a93e89759c22a9e0f8896f8561379",
"page": 1,
"per_page": 10000,
"received": true,
"prove": false,
"order": "asc"
}' | jq '.txs | first, last | [.hash, .height]'

# example
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
#   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# 100 12.5M    0 12.5M  100   137  1364k     14  0:00:09  0:00:09 --:--:-- 4014k
# [
#   "1F6511B8E975C9189E4592C44CA0D8039B3861BC550EE96378D29BC43CE8E552",
#   86099
# ]
# [
#   "CCFB5D49009D37BFB3E472FBBD6E79791CB0B4E972765926074F521CF19A4812",
#   82654
# ]
```

# query new version with last height

```sh
curl --request POST \
  --url http://POKT_RPC_WITH_HEIGHT_INDEX_URL/v1/query/accounttxs \
  --header 'Content-Type: application/json' \
  --data '{
  "address": "cb6ff4204f8a93e89759c22a9e0f8896f8561379",
  "page": 1,
  "per_page": 10000,
  "received": true,
  "prove": false,
  "order": "asc",
  "height": 82654
}' | jq '.txs | first, last | [.hash, .height]'

# should match old version
#   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#                                  Dload  Upload   Total   Spent    Left  Speed
# 100 12.5M    0 12.5M  100   168  10.7M    143  0:00:01  0:00:01 --:--:-- 10.7M
# [
#   "1F6511B8E975C9189E4592C44CA0D8039B3861BC550EE96378D29BC43CE8E552",
#   86099
# ]
# [
#   "CCFB5D49009D37BFB3E472FBBD6E79791CB0B4E972765926074F521CF19A4812",
#   82654
# ]
```

# double check old version still matches

```sh
curl --request POST \
  --url https://POKT_RPC_WITH_HEIGHT_INDEX_URL/v1/query/accounttxs \
  --header 'Content-Type: application/json' \
  --data '{
  "address": "cb6ff4204f8a93e89759c22a9e0f8896f8561379",
  "page": 1,
  "per_page": 10000,
  "received": true,
  "prove": false,
  "order": "asc"
}' | jq '.txs | first, last | [.hash, .height]'

# should match old version
# [
#   "1F6511B8E975C9189E4592C44CA0D8039B3861BC550EE96378D29BC43CE8E552",
#   86099
# ]
# [
#   "CCFB5D49009D37BFB3E472FBBD6E79791CB0B4E972765926074F521CF19A4812",
#   82654
# ]
```

### DIFF
--- a/app/cmd/cli/query.go
+++ b/app/cmd/cli/query.go
@@ -114,10 +114,10 @@ var queryTx = &cobra.Command{
 }
 
 var queryAccountTxs = &cobra.Command{
-	Use:   "account-txs <address> <page> <per_page> <prove (true | false)> <received (true | false)> <order (asc | desc)>",
+	Use:   "account-txs <address> <page> <per_page> <prove (true | false)> <received (true | false)> <order (asc | desc)> <height>",
 	Short: "Get the transactions sent by the address, paginated by page and per_page",
 	Long:  `Retrieves the transactions sent by the address`,
-	Args:  cobra.RangeArgs(1, 6),
+	Args:  cobra.RangeArgs(1, 7),
 	Run: func(cmd *cobra.Command, args []string) {
 		app.InitConfig(datadir, tmNode, persistentPeers, seeds, remoteCLIURL)
 		page := 0
@@ -125,6 +125,7 @@ var queryAccountTxs = &cobra.Command{
 		prove := false
 		received := false
 		order := "desc"
+		height := int64(0)
 		if len(args) >= 2 {
 			parsedPage, err := strconv.Atoi(args[1])
 			if err == nil {
@@ -158,6 +159,12 @@ var queryAccountTxs = &cobra.Command{
 				order = "desc"
 			}
 		}
+		if len(args) >= 7 {
+			parsedHeight, err := strconv.ParseInt(args[6], 10, 64)
+			if err == nil {
+				height = parsedHeight
+			}
+		}
 		var err error
 		params := rpc.PaginateAddrParams{
 			Address:  args[0],
@@ -166,6 +173,7 @@ var queryAccountTxs = &cobra.Command{
 			Received: received,
 			Prove:    prove,
 			Sort:     order,
+			Height:   height,
 		}
 		j, err := json.Marshal(params)
 		if err != nil {
--- a/app/cmd/cli/query.go
+++ b/app/cmd/cli/query.go
@@ -125,7 +125,7 @@ var queryAccountTxs = &cobra.Command{
 		prove := false
 		received := false
 		order := "desc"
-		height := int64(0)
+		height := int64(0) // height=0 is ignored and implies latest height
 		if len(args) >= 2 {
 			parsedPage, err := strconv.Atoi(args[1])
 			if err == nil {
--- a/types/indexer.go
+++ b/types/indexer.go
@@ -131,30 +131,41 @@ func (t *TransactionIndexer) Get(hash []byte) (*types.TxResult, error) {
 // NOTE: Only supports op.Equal for hash, height, signer, or recipient, we only support op.Equal for simplicity and
 // optimization of our use case
 func (t *TransactionIndexer) Search(ctx context.Context, q *query.Query) (res []*types.TxResult, total int, err error) {
-	condition, err := q.Condition()
+	conditions, err := q.Conditions()
 	if err != nil {
-		return nil, 0, errors.Wrap(err, "error during parsing condition from query")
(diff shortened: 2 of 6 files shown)


app/cmd/cli/query.go
```diff
@@ -114,17 +114,18 @@ var queryTx = &cobra.Command{
...
-	Args:  cobra.RangeArgs(1, 6),
+	Args:  cobra.RangeArgs(1, 7),
 	Run: func(cmd *cobra.Command, args []string) {
 		app.InitConfig(datadir, tmNode, persistentPeers, seeds, remoteCLIURL)
 		page := 0
 		perPage := 0
 		prove := false
 		received := false
 		order := "desc"
+		height := int64(0)
```
Olshansk:
```suggestion
		height := int64(0) // height=0 is ignored and implies latest height
```

types/indexer.go
```diff
@@ -273,6 +298,14 @@ func prefixKeyForSigner(signer Address) []byte {
 	))
 }
 
+func prefixKeyForSignerAndHeight(signer Address, height int64) []byte {
+	return []byte(fmt.Sprintf("%s/%s/%s",
+		TxSignerKey,
+		signer,
+		elenEncoder.EncodeInt(int(height)), // totally safe right?
```
Olshansk: `// totally safe right?` -> Do you think we'll hit an overflow issue or something? If so, we can just force the user to pass an int into this function.

types/indexer.go
```diff
@@ -200,19 +211,33 @@ func (t *TransactionIndexer) heightQuery(condition query.Condition, pagination *
 	return t.getByPrefix(prefixKeyForHeight(height), pagination)
 }
 
-func (t *TransactionIndexer) signerQuery(condition query.Condition, pagination *query.Page) (res []*types.TxResult, total int, err error) {
-	signer, err := hex.DecodeString(condition.Operand.(string))
+func (t *TransactionIndexer) signerQuery(primaryCondition query.Condition, secondaryCondition query.Condition, pagination *query.Page) (res []*types.TxResult, total int, err error) {
```
Olshansk, line 219: Not actionable: I don't think we'll need any more features here, otherwise I would've suggested variadic params (for v1 stuff).

Olshansk (dismissed):
One minor comment and question but no blockers on my end. Massive improvement!

types/indexer.go
alleneubank: heh i think this was more of a joke to myself, I don't think we will be still babysitting v0 after 2^63-1 (or 2^31-1 on 32bit systems) blocks.

Olshansk: 💯

Olshansk: Let's remove the joke with a real comment (so future readers aren't confused as much as I do appreciate it), but otherwise lgtm

types/indexer.go
```diff
@@ -25,6 +26,10 @@ var (
 )
 
 const (
+	// Transaction(tx) Block height index key.
+	//
+	// Note: Block height is cast as int when querying using the tx height index due to the lexicographic encoder library,
+	// this is safe because v0 block height should never exceed 2^63-1 on 64-bit systems or 2^31-1 on 32-bit systems.
```
Olshansk, line 32: NIT (not actually): Block height should never exceed `2^31-1` on any 32+ bit system ;)

Olshansk (approved):

PR #1526, opened 2023-02-19
Probabilistic Proofs - Model Claim and Stake Data for Parameter Thresholds

### LINKED ISSUE #1523: [Guardrails] Probabilistic Proofs - Model Claim and Stake Data for Parameter Thresholds
## Objective

Build a data model using claims and stakes to set a ProbabilityOfProofRequest parameter value that proves with 99% confidence that servicers cannot claim more POKT than they risk burning. 

This model should set the parameter value for ProofRequiredThreshold as well by using stake data against ProbabilityOfProofRequest which will enable us to both validate the proposed solution and provide the foundations for calculating how much additional scale can be achieved with these changes. 

## Origin Document

The V0 growth budget has been documented extensively:

- [Project Scale V0](https://docs.google.com/document/d/1uBomaVieGAjsyHeqSlwmqOyPOln1CemVlWXZjQXUMRY/edit)
- [Growth Budget](https://docs.google.com/document/d/1QcsPfhj636zBazw2jAay8H6gdBKXcIMmohY0o321hhQ/edit)
- [Otto’s Guide to Pocket v0](https://drive.google.com/drive/u/0/folders/1t-t0n7uMyvx-wBraDWBKVRLh532ZfD-c)

These documents and measurements have resulted in a proposed scaling solution called Probabilistic Proofs:

- [Probabilistic Proofs Proposal
](https://www.notion.so/pocketnetwork/Idea-Probabilistic-Proofs-f3fa87447f1c4c63b38b1ff46c049f67#d1f7699135fd438cb2251fa11a91f8cb)

## Goals

- Determine the values for ProbabilityOfProofRequest and ProofRequiredThreshold parameters
- Test and validate the values using on chain data (claims and stakes)
- Collect data on % of block space consumed by proofs and claims to determine new growth budget

## Deliverable

- [ ] Build a negative binomial distribution model and determine the initial value for ProbabilityOfProofRequest parameter using a minimum of 1 week of [claims data](https://drive.google.com/drive/u/1/folders/1pSaLblxMTTl3CZGMoiHbVriY0wCbQJSC)
- [ ] Iterate ProbabilityOfProofRequest value over [stake data](https://docs.google.com/spreadsheets/d/17n04vwvHpMWtyp3E7_VfmktDTc-DrP2Lw9br2fmleAI/edit#gid=0) and determine the initial value for ProofRequiredThreshold

## Non-goals / Non-deliverables

- Implementing the scaling solution 

## General issue deliverables

- [ ] Update the appropriate CHANGELOG
- [ ] Update any relevant READMEs (local and/or global)

---

**Creator**: @jessicadaugherty
**Co-owner**: @Olshansk

Olshansk:
Pocket Network V0 has a theoretical cap of 3B average daily relays. This is primarily caused by the bloat of Proofs within a Block, which is part of Pocket's fundamental Claim & Proof relay validation lifecycle. This document outlines a theoretical proposal to help scale Pocket Network V0 by several magnitudes through a Probabilistic Proof mechanism. This reduces the amount of space taken up by Proofs within a Block, and does not cause a linear increase in the network's storage requirements as the number of relays grow.

Fixes #1523

### DIFF
--- a/docs/proposals/probabilistic_proofs.md
+++ b/docs/proposals/probabilistic_proofs.md
@@ -50,12 +50,11 @@ Three new governance parameters will need to be added:
 ### Parameter Usage
 
 $$
-Probably(ProofRequired)=\left\{
-\begin{array}{ll}
-ProbabilityOfProofRequest &\text{if } Claim < ProofRequiredThreshold  \\
-1.0 &\text{if } Claim >= ProofRequiredThreshold.
-\end{array}
-\right.
+Probably(ProofRequired) =
+  \begin{cases}
+    ProbabilityOfProofRequest &\text{if } Claim < ProofRequiredThreshold  \\
+    1 &\text{if } Claim >= ProofRequiredThreshold.
+  \end{cases}
 $$
 
 ### Flow
@@ -125,6 +124,7 @@ The definition for success is taken from the Network's point of view.
   - Servicer submits a false claim and gets away with it
   - Servicer submits a true claim, but fails to prove it
 
+<!--
 The types of questions we can ask are:
 
 - What is the stopping condition?
@@ -134,6 +134,7 @@ The types of questions we can ask are:
 - How many success until N failures?
 - How much scalability does the network need?
 - With what likelihood should an attacker get away?
+-->
 
 ### Example
 
@@ -151,11 +152,12 @@ Since each claim is independent, an attacker would never submit a `Claim` exceed
 A [Geometric_distribution](https://en.wikipedia.org/wiki/Geometric_distribution) is used to identify the probability of `k` failures (sample space containing an attacker getting away) until a single success (an attacker is caught).
 
 $$ p = ProofRequestProbability $$
-$$ q = 1 - p $$
-$$ Pr(X=k) = (1-p)^{k-1}p $$
-$$ k = \frac{ln(\frac{Pr(X=k)}{p})}{ln(1-p)} + 1 $$
 
-#### Model Visualization
+$$ q = 1 - p $$
+
+$$ Pr(X=k) = (1-p)^{k-1}p $$
+
+$$ k = \frac{ln(\frac{Pr(X=k)}{p})}{ln(1-p)} + 1 $$
 
 ![download](https://user-images.githubusercontent.com/1892194/220803154-90dcdd6b-8141-40d2-9cca-ed27a995fcfb.png)
 
@@ -172,7 +174,9 @@ _TODO(olshansky): Look at claims data to figure out the value for ProofRequireme
 `Pr(X=k) > 0` must be greater than 0 while minimizing `k` as much as possible, so it is selected at approximately `0.01.`.
 
 $$ k = \frac{ln(\frac{Pr(X=k)}{p})}{ln(1-p)} + 1 $$
+
 $$ k = \frac{ln(\frac{0.01}{0.25})}{ln(1-0.25)} + 1 $$
+
 $$ k ≈ 12.19 $$
 
 Selecting `k = 12` implies that there is an `~1%` of 12 failures (includes the sample space where an attacker gets away) until a single success (an attacker is caught). To deter this behaviour, the `ProofMissingPenalty` should be selected as 12 times `ProofRequirementThreshold`.
--- a/docs/proposals/probabilistic_proofs.md
+++ b/docs/proposals/probabilistic_proofs.md
@@ -10,15 +10,15 @@ This is a specification & proposal that will be submitted to [forum.pokt.network
 
 **tl;dr Values Selected**
 
-- `ProofRequirementThreshold` = `2σ` above network claim Claim `μ`
-- `ProofRequestProbability` = `0.25`
-- `k` (num failures) = `16`
-- `Pr(X<=k)` = `0.99`
-- `ProofMissingPenalty` = `ProofRequirementThreshold * k`
+- `Pr(X<=k)` = `0.99`; selected manually to maintain 2 nine protocol safety
+- `k` (num failures) = `16`; computed using a cumulative geometric probability distribution
+- `ProofRequestProbability` = `0.25`; selected manually to scale the network by 4x
+- `ProofRequirementThreshold` = `20 POKT`; selected to a value that is above p95 of all POKT claims
+- `ProofMissingPenalty` = `320 POKT`; calculated via `ProofRequirementThreshold * k` to deter malicious behaviour
 
 **The question being answered by the distribution**: What is the probability of the network (i.e. the protocol) failing `k` times or less(i.e. handling a normal claim or not catching an attacker) until a single success (i.e. catching an attacker).
 
-**Answer**: Selecting `k = 16` implies that `99%` of the time, an attacker will get a penalty of `BurnForFailedClaimSubmission`, making it not worthwhile to take the risk.
+**Answer**: Selecting `k = 16` and `ProofRequirementThreshold = 20 POKT` implies that if an attacker continues submitting claims for `19.99 POKT` or less, they will get caught `99.99%` of the time, and will be penalized for `320 POKT`.
 
 ## Table of Contents <!-- omit in toc -->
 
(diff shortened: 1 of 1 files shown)


docs/proposals/probabilistic_proofs.md
```diff
@@ -0,0 +1,241 @@
...
+
+`Pr(X=k) > 0` must be greater than 0 while minimizing `k` as much as possible, so it is selected at approximately `0.01.`.
+
+$$ k = \frac{ln(\frac{Pr(X=k)}{p})}{ln(1-p)} + 1 $$
+
+$$ k = \frac{ln(\frac{0.01}{0.25})}{ln(1-0.25)} + 1 $$
+
+$$ k ≈ 12.19 $$
+
+Selecting `k = 12` implies that there is an `~1%` of 12 failures (includes the sample space where an attacker gets away) until a single success (an attacker is caught). To deter this behaviour, the `ProofMissingPenalty` should be selected as 12 times `ProofRequirementThreshold`.
```
RawthiL: 1% is the chance of doing exactly k=12 failures before a success, what you really want to know is the chance of any attacker getting 12 or more failures before a success. You should use the CDF to get this value, with p=025 and k=12 the CDF will tell you that 96.86% of the cases will fall here. If you want to get the 99% of the cases you should get it from the CDF:
0.99 <= 1-((1-p)**k)  ; p=0.25
-> k =  np.ceil( np.log(1.-0.99)/np.log(1.-p) ) = 17

docs/proposals/probabilistic_proofs.md
```diff
@@ -0,0 +1,241 @@
...
+
+$$ p = ProofRequestProbability $$
+
+$$ q = 1 - p $$
+
+$$ Pr(X=k) = (1-p)^{k-1}p $$
+
+$$ k = \frac{ln(\frac{Pr(X=k)}{p})}{ln(1-p)} + 1 $$
+
+![download](https://user-images.githubusercontent.com/1892194/220803154-90dcdd6b-8141-40d2-9cca-ed27a995fcfb.png)
```
RawthiL: This graph violates the support of the PDF. While the formula that you derived is not wrong, the grpah shows a support over all possible values of P, which results in real-valued Ks (that are not in the support of K={1,2,3,4,...}).
I would suggest to keep this graph discrete, showing only values for K that are in the support of Pr().

RawthiL: imagen
Only the dots are valid, seeing negative Ks in the Y/axis make my eyes hurt haha

docs/proposals/probabilistic_proofs.md
```diff
@@ -0,0 +1,241 @@
...
+
+Selecting `k = 12` implies that there is an `~1%` of 12 failures (includes the sample space where an attacker gets away) until a single success (an attacker is caught). To deter this behaviour, the `ProofMissingPenalty` should be selected as 12 times `ProofRequirementThreshold`.
+
+To answer this question, we need to:
+
+- Pick `p` - variable for scaling the network (e.g. 0.25 => 4x network scale)
+- Select `Pr(X=k)` - the likelihood of `k` failures
+- Introduce `BurnForFailedClaimSubmission` - penalty for getting caught in falsifying a claim (or failing to submit a real claim)
+
+`ProofRequirementThreshold` - Should be set to a value whereby that that is `2σ` greater than the mean claim. This will mean that the other `97.3%` of claims will be subject to `ProofRequestProbability`.
```
RawthiL: You are supposing they have a normal distribution, lets hope that's the case, anyway we will know once I give you the data.

docs/proposals/probabilistic_proofs.md
Olshansk: Great idea!! Totally on the same page regarding my 👀 hurting :)

I've updated the code in the Appendix (it's commented out in markdown) for future readers

Screenshot 2023-02-23 at 6 21 33 PM

docs/proposals/probabilistic_proofs.md
Olshansk: That was definitely an assumption I was making. I saw your comment in #1523 and will follow up there.

docs/proposals/probabilistic_proofs.md
Olshansk: Thanks for the feedback @RawthiL!

100% agree on using a CDF instead of a PDF here.

The most difficult part for me is trying to figure out exactly what a success/failure is and be very clear about what question we're answering and why.

Rather than `any attacker getting 12 or more failures before a success`, what we'll be modelling with a CDF is `the protocol failing 12 times or less (i.e. not catching an attacker, or handling a normal claim) until a single success (i.e. catching an attacker)`

Does that make sense?

Screenshot 2023-02-23 at 7 37 54 PM

Olshansk: I think I'm comfortable with this formulation now

Screenshot 2023-02-23 at 8 08 19 PM

RawthiL: > the protocol failing 12 times or less (i.e. not catching an attacker, or handling a normal claim) until a single success (i.e. catching an attacker)

The CDF will tell us the chance that the protocol successfully catches an attacker in less than 12 consecutive failures. So, it makes sense but I would not word it as the protocol "failing" K times, it is just "trusting". So maybe it could be worded like this?:
> the protocol successfully catching an attacker (or challenging a normal claim) before 12 (or more) consecutive unchallenged claims (i.e. failing catching an attacker or correctly trusting the node).

The graph is OK now.

In the pictor I see that you updated the K to use the CDF, this is OK but you still include the paragraph that models the claims as normal distributions (thats not the case as we observed).

Olshansk: I made the changes below based on your feedback. Let me know if its clearer now

---

> **The question being answered by the distribution**: What is the probability of the protocol trusting (i.e. failing) `k` Claims or less (i.e. handling a normal claim or not catching an attacker) until a single penalty enforcement (i.e. successfully catching an attacker).
> 
> **Answer**: Selecting `k = 16` and `ProofRequirementThreshold = 20 POKT` implies that if an attacker continues submitting claims for `19.99 POKT` or less, they will get caught `99%` of the time, and will be penalized for `320 POKT`.

---


> `ProofRequirementThreshold` should be as small as possible so that most such that most Claims for into the probabilistic bucket, while also balancing out penalties that may be too large for faulty honest Servicers. Ideally, it should be selected to be `2σ` above the Claim `μ` such that `97.3%` fall into the `ProofRequestProbability` part of the piecewise function. However, as seen in the Appendix, the POKT Claim distribution does not follow a normal distribution. Instead, 20 POKT was selected since it is greater than `p95` of POKT claim using the data collected.

RawthiL: Great!

docs/proposals/probabilistic_proofs.md
```diff
@@ -0,0 +1,322 @@
...
+  - [Claim Data](#claim-data)
+  - [Python Code](#python-code)
+    - [Python Code - Geometric PDF](#python-code---geometric-pdf)
+    - [Python Code - Geometric CDF](#python-code---geometric-cdf)
+
+## Summary
+
+The number of relays in Pocket Network (V0) is limited by the amount of block space utilized by Claim Proofs. The estimations done [here](https://docs.google.com/document/d/1QcsPfhj636zBazw2jAay8H6gdBKXcIMmohY0o321hhQ/edit) approximated it to be ~3B/day.
+
+Several solutions were proposed [in this document](https://docs.google.com/document/d/1uBomaVieGAjsyHeqSlwmqOyPOln1CemVlWXZjQXUMRY/edit). This proposal outlines an alternate solution: **Probabilistic Proofs**.
```
RawthiL, line 53: The document is not public I think

RawthiL (approved):
I think it is OK, only a comment on a link that I was not able to follow.

docs/proposals/probabilistic_proofs.md
Olshansk: Should be public now!

Issue #1478, opened 2022-09-23
[BUG REPORT] Pocket nodes getting stuck during sync


bvanderlugt:
**Describe the bug**
The node runner community is reporting stuck nodes. Also, on restart the db is corrupt and will not sync

**To Reproduce**
Not sure how to repro. Node runners are reporting the chain stopped syncing around block 71479, also 68389
.
**Expected behavior**
Pocket nodes should stay syncd

**Screenshots**

Snippets from Pocket Network #node-runners chat:
 
> yup, Ive got a node stuck at 71479 and another stuck at 68389, for the latter, when I tried restarting pocket, it crashed and corrupted the blockchain data

```
panic: enterPrecommit: +2/3 prevoted for an invalid block: wrong Block.Header.AppHash.  Expected 7376BF29ECCFC41FDEEDB1CDD20A14CE9523C98FAEA3F0ECCB71720681CABB47, got E12EC5CF1308FA37148A893
B46DEE30A44AB7B81DD3D2ABA91515E149F92637C
```

> On nodes that got stuck I see lots of these errors:

```
I[2022-09-23|11:27:06.579] Could not check tx                           module=mempool tx=67FA200C97C64C49B1F74F431D77CB11F6089E26E2363707526FF0215291BDAE err="mempool is full: number of txs 9000 (max: 9000), total txs bytes 7736607 (max: 1073741824)"
```

> This is what I see before the service stops.

```
Sep 23 10:33:16 b2.dabble.run pocket[378613]: panic: enterPrecommit: +2/3 prevoted for an invalid block: wrong Block.Header.AppHash.  Expected 88BCA70707E668595338EE7F87489618C01E128BA24582B9CA9F933D7667C9D6, got 682E10813F8B9F00B6F9AF10EBEF5D3CB4303459514C4BBD09D54D9D914A8F22
```

**Operating System or Platform:**
Please indicate the platform(s) where the bug is happening.

**Additional context**
Community thread is [here](https://discord.com/channels/553741558869131266/564836328202567725/1022894079383703618) and sprinkled throughout the #node-runners channel September 22nd and 23rd.  

This task is tracked internally with tickets T-1441, T-14354

Olshansk:
@oten91 Do you have any insight on:
1. Why this happens?
2. How to avoid it in the past?
3. Workarounds in the meantime?

oten91:
1- Hey , Will try to summarize what I observed yesterday and some anecdotal information.

[...]

PR #1529, opened 2023-03-01
Clear the global session cache when a node is unjailed

msmania:
This is a fix for a consensus failure with `wrong Block.Header.AppHash`, which is probably the same issue as #1478.

### Issue description:

For a node to be selected for a session, it must be valid not only at the start of a session but also at the end of a session.  If a node is edit-staked, jailed, or unjailed in the middle of a session, a servicer set of the session is changed.

When a node serves relays, on the other hand, it stores session information into the global session cache.  If the session's servicer set is changed, the node needs to update the cached session.  Otherwise the node accepts a claim that should be rejected, or rejects a claim that should be accepted.  As a result, the node ends up with a consensus failure with `wrong Block.Header.AppHash`.

### Root cause and Fix:

The root cause is simple.  We have a call to `ClearSessionCache` in `EditStakeValidator`, `LegacyForceValidatorUnstake`, `ForceValidatorUnstake`, and `JailValidator`, but not in `UnjailValidator`.  The proposed fix is to add the call in `UnjailValidator` as well.

### DIFF
--- a/x/pocketcore/types/service_test.go
+++ b/x/pocketcore/types/service_test.go
@@ -2,17 +2,14 @@ package types
 
 import (
 	"encoding/hex"
-	"github.com/pokt-network/pocket-core/codec"
-	types2 "github.com/pokt-network/pocket-core/codec/types"
-	"github.com/pokt-network/pocket-core/crypto"
-	exported2 "github.com/pokt-network/pocket-core/x/apps/exported"
-	"github.com/pokt-network/pocket-core/x/auth"
-	"github.com/pokt-network/pocket-core/x/gov"
 	"reflect"
 	"testing"
 	"time"
 
+	"github.com/pokt-network/pocket-core/codec"
+	"github.com/pokt-network/pocket-core/crypto"
 	sdk "github.com/pokt-network/pocket-core/types"
+	exported2 "github.com/pokt-network/pocket-core/x/apps/exported"
 	appsType "github.com/pokt-network/pocket-core/x/apps/types"
 	"github.com/pokt-network/pocket-core/x/nodes/exported"
 	nodesTypes "github.com/pokt-network/pocket-core/x/nodes/types"
@@ -331,7 +328,7 @@ type MockPosKeeper struct {
 type MockPocketKeeper struct{}
 
 func (m MockPocketKeeper) Codec() *codec.Codec {
-	return makeTestCodec()
+	panic("implement me")
 }
 
 func (m MockPocketKeeper) SessionNodeCount(ctx sdk.Ctx) (res int64) {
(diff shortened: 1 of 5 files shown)


Olshansk:
Thanks @msmania. Logically, this makes sense to me and I've confirmed that we call `ClearSessionCache` in all other relevant locations.

My only concern is that I don't have the resources right now to validate this and the testing coverage isn't good enough either.

Ccing @PoktBlade for a 2nd opinion if this is good to just 🚢 ?

nodiesBlade:
Good find, you are correct that we should be clearing the session cache when unjailing, and this could cause consensus issues due to caching causing nondeterministic behavior.

Most validators and servicers should not have the session cached until the claim/proof lifecycle. They would only if the validator/servicer is in the same app session.

Furthermore, sessions are cleared in general based on certain transactions. TLDR - we're riding off luck with nondeterministic behavior

Where does the risk lie?:
Given LeanPocket allows nodes to be stacked onto each other, validators will share the same session cache, and such could result in a consensus failure with enough luck if 1/n stacked validators are paired with the jailed servicer that submits a claim. This is because then all stacked validators have the 'wrong' session cached, and unless there is another transaction that clears the session cache, the stacked validators will likely all disagree with the other validators in the network.

Even without LP, the probability of it happening is non-zero.

I am unsure if we should merge this change, or if we should consider a consensus-breaking change for this otherwise we will have some validators clearing the cache and some validators not clearing the cache, which could cause more issues.

nodiesBlade:
If we want to play it 100% safe, then an upgrade height would be ideal, and release it ASAP.

nodiesBlade:
I don't mean to distract this conversation, but I added a sample implementation here:

https://github.com/pokt-network/pocket-core/pull/1532

It compiles, that's as far as I've tested it.

Olshansk:
@msmania I added a few comments to the PR in #1532 that has the flag. Do you mind copying those changes into this PR?

Olshansk (changes requested):
I added a few comments to the PR in #1532 that has the flag. Do you mind copying those changes into this PR?

x/pocketcore/types/service_test.go
```diff
@@ -2,17 +2,14 @@ package types
 
 import (
 	"encoding/hex"
-	"github.com/pokt-network/pocket-core/codec"
-	types2 "github.com/pokt-network/pocket-core/codec/types"
-	"github.com/pokt-network/pocket-core/crypto"
-	exported2 "github.com/pokt-network/pocket-core/x/apps/exported"
-	"github.com/pokt-network/pocket-core/x/auth"
-	"github.com/pokt-network/pocket-core/x/gov"
```
msmania, line 10: Why do we need this change?
x/gov/module.go importing x/pocketcore/types caused a cyclic import in the x/pocketcore/types package because this file imported x/gov.

It seems we don't need `makeTestCodec()` at all.  I verified all tests under x/pocketcore/types and x/pocketcore/keeper passed.

msmania:
> I added a few comments to the PR in #1532 that has the flag. Do you mind copying those changes into this PR?

Done!

x/pocketcore/types/service_test.go
Olshansk: Sounds reasonable to me

Olshansk (approved):

PR #1532, opened 2023-03-08
Placeholder for clearing session cache for unjailed vals with upgrade height

nodiesBlade:
Disclaimer: Never done an upgrade height transaction before, please do a test in local net, testnet, and so on for this upgrade

Addresses concerns in #1529 by adding an upgrade height

### DIFF
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -50,6 +50,7 @@ const (
 	BlockSizeModifyKey           = "BLOCK"
 	RSCALKey                     = "RSCAL"
 	VEDITKey                     = "VEDIT"
+	ClearUnjailedValSessionKey   = "CRVAL"
 )
 
 func GetCodecUpgradeHeight() int64 {
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -282,6 +282,16 @@ func (cdc *Codec) IsOnNamedFeatureActivationHeight(height int64, key string) boo
 	return UpgradeFeatureMap[key] != 0 && height == UpgradeFeatureMap[key]
 }
 
+func (cdc *Codec) IsBetweenNamedFeatureActivationHeight(height int64, key string, tolerance int64) bool {
+	upgradeHeight := UpgradeFeatureMap[key]
+	if upgradeHeight == 0 {
+		return false
+	}
+	minHeight := upgradeHeight - tolerance
+	maxHeight := upgradeHeight + tolerance
+	return height >= minHeight && height <= maxHeight
+}
+
 // Upgrade Utils for feature map
 
 // SliceToExistingMap merge slice to existing map
--- a/codec/codec_test.go
+++ b/codec/codec_test.go
@@ -1,0 +1,25 @@
+package codec
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestIsBetweenNamedFeatureActivationHeight(t *testing.T) {
+	tolerance := int64(5)
+	codecUpgradeKey := ClearUnjailedValSessionKey
+
+	// have to init global map / not mock friendly.
+	UpgradeFeatureMap = map[string]int64{codecUpgradeKey: 20000}
+	codec := NewCodec(nil)
+
+	// Test zero values / out of bounds
+	assert.False(t, codec.IsBetweenNamedFeatureActivationHeight(0, codecUpgradeKey, tolerance))
+	assert.False(t, codec.IsBetweenNamedFeatureActivationHeight(19994, codecUpgradeKey, tolerance))
+	assert.False(t, codec.IsBetweenNamedFeatureActivationHeight(20006, codecUpgradeKey, tolerance))
+
+	// Test in bounds
+	assert.True(t, codec.IsBetweenNamedFeatureActivationHeight(19995, codecUpgradeKey, tolerance))
+	assert.True(t, codec.IsBetweenNamedFeatureActivationHeight(20005, codecUpgradeKey, tolerance))
+	assert.True(t, codec.IsBetweenNamedFeatureActivationHeight(20000, codecUpgradeKey, tolerance))
+}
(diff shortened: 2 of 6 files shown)


x/gov/module.go
```diff
@@ -120,6 +121,12 @@ func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
 
 	ActivateAdditionalParametersACL(ctx, am)
 
+	// On this upgrade height, Pocket Core will start clearing the global session cache to prevent any non-deterministic cache consistency issues.
+	// This code will ensure it is cleared on the upgrade height for a fresh start.
+	if am.keeper.GetCodec().IsOnNamedFeatureActivationHeight(ctx.BlockHeight(), codec.ClearUnjailedValSessionKey) {
```
Olshansk: I'd probably add some tolerance (+-2 blocks) just to be safe. 

Whatever hit we take in performance during these few blocks in `BeginBlock` isn't worth the lack of testing/infrastructure we have to be confident that no caching issue will break anything.

Olshansk (changes requested):

Olshansk:
@msmania PTAL when you have a chance as another pair of 👀

x/gov/module.go
nodiesBlade: I'm not sure if this is necessary to clear the cache so many times, but regardless added it along with some unit tests.

msmania:
Sorry for being late, and thank you @PoktBlade for creating this.  I totally agree with your comment https://github.com/pokt-network/pocket-core/issues/1529#issuecomment-1459082393.  It's technically possible that the network reaches wrong consensus if the majority of validators end up having unupdated session cache.

This patch looks good to me.

x/nodes/keeper/valStateChanges.go
```diff
@@ -652,6 +652,12 @@ func (k Keeper) UnjailValidator(ctx sdk.Ctx, addr sdk.Address) {
 		k.Logger(ctx).Error(fmt.Sprintf("cannot unjail already unjailed validator, validator: %v at height %d\n", validator, ctx.BlockHeight()))
 		return
 	}
+
+	// After or on this upgrade height, we will start clearing the global session cache to prevent any non-deterministic cache consistency issues.
+	if k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.ClearUnjailedValSessionKey) {
```
Olshansk, line 657:
```suggestion
	// See #1529 for more details
	if k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.ClearUnjailedValSessionKey) {
```

x/gov/module.go
```diff
@@ -161,6 +168,7 @@ func ActivateAdditionalParametersACL(ctx sdk.Ctx, am AppModule) {
 		params.ACL.SetOwner(types.NewACLKey(types.NodesSubspace, "ServicerStakeFloorMultiplierExponent"), am.keeper.GetDAOOwner(ctx))
 		am.keeper.SetParams(ctx, params)
 	}
+
```
Olshansk, line 171:
```suggestion
```

codec/codec.go
```diff
@@ -281,6 +282,16 @@ func (cdc *Codec) IsOnNamedFeatureActivationHeight(height int64, key string) boo
 	return UpgradeFeatureMap[key] != 0 && height == UpgradeFeatureMap[key]
 }
 
+func (cdc *Codec) IsBetweenNamedFeatureActivationHeight(height int64, key string, tolerance int64) bool {
```
Olshansk, line 285:
```suggestion
// IsOnNamedFeatureActivationHeightWithTolerance is used to enable certain business logic within some tolerance
// (i.e. only a few blocks) of feature activation to have more confidence in the feature's release and avoid
// non-deterministic or hard-to-predict behaviour.
func (cdc *Codec) IsOnNamedFeatureActivationHeightWithTolerance(height int64, featureKey string, tolerance int64) bool {
```

Olshansk: Make sure to change the test name et al too

codec/codec_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestIsBetweenNamedFeatureActivationHeight(t *testing.T) {
+	tolerance := int64(5)
+	codecUpgradeKey := ClearUnjailedValSessionKey
+
+	// have to init global map / not mock friendly.
```
Olshansk, line 12: 1. Add a comment at the top that this should never be run in parallel
2. Capture the value of `UpgradeFeatureMap` before the test
3. Run `t.Cleanup(...)` where you reassign the value it had pre-test

x/gov/module.go
```diff
@@ -120,6 +121,12 @@ func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
 
 	ActivateAdditionalParametersACL(ctx, am)
 
+	// On this upgrade height, Pocket Core will start clearing the global session cache to prevent any non-deterministic cache consistency issues.
```
Olshansk, line 124: s/On/Around

x/gov/module.go
```diff
@@ -120,6 +121,12 @@ func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
...
+	// This code will ensure it is cleared on the upgrade height for a fresh start.
+	if am.keeper.GetCodec().IsBetweenNamedFeatureActivationHeight(ctx.BlockHeight(), codec.ClearUnjailedValSessionKey, 2) {
```
Olshansk, line 126: move magic number into a const with a descriptive name

Olshansk (changes requested):

Olshansk:
@PoktBlade Okay to close this one out in favor of #1529 so we can finish it off there?

Olshansk:
Closing it seeing how #1529 was merged in.

PR #1533, opened 2023-03-12
Fix TestValidatorStateChange_EditAndValidateStakeValidator

msmania:
The test `TestValidatorStateChange_EditAndValidateStakeValidator` failed because `got.OutputAddress` was `nil` but `tt.want.OutputAddress` had a valid address.  Since this test is to test the behavior before NCUST, we make sure the fetched output address is nil and the rest of fields are the same as `tt.want`.

The behavior after NCUST is tested through the test `TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgrade`, which passes without failures.

### DIFF
--- a/x/nodes/keeper/valStateChanges_test.go
+++ b/x/nodes/keeper/valStateChanges_test.go
@@ -210,6 +210,9 @@ func TestValidatorStateChange_EditAndValidateStakeValidator(t *testing.T) {
 			tt.want.Status = sdk.Staked
 			// see if the changes stuck
 			got, _ := keeper.GetValidator(context, tt.origApp.Address)
+			assert.Nil(t, got.OutputAddress, "OutputAddress was set before NCUST update")
+			// Manually updated `got` to account for post NCUST updates
+			got.OutputAddress = tt.want.OutputAddress
 			if !got.Equals(tt.want) {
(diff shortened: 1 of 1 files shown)


x/nodes/keeper/valStateChanges_test.go
```diff
@@ -210,6 +210,8 @@ func TestValidatorStateChange_EditAndValidateStakeValidator(t *testing.T) {
 			tt.want.Status = sdk.Staked
 			// see if the changes stuck
 			got, _ := keeper.GetValidator(context, tt.origApp.Address)
+			assert.Nil(t, got.OutputAddress, "OutputAddress was set before NCUST update")
+			got.OutputAddress = tt.want.OutputAddress
```
Olshansk, line 215:
```suggestion
                         // Manually updated `got` to account for post NCUST updates
			got.OutputAddress = tt.want.OutputAddress
```

Olshansk (changes requested):

x/nodes/keeper/valStateChanges_test.go
msmania: Added.

Olshansk (approved):

Issue #1499, opened 2022-11-13
[REQUEST FOR PROPOSALS] Allow changing the output address


nodiesBlade:
**Please explain the goal of the RFP.**

Allow users to change their non-custodial wallet address

**Please provide a justification for your RFP.**

Couple reasons

- Non custodial was designed back in 2021 where upstaking (per PIP-22) did not exist. Legacy accounts cannot upstake without deciding on whether they want to be non custodial or custodial **forever**. This was not a planned side effect.
- Insurance purposes - insurance providers (coincover) have not decided whats their policy around non custodial and the best security practices for this (1 to ..n output wallets?). Node runners are now at a dilemnia on whether they should upstake a node and forever be custodial, or set to non custodial without following guidelines.
- In the event the non-custodial account private keys get exposed, there is no way to mitigate the issue.
- In general, keys should be rotatable.

**Please provide success criteria for proposals responding to this RFP.**
- Users who have access to the output address private keys can change the output address through another stake tx.

**Add aditional context regarding this RFP.**

I did some initial scoping for this - it seems to be a small change in general. 

https://github.com/pokt-network/pocket-core/blob/staging/x/nodes/keeper/valStateChanges.go#L233, here we will need to add a clause to check if the signer address was equal to the old output address

Most of the leg work will be in testing.

Olshansk:
Completed as part of the latest release: https://github.com/pokt-network/pocket-core/releases/tag/RC-0.10.3

PR #1534, opened 2023-03-13
Allow the output address owner to change the output address via MsgStake

msmania:
This patch allows the owner of the output address of a staked node to change the output address to another via the existing stake transaction.  The same idea was proposed as #1499.

### Rationale:

This feature gives users more flexibility.  One of the possible use cases is that a node runner can convert custodial stake nodes into non-custodial stake nodes without unstaking.  Or a current non-custodial stake user may want to change the output address some time after they stake nodes.

### Technical details:

Below are the key functions in the transaction processing flow.  This patch makes a new `MsgStake` transaction to change the output address pass through all Validate* functions and lets the final function `EditStakeValidator` accept the change in the world state if the new feature key “OEDIT” is enabled.

1. BaseApp.runTx (baseapp.go)
    1. auth.ValidateTransaction (ante.go)
    2. BaseApp.runMsg (baseapp.go)
        1. nodes.handleStake (handler.go)
            1. keeper.ValidateValidatorStaking (valStateChanges.go)
                1. keeper.ValidateValidatorMsgSigner (for the new validator)
                2. keeper.ValidateValidatorMsgSigner (for the current validator)
                3. keeper.ValidateEditStake
            2. keeper.StakeValidator
                1. keeper.EditStakeValidator

The challenge was in `ValidateTransaction`, which expects a transaction message to be signed by a key which appears in the transaction itself.  For example, `MsgStake` must be signed by `MsgStake.PublicKey` or `MsgStake.Output`.  Since we intend to allow only the current output address owner to change the output address, however, a new `MsgStake` transaction must be signed by the current output address, and it doesn’t appear in the transaction because a new output address is there.  This patch needs to change that signer restriction rule.

Each message type implements `GetSigners` that returns a slice of `Address`.  When a transaction arrives at `ValidateTransaction`, it calls `GetSigners` and checks if a transaction is signed by any of the addresses returned from `GetSigners`.  If not, the transaction is rejected with `CodeUnauthorized`.  The challenge here is that `ValidateTransaction` is in the auth package that doesn’t know about `MsgStake`.  One possible solution is to make `MsgStake.GetSigners` return three addresses, `MsgStake.PublicKey`, `MsgStake.Output`, and the current output address.  However, implementing it is not straightforward because `MsgStake` is in the types package which doesn’t have direct access to the keeper package that is required to get the current output address.

The proposed solution in this patch is to introduce a new method `GetMsgStakeOutputSigner` in the keeper package and `ValidateTransaction` calls it to append the current output address to the slice returned from `GetSigners`.

The changes in valStateChanges.go are straightforward.  `ValidateValidatorMsgSigner` simply checks if a signer is either the node’s operator or output address.  We need to skip the call to `ValidateValidatorStaking` with the new validator because the signer does not appear in the new validator.  The final validate function `ValidateEditStake` ensures that `MsgStake` to change the output address must be signed by the current output address.  Otherwise it returns a new error `ErrDisallowedOutputAddressChange`.

After all validate functions, `EditStakeValidator` accepts a new output address.

### Testing:

1. Prepare a local network with NCUST enabled.
2. Set the following environment variables accordingly.
    ```
    export POKT=<path to pocket executable>
    export CHAINS=<chain IDs to host>
    export URL=<service url>
    export STAKE=<upokt to stake>
    export NETWORK=<network name>
    export NODE=<node address>
    export PUBKEY=<node public key>
    export PRIVKEY=<node private key>
    export OUTPUT1=<output address>
    export OUTPUT1_PRIVKEY=<OUTPUT1’s private key>
    export OUTPUT2=<another output address>
    export OUTPUT2_PRIVKEY=<OUTPUT2’s private key>
    export DAO=<dao address>
    ```
3. Make sure all addresses are stored in the keybase and have enough tokens.
4. Submit a non-custodial stake tx to stake a new node.
    ```
    $POKT accounts delete $OUTPUT1
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT1 $STAKE $CHAINS $URL $NETWORK 10000 false
    ```
5. Submit three stake transactions to change the output address signed by the new output address, the operator address, and the current output address respectively.  Verify all transactions are rejected with CodeUnauthorizedSigner (pos:125), CodeUnequalOutputAddr (pos:124), and CodeUnauthorized (sdk:4).
    ```
    $POKT accounts import-raw $OUTPUT1_PRIVKEY
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    $POKT accounts delete $OUTPUT2
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    $POKT accounts delete $NODE
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    ```
6. Enable OEDIT.  Choose the upgrade height accordingly.
    ```
    $POKT gov enable $DAO <upgrade height> OEDIT $NETWORK 10000
    ```
7. Submit three stake transactions to change the output address signed by the new output address, the operator address, and the current output address respectively.  Verify the first two transactions are rejected with CodeUnauthorizedSigner (pos:125) and CodeDisallowedOutputAddressEdit (pos:127) respectively, and the last transaction is accepted.
    ```
    $POKT accounts import-raw $PRIVKEY
    $POKT accounts import-raw $OUTPUT2_PRIVKEY
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    $POKT accounts delete $OUTPUT2
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    $POKT accounts delete $NODE
    $POKT nodes stake non-custodial $PUBKEY $OUTPUT2 $STAKE $CHAINS $URL $NETWORK 10000 false
    ```
8. Verify the node’s output address is changed to $OUTPUT2.


codec/codec.go
```diff
@@ -266,6 +267,11 @@ func (cdc *Codec) IsAfterNonCustodialUpgrade(height int64) bool {
 	return (UpgradeFeatureMap[NonCustodialUpdateKey] != 0 && height >= UpgradeFeatureMap[NonCustodialUpdateKey]) || TestMode <= -3
 }
 
+func (cdc *Codec) IsAfterOutputAddressChangeUpgrade(height int64) bool {
+	return UpgradeFeatureMap[OutputAddressEditKey] != 0 &&
+		height >= UpgradeFeatureMap[OutputAddressEditKey]
```
Olshansk: Add a `|| TestMode <= -3` similar to the other recent changes

codec/codec.go
```diff
@@ -266,6 +267,11 @@ func (cdc *Codec) IsAfterNonCustodialUpgrade(height int64) bool {
 	return (UpgradeFeatureMap[NonCustodialUpdateKey] != 0 && height >= UpgradeFeatureMap[NonCustodialUpdateKey]) || TestMode <= -3
 }
 
+func (cdc *Codec) IsAfterOutputAddressChangeUpgrade(height int64) bool {
```
Olshansk: Optional: s/IsAfterOutputAddressChangeUpgrade/IsAfterOutputAddressEditorUpgrade

wdyt?

app/app.go
```diff
@@ -87,6 +87,7 @@ func NewPocketCoreApp(genState GenesisState, keybase keys.Keybase, tmClient clie
 	// give pocket keeper to nodes module for easy cache clearing
 	app.nodesKeeper.PocketKeeper = app.pocketKeeper
 	app.appsKeeper.PocketKeeper = app.pocketKeeper
+	app.accountKeeper.POSKeeper = app.nodesKeeper
```
Olshansk, line 90: For my personal understanding, can you explain to me why this was necessary?

x/auth/ante.go
```diff
@@ -61,8 +62,16 @@ func ValidateTransaction(ctx sdk.Ctx, k Keeper, stdTx types.StdTx, params Params
 	if res != nil {
 		return nil, types.ErrDuplicateTx(ModuleName, hex.EncodeToString(txHash))
 	}
+
+	signersAllowed := stdTx.GetSigners()
```
Olshansk: s/signersAllowed/validSigners

x/auth/ante.go
```diff
@@ -61,8 +62,16 @@ func ValidateTransaction(ctx sdk.Ctx, k Keeper, stdTx types.StdTx, params Params
 	if res != nil {
```
Olshansk, line 62: The intetion of this PR is to allow changes in the `MsgStake` command, but this validates ANY transaction and we are hacking the `MsgStake` piece via `GetExternalSigners`.

I personally would've looked at `x/nodes.handler.go` and made the change in:

```
func handleStake(ctx sdk.Ctx, msg types.MsgStake, k kee
```

This would allow removing the new interface altogether. 

Let me know if I'm missing something or what the thought process was.

x/nodes/types/errors.go
```diff
@@ -11,32 +10,33 @@ import (
...
+	CodeInvalidServiceURL             CodeType          = 118
+	CodeInvalidNetworkIdentifier      CodeType          = 119
+	CodeTooManyChains                 CodeType          = 120
+	CodeStateConvertError             CodeType          = 121
+	CodeMinimumEditStake              CodeType          = 122
+	CodeNilOutputAddr                 CodeType          = 123
+	CodeUnequalOutputAddr             CodeType          = 124
+	CodeUnauthorizedSigner            CodeType          = 125
+	CodeNilSigner                     CodeType          = 126
+	CodeDisallowedOutputAddressChange CodeType          = 127
```
Olshansk: I think this is more descriptive: `s/CodeDisallowedOutputAddressChange/CodeDisallowedOutputAddressEdit`

x/nodes/keeper/valStateChanges.go
```diff
@@ -109,48 +109,58 @@ func (k Keeper) UpdateTendermintValidators(ctx sdk.Ctx) (updates []abci.Validato
...
-func (k Keeper) ValidateValidatorStaking(ctx sdk.Ctx, validator types.Validator, amount sdk.BigInt, signerAddress sdk.Address) sdk.Error {
-
-	//check the "new" validator's signature validity
-	//will recheck if validator exists
-	err, valid := ValidateValidatorMsgSigner(validator, signerAddress, k)
-	if !valid {
-		return err
+func (k Keeper) ValidateValidatorStaking(ctx sdk.Ctx,
+	validatorNew types.Validator, amount sdk.BigInt,
+	signerAddress sdk.Address) sdk.Error {
```
Olshansk:
```suggestion
func (k Keeper) ValidateValidatorStaking(
	ctx sdk.Ctx,
	validatorNew types.Validator,
	amount sdk.BigInt,
	signerAddress sdk.Address
) sdk.Error {
```

x/nodes/keeper/common_test.go
```diff
@@ -135,6 +133,17 @@ func sendFromModuleToAccount(t *testing.T, ctx sdk.Ctx, k *Keeper, module string
 	}
 }
 
+func fundAccount(ctx sdk.Context, k Keeper,
+	address sdk.Address, coin sdk.Coin) sdk.Error {
```
Olshansk: NIT: Either keep everything on one line or multiline the whole thing
```suggestion
func fundAccount(
	ctx sdk.Context,
	k Keeper,
	address sdk.Address,
	coin sdk.Coin
) sdk.Error {
```

x/nodes/keeper/common_test.go
```diff
@@ -135,6 +133,17 @@ func sendFromModuleToAccount(t *testing.T, ctx sdk.Ctx, k *Keeper, module string
...
+	coins := sdk.NewCoins(coin)
+	err := k.AccountKeeper.MintCoins(ctx, types.StakedPoolName, coins)
+	if err != nil {
+		return err
+	}
```
Olshansk:
```suggestion
	if err := k.AccountKeeper.MintCoins(ctx, types.StakedPoolName, coins); err != nil
		return err
	}
```

x/nodes/keeper/common_test.go
```diff
@@ -135,6 +133,17 @@ func sendFromModuleToAccount(t *testing.T, ctx sdk.Ctx, k *Keeper, module string
 	}
 }
 
+func fundAccount(ctx sdk.Context, k Keeper,
```
Olshansk:
```suggestion
// fundAccount mints new tokens in the stake pool and sends them
// to the address provided, for testing purposes.
func fundAccount(ctx sdk.Context, k Keeper,
```

x/auth/types/expectedKeepers.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+import (
+	sdk "github.com/pokt-network/pocket-core/types"
+)
+
+type PosKeeper interface {
+	GetExternalSigners(sdk.Ctx, sdk.Msg) []sdk.Address
```
Olshansk: I left a comment above that might let us eliminate the need for this. 

But, if we do keep it: `GetMsgStakeOutputSigner(sdk.Ctx, sdk.Msg) *sdk.Address`

1. Make it clear and explicit
2. There will only ever be one in V0

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
 	}
 }
 
+func runStake(ctx sdk.Context,
```
Olshansk: Ditto msg header

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	validator := types.NewValidator(
+		sdk.Address(msg.PublicKey.Address()),
+		msg.PublicKey,
+		msg.Chains,
+		msg.ServiceUrl,
+		sdk.ZeroInt(),
+		msg.Output)
+	err := k.ValidateValidatorStaking(
+		ctx, validator, msg.Value, sdk.Address(signer.Address()))
+	if err != nil {
```
Olshansk:
```suggestion
	if err := k.ValidateValidatorStaking(ctx, validator, msg.Value, sdk.Address(signer.Address())); err != nil {
```

x/nodes/keeper/valStateChanges.go
```diff
@@ -109,48 +109,58 @@ func (k Keeper) UpdateTendermintValidators(ctx sdk.Ctx) (updates []abci.Validato
...
+	// check to see if the public key has already been register for that validator
+	validatorCur, found := k.GetValidator(ctx, validatorNew.Address)
+	if found &&
+		validatorNew.Address.Equals(validatorCur.Address) &&
```
Olshansk, line 129: Can you move the formulation of this boolean into a local var.

Then the if will be simple to read and you can add `NOOP` comment inside.

x/nodes/keeper/valStateChanges.go
```diff
@@ -109,48 +109,58 @@ func (k Keeper) UpdateTendermintValidators(ctx sdk.Ctx) (updates []abci.Validato
...
+		validatorNew.OutputAddress != nil &&
+		k.Cdc.IsAfterNonCustodialUpgrade(ctx.BlockHeight()) &&
+		k.Cdc.IsAfterOutputAddressChangeUpgrade(ctx.BlockHeight()) {
+		// In this case, for edit stake, we skip verifying an operation
+		// against the new validator to accept a signer that appears
+		// only in the current validator.
```
Olshansk: Doesn't the `outputAddress` need to be the `signerAddress`?

x/nodes/keeper/valStateChanges.go
```diff
@@ -228,10 +238,18 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentValidator, newValidtor typ
...
 	if k.Cdc.IsAfterNonCustodialUpgrade(ctx.BlockHeight()) {
-		// ensure output address doesn't change
-		if currentValidator.OutputAddress != nil {
-			if !newValidtor.OutputAddress.Equals(currentValidator.OutputAddress) {
-				fmt.Println(currentValidator.String())
+		if k.Cdc.IsAfterOutputAddressChangeUpgrade(ctx.BlockHeight()) {
+			// ensure only the current output address owner can change the output address
+			if currentValidator.OutputAddress != nil &&
+				!signer.Equals(currentValidator.OutputAddress) &&
+				!newValidtor.OutputAddress.Equals(currentValidator.OutputAddress) {
```
Olshansk, line 254: Wouldn't `!newValidtor.OutputAddress.Equals(currentValidator.OutputAddress)` prevent the signer from changing the output address altogether?

x/nodes/keeper/valStateChanges.go
```diff
@@ -228,10 +238,18 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentValidator, newValidtor typ
...
+				return types.ErrDisallowedOutputAddressChange(k.Codespace())
+			}
+		} else {
+			//check the "new" validator's signature validity
+			//will recheck if validator exists
```
Olshansk: Why not keep this as `// ensure output address doesn't change` like before? It's pretty clear.
```suggestion
			// check the "new" validator's signature validity
			// will re-check if validator exists
```

x/nodes/keeper/valStateChanges.go
```diff
@@ -109,48 +109,58 @@ func (k Keeper) UpdateTendermintValidators(ctx sdk.Ctx) (updates []abci.Validato
...
+		//check again based on the "current" validator's state
+		err, valid := ValidateValidatorMsgSigner(validatorCur, signerAddress, k)
 		if !valid {
 			return err
 		}
 		// edit stake in 6.X upgrade
-		if ctx.IsAfterUpgradeHeight() && val.IsStaked() {
-			return k.ValidateEditStake(ctx, val, validator, amount, signerAddress)
+		if ctx.IsAfterUpgradeHeight() && validatorCur.IsStaked() {
+			return k.ValidateEditStake(ctx, validatorCur, validatorNew, amount, signerAddress)
```
Olshansk, line 164: This feature is all about `editing` stake, so why not move the core business logic (with the feature checks) into `ValidateEditStake`? 

I think it'll be easier to read that way.

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+		return err
+	}
+	return k.StakeValidator(ctx, validator, msg.Value, signer)
+}
+
+func Test_OutputAddressChange(t *testing.T) {
```
Olshansk: TestValidatorStateChange_OutputAddressEdit

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	ctx, _, k := createTestInput(t, true)
+
+	// Enable EditStake
+	codec.UpgradeHeight = -1
+
+	stakeAmount := sdk.NewCoin(k.StakeDenom(ctx), sdk.NewInt(k.MinimumStake(ctx)))
+	outputAccount := getRandomPubKey()
```
Olshansk: s/outputAccount/outputPubKey

Olshansk: Ditto below

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	outputAddress := sdk.Address(outputAccount.Address())
+	msgTemplate := types.MsgStake{
```
Olshansk: s/msgTemplate/msgStakeTemplate

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	operatorAcc3 := getRandomPubKey()
+	operatorAddr1 := sdk.Address(operatorAcc1.Address())
+	operatorAddr2 := sdk.Address(operatorAcc2.Address())
+	operatorAddr3 := sdk.Address(operatorAcc3.Address())
+	assert.Nil(t, fundAccount(ctx, k, operatorAddr1, stakeAmount))
+	assert.Nil(t, fundAccount(ctx, k, operatorAddr2, stakeAmount))
+	assert.Nil(t, fundAccount(ctx, k, outputAddress, stakeAmount))
+
+	// Stake two nodes before NCUST
+	msgTemplate.PublicKey = operatorAcc1
```
Olshansk: I think you should copy `msgStake = proto.Clone(msgStakeTemplate)` and then update it.

It doesn't matter from a test passing perspective, but just makes more sense and less prone to error.

Olshansk: Ditto below

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	// Enable NCUST
+	codec.UpgradeFeatureMap[codec.NonCustodialUpdateKey] = -1
+
+	// Set an output address on Addr1 after NCUST
+	msgTemplate.PublicKey = operatorAcc1
+	msgTemplate.Output = outputAddress
+	assert.Nil(t, runStake(ctx, k, msgTemplate, operatorAcc1))
+	validatorCur, found = k.GetValidator(ctx, operatorAddr1)
+	assert.True(t, found)
+	assert.True(t, validatorCur.OutputAddress.Equals(outputAddress))
```
Olshansk: Use `assert.Equal` directly

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+
+	// Attempt to change the output address with operator's signature --> Fail
+	msgTemplate.PublicKey = operatorAcc1
+	msgTemplate.Output = operatorAddr1
+	err := runStake(ctx, k, msgTemplate, operatorAcc1)
```
Olshansk: assert that it's NotNil first. Ditto below a few times

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+	// Attempt to change the output address with output's signature --> Success
+	msgTemplate.PublicKey = operatorAcc1
+	msgTemplate.Output = operatorAddr1
+	err = runStake(ctx, k, msgTemplate, outputAccount)
+	assert.Nil(t, err)
+	validatorCur, found = k.GetValidator(ctx, operatorAddr1)
+	assert.True(t, found)
+	assert.True(t, validatorCur.OutputAddress.Equals(operatorAddr1))
+
+	// Attempt to change the output address from nil
```
Olshansk, line 465: This change is reasonable but wasn't clear from the github description. Please update it appropriately.

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -350,6 +352,129 @@ func TestValidatorStateChange_EditAndValidateStakeValidatorAfterNonCustodialUpgr
...
+
+	// New stake acc3 with outputAccount with output's signature --> Success
+	msgTemplate.PublicKey = operatorAcc3
+	msgTemplate.Output = outputAddress
+	err = runStake(ctx, k, msgTemplate, outputAccount)
+	assert.Nil(t, err)
+	validatorCur, found = k.GetValidator(ctx, operatorAddr3)
+	assert.True(t, found)
+	assert.True(t, validatorCur.OutputAddress.Equals(outputAddress))
+}
```
Olshansk, line 479: Can you please update the github description with:
1. A list of all the changes
2. Example `pocket` commands you ran to verify this on a locally running node
3. Just two sentences on a reasoning / rationale that we'll include in the release notes

Olshansk (changes requested):

x/nodes/keeper/valStateChanges_test.go
msmania: Yeah, if it's named `msgTemplate`, it should behave as a template.  I found a better way to call `runStake`.

x/nodes/keeper/valStateChanges_test.go
msmania: Oh I didn't know `assert.Equal` is that smart.  Good to know.

x/nodes/keeper/valStateChanges.go
msmania: Before `ValidateEditStake`, `ValidateValidatorStaking` calls `ValidateValidatorMsgSigner` twice, one with the new validator and one with the current validator.  We need to accept a stake transaction signed by the current output address, but it doesn't pass the first `ValidateValidatorMsgSigner` against the new validator because the new validator doesn't have the signer, the current output address.  That's why we need a logic to skip the first `ValidateValidatorMsgSigner`.  We can't move that part to `ValidateEditStake`.

Well, I'd like to remove `ValidateValidatorMsgSigner` and do all checks inside `ValidateEditStake`, but that kind of refactoring change shouldn't be included in this patch.

x/nodes/keeper/valStateChanges.go
msmania: Yes, I delegated that check to `ValidateEditStake`, but having it here too does no harm.  Will add it.

x/auth/ante.go
msmania: Before reaching `handleStake`, all transactions must pass the checks in `auth.ValidateTransaction`.  So we need to change that function.  I'll update the description with more details.

app/app.go
msmania: This is for cross-package keeper access.  What we need to achieve is to retrieve the current output address in `ValidateTransaction`, which needs `nodes/Keeper`.  `ValidateTransaction` is in the `auth` package, however, the keeper `ValidateTransaction` takes as a parameter is `auth/Keeper`.  This property `POSKeeper` gives us access to `nodes/Keeper`.

x/nodes/keeper/valStateChanges.go
msmania: If a signer is not the current output address, yes, we prevent the output address change.

Olshansk:
@msmania 

Here's a mermaid diagram ChatGPT gave me using your github description.

Screenshot 2023-04-04 at 7 44 37 PM

Screenshot 2023-04-04 at 7 44 02 PM

```mermaid
flowchart TD
    A(BaseApp.runTx) --> B(auth.ValidateTransaction)
    B --> C(BaseApp.runMsg)
    C --> D(nodes.handleStake)
    D --> E(keeper.ValidateValidatorStaking)
    E --> F(keeper.ValidateValidatorMsgSigner)
    F --> G(keeper.ValidateValidatorMsgSigner)
    G --> H(keeper.ValidateEditStake)
    H --> I(keeper.StakeValidator)
    I --> J(keeper.EditStakeValidator)

    subgraph Before
        B -->|Validate using signed keys|L(MsgStake.GetSigners)
        H -->|Ensure signed by current output address| M(Error)
    end

    subgraph After
        B --> K(GetMsgStakeOutputSigner)
        K -->|Append current output address|L(MsgStake.GetSigners)
        H -->|Ensure signed by current output address| M(ValidateEditStake)
    end
```

app/cmd/cli/txUtil.go
```diff
@@ -123,11 +122,41 @@ func LegacyStakeNode(chains []string, serviceURL, fromAddr, passphrase, chainID
...
 
+func getCurrentOutputAddress(operatorAddr sdk.Address) (sdk.Address, error) {
+	j, err := json.Marshal(rpc.HeightAndAddrParams{
+		Height:  0,
+		Address: operatorAddr.String(),
+	})
+	if err != nil {
+		return nil, err
+	}
+	var node nodeTypes.Validator
```
Olshansk: NIT: This should go above the unmarshal where you use it

app/cmd/cli/txUtil.go
```diff
@@ -123,11 +122,41 @@ func LegacyStakeNode(chains []string, serviceURL, fromAddr, passphrase, chainID
 	}, nil
 }
 
+func getCurrentOutputAddress(operatorAddr sdk.Address) (sdk.Address, error) {
```
Olshansk, line 127:
```suggestion
// getCurrentOutputAddress returns the current output address for the operator provided.
// nil is returned if there is no output address.
func getCurrentOutputAddress(operatorAddr sdk.Address) (sdk.Address, error) {
```

app/cmd/cli/txUtil.go
```diff
@@ -123,11 +122,41 @@ func LegacyStakeNode(chains []string, serviceURL, fromAddr, passphrase, chainID
...
+	res, err := QueryRPC(GetNodePath, j)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := json.Unmarshal([]byte(res), &node); err != nil {
+		return nil, err
+	}
+
+	return node.OutputAddress, err
```
Olshansk:
```suggestion
	return node.OutputAddress, nil
```

app/cmd/cli/txUtil.go
```diff
@@ -123,11 +122,41 @@ func LegacyStakeNode(chains []string, serviceURL, fromAddr, passphrase, chainID
...
+}
+
+func getFirstAddressAvailableInKeybase(
+	kb keys.Keybase,
+	addresses []sdk.Address,
+) (sdk.Address, bool) {
```
Olshansk: Why not return an `*sdk.Address` so you can check for a nil or non-nil?

app/cmd/cli/txUtil.go
```diff
@@ -123,11 +122,41 @@ func LegacyStakeNode(chains []string, serviceURL, fromAddr, passphrase, chainID
...
+	}
+
+	if err := json.Unmarshal([]byte(res), &node); err != nil {
+		return nil, err
+	}
+
+	return node.OutputAddress, err
+}
+
+func getFirstAddressAvailableInKeybase(
```
Olshansk, line 152: NIT: Can you add a comment on the use case for this? I kind of feel like you're use-case is more like `anyExistsInKeybase`

app/cmd/cli/stake.go
```diff
@@ -93,11 +94,19 @@ var nonCustodialstakeCmd = &cobra.Command{
...
+Will prompt the user for the signer account passphrase, fund and fees are
+collected from signer account.  If the node is already staked, this transaction
+acts as an *update* transaction.  A node can update relayChainIDs, serviceURI,
+and raise the stake amount with this transaction.  The output address can be
+changed when the output address is empty or a transaction is signed by the
+current output address.
+If the node is currently staked at X and you submit an update with new stake Y,
+only Y-X will be subtracted from an account.
+If no changes are desired for the parameter, just enter the current param value
+just as before.`,
```
Olshansk:
```suggestion

If the node is currently staked at X and you submit an update with new stake Y,
only Y-X will be subtracted from an account.

If no changes are desired for the parameter, just enter the current param values as before.`,
```

x/nodes/keeper/valStateChanges.go
```diff
@@ -109,48 +108,68 @@ func (k Keeper) UpdateTendermintValidators(ctx sdk.Ctx) (updates []abci.Validato
...
-	err, valid := ValidateValidatorMsgSigner(validator, signerAddress, k)
-	if !valid {
-		return err
+	// Before the upgrade, we validate a signer against the new validator's state
+	// via ValidateValidatorMsgSigner.  After the upgrade, however, we accept
+	// MsgStake signed by the current output address that doesn't appear in the
+	// new validator's state.  So we skip ValidateValidatorMsgSigner if 1) the
+	// current height is after the upgrade, 2) this is an edit-stake changing the
+	// output address, and 3) the signer is the current output address.
+	skipMsgSignerValidation := found &&
```
Olshansk, line 126: This is ultra clear and safe. Ty!!

Olshansk (changes requested):
A couple of minor NITS/comments, but otherwise this PR LGTM from a code perspective. Amazing test and amazingly clear code!

We need to submit a proposal to the DAO to approve this (cc @jessicadaugherty), but I don't expect any opposition from the community.

app/cmd/cli/txUtil.go
msmania: oops, copy-n-paste mistake

app/cmd/cli/txUtil.go
msmania: Ah right, that's why you suggested a pointer.  I got it now.

app/cmd/cli/txUtil.go
msmania: I'd prefer to keep "First" in the name because the searching order is important.  Will add a leading comment.

Olshansk (dismissed):
LGTM!! :shipit:

msmania:
> @msmania
> 
> Here's a mermaid diagram ChatGPT gave me using your github description.

Interesting.  Fortunately or unfortunately, AI made a mistake.  That flow is not correct.  I think it's hard to overlap before and after in the same chart.  Here are two charts before and after.  This matches the test scenario in the first description.

### Sending MsgStake without this patch

```mermaid
flowchart TD
  A(runTx) --> B(ValidateTransaction)
  B -- If signed by the current output address --> K([sdk:4])
  A --> C(runMsg)
  C --> D(handleStake)
  D --> E(ValidateValidatorStaking)
  D -.-> F(StakeValidator)
  E --> G(ValidateValidatorMsgSigner\nwith the current state)
  G -- If signed by the new output address --> L([pos:125])
  E --> H(ValidateValidatorMsgSigner\nwith the new state)
  E --> I(ValidateEditStake)
  I -- If signed by the operator --> M([pos:124])
  F -.-> J(EditStakeValidator)
```

### Sending MsgStake with this patch

```mermaid
flowchart TD
  A(runTx) --> B(ValidateTransaction)
  B --> N(GetMsgStakeOutputSigner)
  N -- If signed by the current output address --> K([ok])
  A --> C(runMsg)
  C --> D(handleStake)
  D --> E(ValidateValidatorStaking)
  D --> F(StakeValidator)
  E --> O{Valid output address edit?}
  O -. No .-> G(ValidateValidatorMsgSigner\nwith the current state)
  G -.-> L([pos:125])
  E --> H(ValidateValidatorMsgSigner\nwith the new state)
  E --> I(ValidateEditStake)
  I -- If signed by the operator --> M([pos:127])
  F --> J(EditStakeValidator)
  J --> P([Happy ending!])
```

Olshansk (approved):

Olshansk:
@msmania Wanted to encourage you to submit another PR updating the documentation so these awesome diagrams don't go to waste!

msmania:
> @msmania Wanted to encourage you to submit another PR updating the documentation so these awesome diagrams don't go to waste!

Created #1544

Issue #1464, opened 2022-07-31
[RFP/Improvement] Allowing for configurable session block height tolerance


nodiesBlade:
**Please explain the goal of the RFP.**
Pocket Core should allow for an allowance/tolerance of `n` (configurable) session block difference when validating relays.

**Please provide a justification for your RFP.**

Currently, the client requires that the dispatch nodes (Portal) and servicer nodes be on the same "latest" block height (defintion of latest [here](https://github.com/pokt-network/pocket-core/blob/d2ae3f8e28acf7f829fd0bf38b8a19e3bae9e3a4/x/pocketcore/keeper/session.go#L59)) before servicing a relay. If both nodes are not synced on same session block, the servicer will [return invalid block height errors](https://github.com/pokt-network/pocket-core/blob/4e55cecdae4438039d67ae895fd82ba7b7f7e6e2/x/pocketcore/types/service.go#L46). While this can help with synchronization attacks, it has caused numerous relays to be returned as invalid and affects our app's perceived QoS as seen as part of these poktscan errors:
<img width="1194" alt="Screen Shot 2022-07-31 at 1 14 12 PM" src="https://user-images.githubusercontent.com/99137075/182039795-169a738f-3f12-4bad-9be5-7280daaea908.png">

## Scenario
Dispatch node A is lagging behind (SessionBlockHeight: 1)
Servicer node B is able to stay synced (SessionBlockHeight: 5).

Since Pocket node now has a different "latest" session block height, it will no longer service incoming relays from apps that use dispatch node A.

**Please provide a success criteria for proposals responding to this RFP.**
- Allows for wiggle room on session block height when validating a relay
- Configurable via client setting, default value should be at `1` session block tolerance

**Add aditional context regarding this RFP.**
I would add onto this, session block height as a form of time synchronization is as not as effective, so the ask to change this behavior and the implied risks seems low. If we allow for a tolerance of ~1 session block height difference, we're allowing allowing for the time window to be increased to 8 blocks instead of 4.

I think we can take it up a notch and minimize this window further by allowing for a flexible sessionBlockHeightTolerance, resulting in a time window increase of one block.

Psuedo code:
```
If latestBlockHeight % 4 == 0 then
           sessionBlockHeightTolerance = 1
else
          sessionBlockHeightTolerance = 0
```
      
By doing this, we allow for dispatch nodes or pocket nodes to catch up within a single block before making the tolerance more stricter.

POKT-Discourse:
[...]

PR #1535, opened 2023-03-15
LocalNet documentation using playground

Olshansk:
# Description

Add documentation on how to use [e2e-stack](https://github.com/pokt-foundation/pocket-e2e-stack), [tx-bot](https://github.com/pokt-network/tx-bot) and [local_playground](https://github.com/pokt-network/local_playground) to run a 4 node LocalNet that enables hot reloading of our [Tendermint fork](https://github.com/pokt-network/tendermint).

## Import Note

Note that this is a hacky & ugly reuse of existing infrastructure and is not meant to be an elegant long-term solution. It is just enough to test & ship new features while v1 is in development.

## Dependencies

- Changes in [tx-bot](https://github.com/pokt-network/tx-bot/pull/6)
- Changes in [pocket-e2e-stack](https://github.com/pokt-foundation/pocket-e2e-stack/pull/30)
- Introduce of [local_playground](https://github.com/pokt-network/local_playground); see [this comment](https://github.com/pokt-network/playground/pull/52#issuecomment-1470946737) for more details

## Demo

https://user-images.githubusercontent.com/1892194/225497350-3817b262-f5d5-483b-bb8f-47fb0b614321.mov

PR #1536, opened 2023-03-16
Session roll over fix

nodiesBlade:
Addresses #1464 by adding a configurable session block tolerance into pocket config `client_session_sync_allowance` with a default value of `1`

This means that once a new session appears, servicer nodes will still respect old sessions up until 1 session block (4 blocks at the time of this PR). 

DISCLAIMER: SHOULD DEFINITELY TEST.......................... USING THE REGULAR RELEASE CYCLE MECHANISMS

### DIFF
--- a/types/config.go
+++ b/types/config.go
@@ -18,36 +18,37 @@ type SDKConfig struct {
 }
 
 type PocketConfig struct {
-	DataDir                   string `json:"data_dir"`
-	GenesisName               string `json:"genesis_file"`
-	ChainsName                string `json:"chains_name"`
-	EvidenceDBName            string `json:"evidence_db_name"`
-	TendermintURI             string `json:"tendermint_uri"`
-	KeybaseName               string `json:"keybase_name"`
-	RPCPort                   string `json:"rpc_port"`
-	ClientBlockSyncAllowance  int    `json:"client_block_sync_allowance"`
-	MaxEvidenceCacheEntires   int    `json:"max_evidence_cache_entries"`
-	MaxSessionCacheEntries    int    `json:"max_session_cache_entries"`
-	JSONSortRelayResponses    bool   `json:"json_sort_relay_responses"`
-	RemoteCLIURL              string `json:"remote_cli_url"`
-	UserAgent                 string `json:"user_agent"`
-	ValidatorCacheSize        int64  `json:"validator_cache_size"`
-	ApplicationCacheSize      int64  `json:"application_cache_size"`
-	RPCTimeout                int64  `json:"rpc_timeout"`
-	PrometheusAddr            string `json:"pocket_prometheus_port"`
-	PrometheusMaxOpenfiles    int    `json:"prometheus_max_open_files"`
-	MaxClaimAgeForProofRetry  int    `json:"max_claim_age_for_proof_retry"`
-	ProofPrevalidation        bool   `json:"proof_prevalidation"`
-	CtxCacheSize              int    `json:"ctx_cache_size"`
-	ABCILogging               bool   `json:"abci_logging"`
-	RelayErrors               bool   `json:"show_relay_errors"`
-	DisableTxEvents           bool   `json:"disable_tx_events"`
-	Cache                     bool   `json:"-"`
-	IavlCacheSize             int64  `json:"iavl_cache_size"`
-	ChainsHotReload           bool   `json:"chains_hot_reload"`
-	GenerateTokenOnStart      bool   `json:"generate_token_on_start"`
-	LeanPocket                bool   `json:"lean_pocket"`
-	LeanPocketUserKeyFileName string `json:"lean_pocket_user_key_file"`
+	DataDir                         string `json:"data_dir"`
+	GenesisName                     string `json:"genesis_file"`
+	ChainsName                      string `json:"chains_name"`
+	EvidenceDBName                  string `json:"evidence_db_name"`
+	TendermintURI                   string `json:"tendermint_uri"`
+	KeybaseName                     string `json:"keybase_name"`
+	RPCPort                         string `json:"rpc_port"`
+	ClientBlockSyncAllowance        int    `json:"client_block_sync_allowance"`
+	ClientSessionBlockSyncAllowance int64  `json:"client_session_block_sync_allowance"`
+	MaxEvidenceCacheEntires         int    `json:"max_evidence_cache_entries"`
+	MaxSessionCacheEntries          int    `json:"max_session_cache_entries"`
+	JSONSortRelayResponses          bool   `json:"json_sort_relay_responses"`
+	RemoteCLIURL                    string `json:"remote_cli_url"`
+	UserAgent                       string `json:"user_agent"`
+	ValidatorCacheSize              int64  `json:"validator_cache_size"`
+	ApplicationCacheSize            int64  `json:"application_cache_size"`
+	RPCTimeout                      int64  `json:"rpc_timeout"`
+	PrometheusAddr                  string `json:"pocket_prometheus_port"`
+	PrometheusMaxOpenfiles          int    `json:"prometheus_max_open_files"`
+	MaxClaimAgeForProofRetry        int    `json:"max_claim_age_for_proof_retry"`
+	ProofPrevalidation              bool   `json:"proof_prevalidation"`
+	CtxCacheSize                    int    `json:"ctx_cache_size"`
+	ABCILogging                     bool   `json:"abci_logging"`
+	RelayErrors                     bool   `json:"show_relay_errors"`
+	DisableTxEvents                 bool   `json:"disable_tx_events"`
+	Cache                           bool   `json:"-"`
+	IavlCacheSize                   int64  `json:"iavl_cache_size"`
+	ChainsHotReload                 bool   `json:"chains_hot_reload"`
+	GenerateTokenOnStart            bool   `json:"generate_token_on_start"`
+	LeanPocket                      bool   `json:"lean_pocket"`
+	LeanPocketUserKeyFileName       string `json:"lean_pocket_user_key_file"`
 }
 
 func (c PocketConfig) GetLeanPocketUserKeyFilePath() string {
@@ -82,6 +83,7 @@ const (
 	DefaultMaxEvidenceCacheEntries     = 500
 	DefaultListenAddr                  = "tcp://0.0.0.0:"
 	DefaultClientBlockSyncAllowance    = 10
+	DefaultSessionBlockSyncAllowance   = 1 // 1 session (irrespective of num blocks per session)
 	DefaultJSONSortRelayResponses      = true
 	DefaultTxIndexer                   = "kv"
 	DefaultRPCDisableTransactionEvents = true
@@ -117,35 +119,36 @@ func DefaultConfig(dataDir string) Config {
 	c := Config{
 		TendermintConfig: *config.DefaultConfig(),
 		PocketConfig: PocketConfig{
-			DataDir:                   dataDir,
-			GenesisName:               DefaultGenesisName,
-			ChainsName:                DefaultChainsName,
-			EvidenceDBName:            DefaultEvidenceDBName,
-			TendermintURI:             DefaultTMURI,
-			KeybaseName:               DefaultKeybaseName,
-			RPCPort:                   DefaultRPCPort,
-			ClientBlockSyncAllowance:  DefaultClientBlockSyncAllowance,
-			MaxEvidenceCacheEntires:   DefaultMaxEvidenceCacheEntries,
-			MaxSessionCacheEntries:    DefaultMaxSessionCacheEntries,
-			JSONSortRelayResponses:    DefaultJSONSortRelayResponses,
-			RemoteCLIURL:              DefaultRemoteCLIURL,
-			UserAgent:                 DefaultUserAgent,
-			ValidatorCacheSize:        DefaultValidatorCacheSize,
-			ApplicationCacheSize:      DefaultApplicationCacheSize,
-			RPCTimeout:                DefaultRPCTimeout,
-			PrometheusAddr:            DefaultPocketPrometheusListenAddr,
-			PrometheusMaxOpenfiles:    DefaultPrometheusMaxOpenFile,
-			MaxClaimAgeForProofRetry:  DefaultMaxClaimProofRetryAge,
-			ProofPrevalidation:        DefaultProofPrevalidation,
-			CtxCacheSize:              DefaultCtxCacheSize,
-			ABCILogging:               DefaultABCILogging,
-			RelayErrors:               DefaultRelayErrors,
-			DisableTxEvents:           DefaultRPCDisableTransactionEvents,
-			IavlCacheSize:             DefaultIavlCacheSize,
-			ChainsHotReload:           DefaultChainHotReload,
-			GenerateTokenOnStart:      DefaultGenerateTokenOnStart,
-			LeanPocket:                DefaultLeanPocket,
-			LeanPocketUserKeyFileName: DefaultLeanPocketUserKeyFileName,
+			DataDir:                         dataDir,
+			GenesisName:                     DefaultGenesisName,
+			ChainsName:                      DefaultChainsName,
+			EvidenceDBName:                  DefaultEvidenceDBName,
+			TendermintURI:                   DefaultTMURI,
+			KeybaseName:                     DefaultKeybaseName,
+			RPCPort:                         DefaultRPCPort,
+			ClientBlockSyncAllowance:        DefaultClientBlockSyncAllowance,
+			ClientSessionBlockSyncAllowance: DefaultSessionBlockSyncAllowance,
+			MaxEvidenceCacheEntires:         DefaultMaxEvidenceCacheEntries,
+			MaxSessionCacheEntries:          DefaultMaxSessionCacheEntries,
+			JSONSortRelayResponses:          DefaultJSONSortRelayResponses,
+			RemoteCLIURL:                    DefaultRemoteCLIURL,
+			UserAgent:                       DefaultUserAgent,
+			ValidatorCacheSize:              DefaultValidatorCacheSize,
+			ApplicationCacheSize:            DefaultApplicationCacheSize,
+			RPCTimeout:                      DefaultRPCTimeout,
+			PrometheusAddr:                  DefaultPocketPrometheusListenAddr,
+			PrometheusMaxOpenfiles:          DefaultPrometheusMaxOpenFile,
+			MaxClaimAgeForProofRetry:        DefaultMaxClaimProofRetryAge,
+			ProofPrevalidation:              DefaultProofPrevalidation,
+			CtxCacheSize:                    DefaultCtxCacheSize,
+			ABCILogging:                     DefaultABCILogging,
+			RelayErrors:                     DefaultRelayErrors,
+			DisableTxEvents:                 DefaultRPCDisableTransactionEvents,
+			IavlCacheSize:                   DefaultIavlCacheSize,
+			ChainsHotReload:                 DefaultChainHotReload,
+			GenerateTokenOnStart:            DefaultGenerateTokenOnStart,
+			LeanPocket:                      DefaultLeanPocket,
+			LeanPocketUserKeyFileName:       DefaultLeanPocketUserKeyFileName,
 		},
 	}
 	c.TendermintConfig.LevelDBOptions = config.DefaultLevelDBOpts()
--- a/types/config.go
+++ b/types/config.go
@@ -18,37 +18,37 @@ type SDKConfig struct {
 }
 
 type PocketConfig struct {
-	DataDir                         string `json:"data_dir"`
-	GenesisName                     string `json:"genesis_file"`
-	ChainsName                      string `json:"chains_name"`
-	EvidenceDBName                  string `json:"evidence_db_name"`
-	TendermintURI                   string `json:"tendermint_uri"`
-	KeybaseName                     string `json:"keybase_name"`
-	RPCPort                         string `json:"rpc_port"`
-	ClientBlockSyncAllowance        int    `json:"client_block_sync_allowance"`
-	ClientSessionBlockSyncAllowance int64  `json:"client_session_block_sync_allowance"`
-	MaxEvidenceCacheEntires         int    `json:"max_evidence_cache_entries"`
-	MaxSessionCacheEntries          int    `json:"max_session_cache_entries"`
-	JSONSortRelayResponses          bool   `json:"json_sort_relay_responses"`
-	RemoteCLIURL                    string `json:"remote_cli_url"`
-	UserAgent                       string `json:"user_agent"`
-	ValidatorCacheSize              int64  `json:"validator_cache_size"`
-	ApplicationCacheSize            int64  `json:"application_cache_size"`
-	RPCTimeout                      int64  `json:"rpc_timeout"`
-	PrometheusAddr                  string `json:"pocket_prometheus_port"`
-	PrometheusMaxOpenfiles          int    `json:"prometheus_max_open_files"`
-	MaxClaimAgeForProofRetry        int    `json:"max_claim_age_for_proof_retry"`
-	ProofPrevalidation              bool   `json:"proof_prevalidation"`
-	CtxCacheSize                    int    `json:"ctx_cache_size"`
-	ABCILogging                     bool   `json:"abci_logging"`
-	RelayErrors                     bool   `json:"show_relay_errors"`
-	DisableTxEvents                 bool   `json:"disable_tx_events"`
-	Cache                           bool   `json:"-"`
-	IavlCacheSize                   int64  `json:"iavl_cache_size"`
-	ChainsHotReload                 bool   `json:"chains_hot_reload"`
-	GenerateTokenOnStart            bool   `json:"generate_token_on_start"`
-	LeanPocket                      bool   `json:"lean_pocket"`
-	LeanPocketUserKeyFileName       string `json:"lean_pocket_user_key_file"`
+	DataDir                    string `json:"data_dir"`
+	GenesisName                string `json:"genesis_file"`
+	ChainsName                 string `json:"chains_name"`
+	EvidenceDBName             string `json:"evidence_db_name"`
+	TendermintURI              string `json:"tendermint_uri"`
+	KeybaseName                string `json:"keybase_name"`
+	RPCPort                    string `json:"rpc_port"`
+	ClientBlockSyncAllowance   int    `json:"client_block_sync_allowance"`
+	ClientSessionSyncAllowance int64  `json:"client_session_sync_allowance"`
+	MaxEvidenceCacheEntires    int    `json:"max_evidence_cache_entries"`
+	MaxSessionCacheEntries     int    `json:"max_session_cache_entries"`
+	JSONSortRelayResponses     bool   `json:"json_sort_relay_responses"`
+	RemoteCLIURL               string `json:"remote_cli_url"`
+	UserAgent                  string `json:"user_agent"`
+	ValidatorCacheSize         int64  `json:"validator_cache_size"`
+	ApplicationCacheSize       int64  `json:"application_cache_size"`
+	RPCTimeout                 int64  `json:"rpc_timeout"`
+	PrometheusAddr             string `json:"pocket_prometheus_port"`
+	PrometheusMaxOpenfiles     int    `json:"prometheus_max_open_files"`
+	MaxClaimAgeForProofRetry   int    `json:"max_claim_age_for_proof_retry"`
+	ProofPrevalidation         bool   `json:"proof_prevalidation"`
+	CtxCacheSize               int    `json:"ctx_cache_size"`
+	ABCILogging                bool   `json:"abci_logging"`
+	RelayErrors                bool   `json:"show_relay_errors"`
+	DisableTxEvents            bool   `json:"disable_tx_events"`
+	Cache                      bool   `json:"-"`
+	IavlCacheSize              int64  `json:"iavl_cache_size"`
+	ChainsHotReload            bool   `json:"chains_hot_reload"`
+	GenerateTokenOnStart       bool   `json:"generate_token_on_start"`
+	LeanPocket                 bool   `json:"lean_pocket"`
+	LeanPocketUserKeyFileName  string `json:"lean_pocket_user_key_file"`
 }
 
 func (c PocketConfig) GetLeanPocketUserKeyFilePath() string {
@@ -83,7 +83,7 @@ const (
 	DefaultMaxEvidenceCacheEntries     = 500
 	DefaultListenAddr                  = "tcp://0.0.0.0:"
 	DefaultClientBlockSyncAllowance    = 10
-	DefaultSessionBlockSyncAllowance   = 1 // 1 session (irrespective of num blocks per session)
+	DefaultSessionSyncAllowance        = 1 // 1 session (irrespective of num blocks per session)
 	DefaultJSONSortRelayResponses      = true
 	DefaultTxIndexer                   = "kv"
 	DefaultRPCDisableTransactionEvents = true
@@ -119,36 +119,36 @@ func DefaultConfig(dataDir string) Config {
 	c := Config{
 		TendermintConfig: *config.DefaultConfig(),
 		PocketConfig: PocketConfig{
-			DataDir:                         dataDir,
-			GenesisName:                     DefaultGenesisName,
-			ChainsName:                      DefaultChainsName,
-			EvidenceDBName:                  DefaultEvidenceDBName,
-			TendermintURI:                   DefaultTMURI,
-			KeybaseName:                     DefaultKeybaseName,
-			RPCPort:                         DefaultRPCPort,
-			ClientBlockSyncAllowance:        DefaultClientBlockSyncAllowance,
-			ClientSessionBlockSyncAllowance: DefaultSessionBlockSyncAllowance,
-			MaxEvidenceCacheEntires:         DefaultMaxEvidenceCacheEntries,
-			MaxSessionCacheEntries:          DefaultMaxSessionCacheEntries,
-			JSONSortRelayResponses:          DefaultJSONSortRelayResponses,
-			RemoteCLIURL:                    DefaultRemoteCLIURL,
-			UserAgent:                       DefaultUserAgent,
-			ValidatorCacheSize:              DefaultValidatorCacheSize,
-			ApplicationCacheSize:            DefaultApplicationCacheSize,
-			RPCTimeout:                      DefaultRPCTimeout,
-			PrometheusAddr:                  DefaultPocketPrometheusListenAddr,
-			PrometheusMaxOpenfiles:          DefaultPrometheusMaxOpenFile,
-			MaxClaimAgeForProofRetry:        DefaultMaxClaimProofRetryAge,
-			ProofPrevalidation:              DefaultProofPrevalidation,
-			CtxCacheSize:                    DefaultCtxCacheSize,
-			ABCILogging:                     DefaultABCILogging,
-			RelayErrors:                     DefaultRelayErrors,
-			DisableTxEvents:                 DefaultRPCDisableTransactionEvents,
-			IavlCacheSize:                   DefaultIavlCacheSize,
-			ChainsHotReload:                 DefaultChainHotReload,
-			GenerateTokenOnStart:            DefaultGenerateTokenOnStart,
-			LeanPocket:                      DefaultLeanPocket,
-			LeanPocketUserKeyFileName:       DefaultLeanPocketUserKeyFileName,
+			DataDir:                    dataDir,
+			GenesisName:                DefaultGenesisName,
+			ChainsName:                 DefaultChainsName,
+			EvidenceDBName:             DefaultEvidenceDBName,
+			TendermintURI:              DefaultTMURI,
+			KeybaseName:                DefaultKeybaseName,
+			RPCPort:                    DefaultRPCPort,
+			ClientBlockSyncAllowance:   DefaultClientBlockSyncAllowance,
+			ClientSessionSyncAllowance: DefaultSessionSyncAllowance,
+			MaxEvidenceCacheEntires:    DefaultMaxEvidenceCacheEntries,
+			MaxSessionCacheEntries:     DefaultMaxSessionCacheEntries,
+			JSONSortRelayResponses:     DefaultJSONSortRelayResponses,
+			RemoteCLIURL:               DefaultRemoteCLIURL,
+			UserAgent:                  DefaultUserAgent,
+			ValidatorCacheSize:         DefaultValidatorCacheSize,
+			ApplicationCacheSize:       DefaultApplicationCacheSize,
+			RPCTimeout:                 DefaultRPCTimeout,
+			PrometheusAddr:             DefaultPocketPrometheusListenAddr,
+			PrometheusMaxOpenfiles:     DefaultPrometheusMaxOpenFile,
+			MaxClaimAgeForProofRetry:   DefaultMaxClaimProofRetryAge,
+			ProofPrevalidation:         DefaultProofPrevalidation,
+			CtxCacheSize:               DefaultCtxCacheSize,
+			ABCILogging:                DefaultABCILogging,
+			RelayErrors:                DefaultRelayErrors,
+			DisableTxEvents:            DefaultRPCDisableTransactionEvents,
+			IavlCacheSize:              DefaultIavlCacheSize,
+			ChainsHotReload:            DefaultChainHotReload,
+			GenerateTokenOnStart:       DefaultGenerateTokenOnStart,
+			LeanPocket:                 DefaultLeanPocket,
+			LeanPocketUserKeyFileName:  DefaultLeanPocketUserKeyFileName,
 		},
 	}
 	c.TendermintConfig.LevelDBOptions = config.DefaultLevelDBOpts()
--- a/types/utils.go
+++ b/types/utils.go
@@ -34,6 +34,10 @@ func GetCacheKey(height int, value string) (key string) {
 	return key
 }
 
+func IsBetween(target, minInclusive, maxInclusive int64) bool {
+	return minInclusive <= target && target <= maxInclusive
+}
+
 // SortedJSON takes any JSON and returns it sorted by keys. Also, all white-spaces
 // are removed.
 // This method can be used to canonicalize JSON to be returned by GetSignBytes,
--- a/x/pocketcore/keeper/service.go
+++ b/x/pocketcore/keeper/service.go
@@ -12,8 +12,14 @@ import (
 // HandleRelay handles an api (read/write) request to a non-native (external) blockchain
 func (k Keeper) HandleRelay(ctx sdk.Ctx, relay pc.Relay) (*pc.RelayResponse, sdk.Error) {
 	relayTimeStart := time.Now()
-	// get the latest session block height because this relay will correspond with the latest session
-	sessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)
+
+	sessionBlockHeight := relay.Proof.SessionBlockHeight
+
+	if !k.IsProofSessionHeightWithinTolerance(ctx, sessionBlockHeight) {
+		// For legacy support, we are intentionally returning the invalid block height error.
+		return nil, pc.NewInvalidBlockHeightError(pc.ModuleName)
+	}
+
 	var node *pc.PocketNode
 	// There is reference to node address so that way we don't have to recreate address twice for pre-leanpokt
 	var nodeAddress sdk.Address
--- a/x/pocketcore/keeper/service_test.go
+++ b/x/pocketcore/keeper/service_test.go
@@ -119,7 +119,11 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	ctx, keeper, kvkeys, clientPrivateKey, appPrivateKey, nodePubKey, chain :=
 		setupHandleRelayTest(t)
 
-	// Eliminate the impact of ClientBlockSyncAllowance
+	// Store the original allowances to clean up at the end of this test
+	originalClientBlockSyncAllowance := types.GlobalPocketConfig.ClientBlockSyncAllowance
+	originalClientSessionSyncAllowance := types.GlobalPocketConfig.ClientSessionSyncAllowance
+
+	// Eliminate the impact of ClientBlockSyncAllowance to avoid any relay meta validation errors (OutOfSyncError)
 	types.GlobalPocketConfig.ClientBlockSyncAllowance = 10000
 
 	nodeBlockHeight := ctx.BlockHeight()
@@ -127,6 +131,8 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	latestSessionHeight := keeper.GetLatestSessionBlockHeight(ctx)
 
 	t.Cleanup(func() {
+		types.GlobalPocketConfig.ClientBlockSyncAllowance = originalClientBlockSyncAllowance
+		types.GlobalPocketConfig.ClientSessionSyncAllowance = originalClientSessionSyncAllowance
 		gock.Off() // Flush pending mocks after test execution
 	})
 
--- a/x/pocketcore/keeper/service_test.go
+++ b/x/pocketcore/keeper/service_test.go
@@ -143,10 +143,10 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	mockCtx.On("Logger").Return(ctx.Logger())
 	mockCtx.On("PrevCtx", nodeBlockHeight).Return(ctx, nil)
 
-	AllSessionRangesTests := 4 // The range of block heights we will mock
+	allSessionRangesTests := 4 // The range of block heights we will mock
 
 	// Set up mocks for heights we'll query later.
-	for i := int64(1); i <= blocksPerSesssion*int64(AllSessionRangesTests); i++ {
+	for i := int64(1); i <= blocksPerSesssion*int64(allSessionRangesTests); i++ {
 		mockCtx.On("PrevCtx", nodeBlockHeight-i).Return(ctx, nil)
 		mockCtx.On("PrevCtx", nodeBlockHeight+i).Return(ctx, nil)
 	}
@@ -162,7 +162,7 @@ func TestKeeper_HandleRelay(t *testing.T) {
 		nodePubKey,
 		chain,
 	)
-	assert.Nil(t, err, err)
+	assert.Nil(t, err)
 	assert.NotNil(t, resp)
 	assert.NotEmpty(t, resp)
 	assert.Equal(t, resp.Response, "bar")
@@ -171,7 +171,7 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	// Client is behind or advanced beyond Node's height with ClientSessionSyncAllowance 0
 	// --> CodeInvalidBlockHeightError
 	types.GlobalPocketConfig.ClientSessionSyncAllowance = 0
-	for i := 1; i <= AllSessionRangesTests; i++ {
+	for i := 1; i <= allSessionRangesTests; i++ {
 		resp, err = testRelayAt(
 			t,
 			mockCtx,
@@ -205,33 +205,33 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	// TC2:
 	// Test a relay while one session behind and forward,
 	// while ClientSessionSyncAllowance = 1
-	// --> Success on behind session block height
-	// --> InvalidBlockHeightError on forward session block height
-	SessionRangeTc := 1
-	types.GlobalPocketConfig.ClientSessionSyncAllowance = int64(SessionRangeTc)
+	// --> Success on one session behind
+	// --> InvalidBlockHeightError on one session forward
+	sessionRangeTc := 1
+	types.GlobalPocketConfig.ClientSessionSyncAllowance = int64(sessionRangeTc)
 
 	// First test the minimum boundary
 	resp, err = testRelayAt(
 		t,
 		mockCtx,
 		keeper,
-		latestSessionHeight-blocksPerSesssion*int64(SessionRangeTc),
+		latestSessionHeight-blocksPerSesssion*int64(sessionRangeTc),
 		clientPrivateKey,
 		appPrivateKey,
 		nodePubKey,
 		chain,
 	)
-	assert.Nil(t, err, err)
+	assert.Nil(t, err)
 	assert.NotNil(t, resp)
 	assert.NotEmpty(t, resp)
 	assert.Equal(t, resp.Response, "bar")
 
-	// Second test the maximum boundary
+	// Second test the maximum boundary - Error
 	resp, err = testRelayAt(
 		t,
 		mockCtx,
 		keeper,
-		latestSessionHeight+blocksPerSesssion*int64(SessionRangeTc),
+		latestSessionHeight+blocksPerSesssion*int64(sessionRangeTc),
 		clientPrivateKey,
 		appPrivateKey,
 		nodePubKey,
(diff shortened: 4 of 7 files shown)


nodiesBlade:
Newly added tests
```
=== RUN   TestKeeper_HandleRelay
I[2023-03-16|00:54:58.549] Initializing 1512ec818e72bfec7696f443e2324f79d90b4762 session and evidence cache 
--- PASS: TestKeeper_HandleRelay (0.20s)
=== RUN   TestKeeper_HandleRelayWithinTolerance
I[2023-03-16|00:54:58.726] Initializing 5408648319da126ccd8cce3c5efa44a93a9d8943 session and evidence cache 
--- PASS: TestKeeper_HandleRelayWithinTolerance (0.15s)
=== RUN   TestKeeper_HandleRelayOutsideTolerance
I[2023-03-16|00:54:58.904] Initializing 0e694b1904edc2fbdc13978ec27487a717d51bde session and evidence cache 
I[2023-03-16|00:54:59.058] Initializing d8dbfee56de47cb4c4708e61c012d1b99bb42d8b session and evidence cache 
--- PASS: TestKeeper_HandleRelayOutsideTolerance (0.33s)
PASS

Process finished with the exit code 0
```

nodiesBlade:
Note: some nodes will seal their evidence claim immediately after sesson rollover and this will result in evidence sealed errors. In order to remediate this, we could modify claim and proof tx to only occur on 2,3,4th of a session block instead depending on the nodes address offset. incorporating this network wide will require more due diligence as the reason why this code exists in first place is to prevent too many claims/proofs from being submitted at once in a single block

Olshansk:
@PoktBlade Thanks for doing this!

I haven't reviewed the code but just wanted to ACK that I'm aware and it's on my backlog.

> DISCLAIMER: SHOULD DEFINITELY TEST.......................... USING THE REGULAR RELEASE CYCLE MECHANISMS

I, unfortunately, think that testing and getting confidence in this change is going to be one of the hardest parts, aside from the social coordination of deploying it.

I have two requests:

1. Can you look into whether we can extend the [e2e testing framework](https://github.com/pokt-network/pocket-core/tree/staging/e2e) to support relays and have an integration test that validates this change.

2. I have an [open PR](https://github.com/pokt-network/pocket-core/pull/1535) that enables spinning up a LocaNet and sending relays by specifying altruist endpoints (details are in the docs). 
2.1 Can you use it to validate/test this change?
2.2 Please note that the infra is janky but the best we have (to the best of my knowledge), so I apologize in advance.

nodiesBlade:
Hey,

I have set up my own localnet stack from rigging up my own E2E stack. So I can test to see if session roll over works.

the e2e automatic test suite is new to me, I don't have the bandwidth to test that. It was new and delivered abruptly so no one besides pni has context of that. The previous testing mechanisms included a QA suite of test cases and then a manual regression test using said qa test suite. I can add some scenarios to that, and someone can translate it to the automatic version.

Olshansk:
**tl;dr Noted, I will put the time into this myself.**

--- 

> I have set up my own localnet stack from rigging up my own E2E stack. So I can test to see if session roll over works.

I tried to put together good & easy-to-follow documentation (so it can be used by others), but if you can share the results of your own stack (potentially a video and/or documentation), that would be good too. Ref: https://github.com/pokt-network/pocket-core/blob/fixing_localnet/localnet.md

> the e2e automatic test suite is new to me, I don't have the bandwidth to test that. It was new and delivered abruptly so no one besides pni has context of that

Understood. I tried to make the documentation easy to follow but will invest time in extending, improving and testing it. https://github.com/pokt-network/pocket-core/tree/staging/e2e

types/config.go
```diff
@@ -82,6 +83,7 @@ const (
 	DefaultMaxEvidenceCacheEntries     = 500
 	DefaultListenAddr                  = "tcp://0.0.0.0:"
 	DefaultClientBlockSyncAllowance    = 10
+	DefaultSessionBlockSyncAllowance   = 1
```
Olshansk:
```suggestion
	DefaultSessionBlockSyncAllowance   = 1 // 1 session (irrespective of num blocks per session)
```

types/utils.go
```diff
@@ -34,6 +34,10 @@ func GetCacheKey(height int, value string) (key string) {
 	return key
 }
 
+func IsBetween(target, minInclusive, maxInclusive int64) bool {
+	return target >= minInclusive && target <= maxInclusive
```
Olshansk:
```suggestion
	return minInclusive <= target && target <= maxInclusive
```

Olshansk: NIT: I just feel like its easier to read. I think in `C` land it's lest efficient, but it's funny to discuss the micro-optimization.

x/pocketcore/keeper/service.go
```diff
@@ -12,8 +12,13 @@ import (
 // HandleRelay handles an api (read/write) request to a non-native (external) blockchain
 func (k Keeper) HandleRelay(ctx sdk.Ctx, relay pc.Relay) (*pc.RelayResponse, sdk.Error) {
 	relayTimeStart := time.Now()
-	// get the latest session block height because this relay will correspond with the latest session
-	sessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)
+
+	sessionBlockHeight := relay.Proof.SessionBlockHeight
+
+	if !k.IsLatestSessionHeightWithinTolerance(ctx, sessionBlockHeight) {
+		return nil, pc.NewInvalidBlockHeightError(pc.ModuleName)
```
Olshansk, line 20: Let's take it as an opportunity to give more details to the user. For example:

```
func NewOutsideBlockToleranceHeightError(codespace sdk.CodespaceType, sessionHeight, latestSessionBlockHeight uint64) sdk.Error {
	return sdk.NewError(codespace, fmt.Sprintf("... %d ... %d ..", sessionHeight, latestSessionBlockHeight), InvalidBlockHeightError.Error())
}
```

x/pocketcore/keeper/service.go
```diff
@@ -12,8 +12,13 @@ import (
 // HandleRelay handles an api (read/write) request to a non-native (external) blockchain
 func (k Keeper) HandleRelay(ctx sdk.Ctx, relay pc.Relay) (*pc.RelayResponse, sdk.Error) {
 	relayTimeStart := time.Now()
-	// get the latest session block height because this relay will correspond with the latest session
-	sessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)
+
+	sessionBlockHeight := relay.Proof.SessionBlockHeight
```
Olshansk, line 16: s/sessionBlockHeight/proofSessionHeight

x/pocketcore/keeper/service_test.go
```diff
@@ -38,7 +38,7 @@ func TestKeeper_HandleRelay(t *testing.T) {
 		Path:    "",
 		Headers: nil,
 	}
-	validRelay := types.Relay{
+	validRelay = types.Relay{
 		Payload: p,
 		Meta:    types.RelayMeta{BlockHeight: 976},
```
Olshansk: Please move `976` into a local const and reuse below.

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +66,60 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
 }
+
+func TestKeeper_HandleRelayWithinTolerance(t *testing.T) {
+	defer gock.Off() // Flush pending mocks after test execution
+
+	gock.New("https://www.google.com:443").
+		Post("/").
+		Reply(200).
+		BodyString("bar")
+	keeper, mockCtx, validRelay := setupHandleRelayTest(t, 977)
```
Olshansk: validRelayHeight + 1

x/pocketcore/keeper/service.go
```diff
@@ -12,8 +12,13 @@ import (
 // HandleRelay handles an api (read/write) request to a non-native (external) blockchain
```
Olshansk, line 12: I ran this test (`go test -count=1 -v -run TestKeeper_HandleRelay ./x/pocketcore/keeper/...`) after making the following diff on your branch (`i.e. removing the logic for block height tolerance added in this PR`) which leads me to believe the change was not implemented correctly.

Am I missing something?

```diff
--- a/x/pocketcore/keeper/service.go
+++ b/x/pocketcore/keeper/service.go
@@ -3,21 +3,23 @@ package keeper
 import (
        "encoding/hex"
        "fmt"
+       "time"
+
        "github.com/pokt-network/pocket-core/crypto"
        sdk "github.com/pokt-network/pocket-core/types"
        pc "github.com/pokt-network/pocket-core/x/pocketcore/types"
-       "time"
 )

 // HandleRelay handles an api (read/write) request to a non-native (external) blockchain
 func (k Keeper) HandleRelay(ctx sdk.Ctx, relay pc.Relay) (*pc.RelayResponse, sdk.Error) {
        relayTimeStart := time.Now()

-       sessionBlockHeight := relay.Proof.SessionBlockHeight
+       // sessionBlockHeight := relay.Proof.SessionBlockHeight
+       sessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)

-       if !k.IsLatestSessionHeightWithinTolerance(ctx, sessionBlockHeight) {
-               return nil, pc.NewInvalidBlockHeightError(pc.ModuleName)
-       }
+       // if !k.IsLatestSessionHeightWithinTolerance(ctx, sessionBlockHeight) {
+       //      return nil, pc.NewInvalidBlockHeightError(pc.ModuleName)
+       // }

        var node *pc.PocketNode
        // There is reference to node address so that way we don't have to recreate address twice for pre-leanpokt
diff --git a/x/pocketcore/keeper/session.go b/x/pocketcore/keeper/session.go
index 3225e55b..d0a5ab73 100644
--- a/x/pocketcore/keeper/session.go
+++ b/x/pocketcore/keeper/session.go
@@ -2,6 +2,7 @@ package keeper

 import (
        "encoding/hex"
+
        sdk "github.com/pokt-network/pocket-core/types"
        "github.com/pokt-network/pocket-core/x/nodes/exported"
        "github.com/pokt-network/pocket-core/x/pocketcore/types"
@@ -10,13 +11,9 @@ import (
 // "HandleDispatch" - Handles a client request for their session information
 func (k Keeper) HandleDispatch(ctx sdk.Ctx, header types.SessionHeader) (*types.DispatchResponse, sdk.Error) {
        // retrieve the latest session block height
-       var sessionBlockHeight int64
-       if types.GlobalPocketConfig.ClientSessionSyncAllowance > 0 && header.SessionBlockHeight != 0 {
-               sessionBlockHeight = header.SessionBlockHeight
-       } else {
-               sessionBlockHeight = k.GetLatestSessionBlockHeight(ctx)
-               header.SessionBlockHeight = sessionBlockHeight
-       }
+       latestSessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)
+       // set the session block height
+       header.SessionBlockHeight = latestSessionBlockHeight

        // validate the header
        err := header.ValidateHeader()
@@ -24,7 +21,7 @@ func (k Keeper) HandleDispatch(ctx sdk.Ctx, header types.SessionHeader) (*types.
                return nil, err
        }
        // get the session context
-       sessionCtx, er := ctx.PrevCtx(sessionBlockHeight)
+       sessionCtx, er := ctx.PrevCtx(latestSessionBlockHeight)
        if er != nil {
                return nil, sdk.ErrInternal(er.Error())
        }
```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +66,60 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
 	mockCtx.On("PrevCtx", int64(976)).Return(ctx, nil)
 	mockCtx.On("PrevCtx", keeper.GetLatestSessionBlockHeight(mockCtx)).Return(ctx, nil)
 	mockCtx.On("Logger").Return(ctx.Logger())
+	return
+}
+
+func TestKeeper_HandleRelay(t *testing.T) {
+	defer gock.Off() // Flush pending mocks after test execution
 
+	gock.New("https://www.google.com:443").
```
Olshansk: For my personal learning, can you explain why we need this?

types/config.go
```diff
@@ -18,36 +18,37 @@ type SDKConfig struct {
...
-	LeanPocketUserKeyFileName string `json:"lean_pocket_user_key_file"`
+	DataDir                    string `json:"data_dir"`
+	GenesisName                string `json:"genesis_file"`
+	ChainsName                 string `json:"chains_name"`
+	EvidenceDBName             string `json:"evidence_db_name"`
+	TendermintURI              string `json:"tendermint_uri"`
+	KeybaseName                string `json:"keybase_name"`
+	RPCPort                    string `json:"rpc_port"`
+	ClientBlockSyncAllowance   int    `json:"client_block_sync_allowance"`
+	ClientSessionSyncAllowance int64  `json:"client_session_sync_allowance"`
```
Olshansk, line 29: This feature needs to have a feature switch, otherwise weird things can happen when synching from scratch.

x/pocketcore/keeper/session.go
```diff
@@ -10,16 +10,21 @@ import (
 // "HandleDispatch" - Handles a client request for their session information
 func (k Keeper) HandleDispatch(ctx sdk.Ctx, header types.SessionHeader) (*types.DispatchResponse, sdk.Error) {
 	// retrieve the latest session block height
-	latestSessionBlockHeight := k.GetLatestSessionBlockHeight(ctx)
-	// set the session block height
-	header.SessionBlockHeight = latestSessionBlockHeight
+	var sessionBlockHeight int64
```
Olshansk: I don't think we shold be touching this part of the code at all. Handling the relay with validate the entire proof.

Leaving dispatching as is.

x/pocketcore/keeper/session.go
```diff
@@ -55,6 +60,18 @@ func (k Keeper) IsSessionBlock(ctx sdk.Ctx) bool {
 	return ctx.BlockHeight()%k.posKeeper.BlocksPerSession(ctx) == 1
 }
 
+// IsLatestSessionBlockHeightWithinTolerance checks if the latest session block height is bounded by minSessionBlockHeight <= x <= latestSessionBlockHeight
```
Olshansk: I have less context about session rollover than node runners, so please correct me if I'm wrong below.

---
```go
// "GetLatestSessionBlockHeight" - Returns the latest session block height (first block of the session, (see blocksPerSession))
```

A session rollover issue happens when:
1. Assume the session starts at `height=96` => `latestSessionHeight=96` for any of `[96, 97, 98, 99]`
2. proofBlockHeight can be in `[96, 97, 98, 99]` to be valid 
_* We encounter issues in this approach in the world today*_
Two things can now happen on a follow-up relay and we want to account for both:
- Dispatch session > node session (e.g. 100 > 96)
- Node session  dispatch session (e.g. 100 > 96(


```go
func (k Keeper) IsProofSessionHeightWithinTolerance(ctx sdk.Ctx, proofSessionHeight int64) bool {
	// Session block height can never be zero.
	if proofSessionHeight <= 0 {
		return false
	}
	sessionTolerance := types.GlobalPocketConfig.ClientSessionSyncAllowance * k.posKeeper.BlocksPerSession(ctx) // num of blocks to tolerate
	blocksPerSession = // get this from global var
	minHeight := proofSessionHeight - sessionTolerance * blocksPerSession
	maxHeight := latestSessionHeight + sessionTolerance * blocksPerSession
	return sdk.IsBetween(proofSessionHeight, minHeight, maxHeight)
}
```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +66,60 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+
+func TestKeeper_HandleRelayOutsideTolerance(t *testing.T) {
+	// Each session block is 25 blocks each
+	for _, height := range []int64{926} {
+		gock.New("https://www.google.com:443").
+			Post("/").
+			Reply(200).
+			BodyString("bar")
+		keeper, mockCtx, validRelay := setupHandleRelayTest(t, height)
+		types.GlobalPocketConfig.ClientBlockSyncAllowance = 100
```
Olshansk: You can set this outside of the loop

x/pocketcore/keeper/service_test.go
Olshansk: Also, per my other comment, can you add a test that it works for 976, 977, 978 and 977 when `types.GlobalPocketConfig.ClientBlockSyncAllowance = 0`

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +66,60 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	keeper, mockCtx, validRelay := setupHandleRelayTest(t, 977)
+	resp, err := keeper.HandleRelay(mockCtx, validRelay)
+	assert.Nil(t, err, err)
+	assert.NotNil(t, resp)
+	assert.NotEmpty(t, resp)
+	assert.Equal(t, resp.Response, "bar")
+}
+
+func TestKeeper_HandleRelayOutsideTolerance(t *testing.T) {
+	// Each session block is 25 blocks each
```
Olshansk: Why?
```suggestion
	// Each session is 25 blocks
```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +66,60 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	for _, height := range []int64{926} {
```
Olshansk: Where did this magic number come from? No need to answer, just move to a const or document, etc...

Olshansk (changes requested):

x/pocketcore/keeper/service.go
baaspoolsllc: I intentionally chose to return the same error to keep same functionality. This is important as if we change the error message or code (I know you didn't in this one), then portal operators who have built on top of invalid messages will be forced to change their error handling, and that can involve more work for them and undesired effects.

nodiesBlade: I intentionally chose to return the same error to keep same functionality. This is important as if we change the error message or code (I know you didn't in this one), then portal operators who have built on top of invalid messages will be forced to change their error handling, and that can involve more work for them and undesired effects.

x/pocketcore/keeper/service.go
nodiesBlade: Context of this message?

x/pocketcore/keeper/service_test.go
nodiesBlade: To be honest, it was mostly copy and pasta, but I do have experience in mocking out requests in tests. We don't want to send a real request, certainly not to google, and so we mock the request out with an expected request which allows us to test the intended function.

x/pocketcore/keeper/session.go
nodiesBlade: Can you define proof block height?  What is proof block height in your example?

x/pocketcore/keeper/session.go
nodiesBlade: I modified dispatch with additional functionality, so that way portals can dispatch and retrieve information about old sessions. This is important if the portal wants to account for node's are still behind. It's optional, and coded as such.

Otherwise, there is no RPC call to get app's session information about a certain block height except the latest, which is counterproductive to adding a tolerance.

x/pocketcore/keeper/session.go
nodiesBlade: My initial implementation looked like yours, but ultimately I changed it.

We only want to account for the latter, because if a node height is behind the dispatched one, they will not have the block context to service the relay. I believe this is why I changed it to maxheight to the latest session blockheight that the node is aware of.

Let's just say we do use the max block height tolerance as the maximum range, what if the block height known for that node is `99`? but the actual block height is 100.

x/pocketcore/keeper/service.go
nodiesBlade: Can you provide your test results?

types/config.go
nodiesBlade: This is actually 1 block * sessionsPerBlock, hence why the name is `SessionBlock`. But to your point, I can include this for clarity sakes.

nodiesBlade:
Hey @Olshansk I responded to most of your comments - thanks for taking the time to review it. I think the largest I concern was that you mentioned one of the tests were not running for you after reverting the major core piece of this branch, indicating something might be wrong. Can you provide your test results and we can take it from there, thank you!

Olshansk:
Shouldn't this test fail if I revert the changes you made?

Screenshot 2023-03-28 at 7 58 28 PM

x/pocketcore/keeper/service_test.go
nodiesBlade: I also asked myself the same question when making these tests and was surprised myself.

https://github.com/pokt-network/pocket-core/blob/629a38891850507c4374bdd8d75cbdfd343d469c/x/nodes/types/params.go#L22

It's part of the default values, which I assume is then the default values for the tests since we use global variables instead of something mockable in these tests

x/pocketcore/keeper/session.go
nodiesBlade: maxHeight should actually be the node's latest height here for better accuracy of what i was intending, but relays should always include a session block height.

nodiesBlade:
> Shouldn't this test fail if I revert the changes you made?
> 
> Screenshot 2023-03-28 at 7 58 28 PM

Tests should not fail under the current designs.


## Explanation
So the test scenario is mainly bootstrapped by a function `setupHandleRelayTest` which initializes practically everything that wasn't mockable and whatnot so that these functions actually run without runtime errors. This function also returns a signed relay which is encoded with a session block height of `976` (referred as `relaySessionBlockHeight` from now on), always. Finally, it mocks out the context so that way the current height is set by `currentBlockHeight` in the function signature.

### Change made
In your scenario, you revert my changes, so the network will always grab the latest session block height dictated by `currentBlockHeight`.

### Tests:

#### Test One (Simple handle relay):
1. Block height `976`
2. relay encoded with `relaySessionBlockHeight` as `976`
3. `GetLatestSessionBlockHeight` is still `976`, 
4. Validation passes, relays is handled.

#### Test Two (Within tolernace):
1. Block height `977`
2. relay encoded with `relaySessionBlockHeight` as `976`
3. `GetLatestSessionBlockHeight` is  still 976, 
4. Validation passes, relays is handled.

#### Test Three (outside tolernace):
1. Block height `926`
2. relay encoded with `relaySessionBlockHeight` as `976`
3. `GetLatestSessionBlockHeight` is  **now** 926
4. so the` relayBlockSessionHeight` != `GetLatestSessionBlockHeight`, and therefore invalid block height error assertion matches.

# pls explain
Now you might be asking, what's up with these random/magic numbers in these tests, and why are the tests structured this way. Unfortunately, this is just tech debt within the code base already, and I'm happy to resolve as much as I can and provide clarity as you see fit. What's more important to me is that **you understand the code that I'm writing and you agree that the logic works.** Everything else, I prefer to be a bit more pragmatic and timely on how to solve these PRs. I'll solve as much as I can in the same PR and will suggest moving the unresolved ones to another GH issue/PR. Can you mark the ones as blockers, and the ones as nice to have, and I will prioritize as needed.

P.S - I used delve to debug to see the values on the heap/stack and the function calls trace, good tool to use

x/pocketcore/keeper/session.go
nodiesBlade: Added change for this.

types/config.go
Olshansk: How about `s/DefaultSessionBlockSyncAllowance/DefaultRelayNumSessionsAllowance` then?

Also, updated my comment suggestion.

x/pocketcore/keeper/service.go
Olshansk: Ongoing discussion in main PR thread.

x/pocketcore/keeper/service.go
Olshansk: It's `sed` syntax for "suggestion to replace `sessionBlockHeight` with `proofSessionHeight` so its clearer/easier to read & understand.

x/pocketcore/keeper/service.go
Olshansk: Makes sense, can you just leave a comment in the code along the lines of "Intentionally returning legacy error due to..." similar to the comment you left.

x/pocketcore/keeper/service_test.go
Olshansk: Let's reference `params.DefaultSessionBlocktime` in the comment

x/pocketcore/keeper/session.go
Olshansk: > Can you define proof block height? What is proof block height in your example?

Do you mean `proofSessionHeight`? If so, it's the value returned by `GetLatestSessionBlockHeight`

---

Thoughts on my suggestion above re renaming the function to `IsProofSessionHeightWithinTolerance`? I'm having a hard time wrapping my head around `IsLatestSessionHeightWithinTolerance`.

Could you add a godoc comment with an example please?

Olshansk (changes requested):
> What's more important to me is that you understand the code that I'm writing and you agree that the logic works. 

I spent several hours yesterday looking at this and string to understand it, but I personally still don't have confidence that this works.

What I'd prefer is something similar to what @msmania has in #1534.

1. Add feature switch
2. Show failing test with feature switch off
3. Turn on feature switch
4. Show successful test with feature switch on

> Now you might be asking, what's up with these random/magic numbers in these tests, and why are the tests structured this way. Unfortunately, this is just tech debt within the code base already, and I'm happy to resolve as much as I can and provide clarity as you see fit.

Sorry about that. I'm having a hard time wrapping my head around it, which is why I'm asking for the renaming of certain functions/variables and adding comments where appropriate.

> Can you mark the ones as blockers, and the ones as nice to have, and I will prioritize as needed.

I added follow-up comments and re-reviewed the PR.

1. Please take a look at the my comments and re-request a review once it's ready for another look.
2. If you add me as an editor on your fork (or move the branch to the main repo), I can help make some of the edits myself on this branch.

nodiesBlade:
Hey @Olshansk 

Addressed most of your comments, try giving it a read again with the test cases, it should provide more clarity for your reading. If it takes too long, just message me async and I'll do my best to explain the logic.

nodiesBlade:
Also, you mention we should have a feature flag.. Just a couple notes to that:

1. The tests written previously never accounted for invalid block height tolerance, it was a simple test for the relay test in a happy route, that's why you don't see a 'failure case' test. The `outside tolerance` test adds a invalid block height test for both my changes and before my changes, effectively. 
2. The handleRelay function is purely client sided, it has nothing to do with consensus, so there is not really a need for a true 'feature flag'.

msmania:
Sorry for interrupting the ongoing discussion.  If I understand this patch correctly, the change in `HandleDispatch` is not necessary, right?

nodiesBlade:
> Sorry for interrupting the ongoing discussion. If I understand this patch correctly, the change in `HandleDispatch` is not necessary, right?

I would deem it necessary as if we are going to allow for “sessions to be rolled over” then we should allow the node to dispatch for the old sessions as well.

msmania:
> I would deem it necessary as if we are going to allow for “sessions to be rolled over” then we should allow the node to dispatch for the old sessions as well.

Session role over can be achieved by the change in `HandleRelay`, right?  The other part is not necessary.

When someone, basically the Portal, starts a relay, it should always use the latest session and shouldn't request a relay with an old session because app or node's staking status may have been changed.  And the Portal indeed does that by setting `SessionHeader.SessionBlockHeight = 0` (I think it happens [here](https://github.com/pokt-foundation/portal-api/blob/develop/src/services/pocket-relayer.ts#L636)), so it doesn't change the relay behavior anyway.

What your change in `HandleDispatch` achieves is to enable us to query an old session.  I agree this is convenient, but I think it's out of the scope of this PR.

Moreover, you need to be careful about the second argument of `NewSession`, which must be at most the end of the session of the first argument.  Otherwise it may return a wrong session as described in #1529.

nodiesBlade:
> What your change in `HandleDispatch` achieves is to enable us to query an old session. I agree this is convenient, but I think it's out of the scope of this PR.

Fair enough. Removed it. Portals who want to dispatch can later send in a PR or modify their own PCC to dispatch for an old session.

msmania:
Thanks!  One more ask about test code.  In your test, each testcase calls `setupHandleRelayTest`, creating the whole thing from scratch.  This is not an ideal design.  Also, you should not hardcode block numbers even though they're stored separately.  Please use `ctx.BlockHeight()`, which is 976, and `keeper.BlocksPerSession(ctx)`, which is 25.

I think the test should be something like this.  With this, one setup covers all scenarios and magic numbers are removed completely.

```go
func TestKeeper_HandleRelay(t *testing.T) {
	setupHandleRelayTest(...)

	blockPerSession := keeper.BlocksPerSession(ctx)
	allowance = some positive value

	types.GlobalPocketConfig.ClientSessionBlockSyncAllowance = 0

	// Send relay with older heights without session roll over
	for i := 0; i < blockPerSession*(allowance+1); i++ {
		height := ctx.BlockHeight() - i
		// create a relay payload with `height`
		resp, err := keeper.HandleRelay(mockCtx, validRelay)
		// verify the result accordingly
	}

	types.GlobalPocketConfig.ClientSessionBlockSyncAllowance = allowance

	// Send relay with older heights with session roll over
	for i := 0; i < blockPerSession*(allowance+1); i++ {
		height := ctx.BlockHeight() - i
		// create a relay payload with `height`
		resp, err := keeper.HandleRelay(mockCtx, validRelay)
		// verify the result accordingly
	}
}
```

nodiesBlade:
> Thanks! One more ask about test code. In your test, each testcase calls `setupHandleRelayTest`, creating the whole thing from scratch. This is not an ideal design. Also, you should not hardcode block numbers even though they're stored separately. Please use `ctx.BlockHeight()`, which is 976, and `keeper.BlocksPerSession(ctx)`, which is 25.
> 
> I think the test should be something like this. With this, one setup covers all scenarios and magic numbers are removed completely.
> 
> ```go
> func TestKeeper_HandleRelay(t *testing.T) {
> 	setupHandleRelayTest(...)
> 
> 	blockPerSession := keeper.BlocksPerSession(ctx)
> 	allowance = some positive value
> 
> 	types.GlobalPocketConfig.ClientSessionBlockSyncAllowance = 0
> 
> 	// Send relay with older heights without session roll over
> 	for i := 0; i < blockPerSession*(allowance+1); i++ {
> 		height := ctx.BlockHeight() - i
> 		// create a relay payload with `height`
> 		resp, err := keeper.HandleRelay(mockCtx, validRelay)
> 		// verify the result accordingly
> 	}
> 
> 	types.GlobalPocketConfig.ClientSessionBlockSyncAllowance = allowance
> 
> 	// Send relay with older heights with session roll over
> 	for i := 0; i < blockPerSession*(allowance+1); i++ {
> 		height := ctx.BlockHeight() - i
> 		// create a relay payload with `height`
> 		resp, err := keeper.HandleRelay(mockCtx, validRelay)
> 		// verify the result accordingly
> 	}
> }
> ```

If we were using something like [Ginkgo](https://github.com/onsi/ginkgo) to structure our tests, I'd agree. Otherwise, having the tests separately helps isolate the environments and what we're testing for since these are unit tests and not integration / e2e tests.


> Also, you should not hardcode block numbers even though they're stored separately. Please use `ctx.BlockHeight()`, which is 976, and `keeper.BlocksPerSession(ctx)`, which is 25.

Also agree, but this would require more restructuring. I would push for a follow-up PR to improve tests if we really want to sharpen this old cold base (not worth it for this use case IMO), my bandwidth is quite limited atm to do more refactoring. If it's an absolute must, then please let me know and I'll find time to do it as I'd like see pokt continue to improve. (Or if you'd like to set up these changes, that'd be great too)

x/pocketcore/keeper/session.go
```diff
@@ -55,6 +55,18 @@ func (k Keeper) IsSessionBlock(ctx sdk.Ctx) bool {
 	return ctx.BlockHeight()%k.posKeeper.BlocksPerSession(ctx) == 1
 }
 
+// IsProofSessionHeightWithinTolerance checks if the latest session block height is bounded by minSessionBlockHeight <= x <= latestBlockHeight
```
Olshansk:
```suggestion
// IsProofSessionHeightWithinTolerance checks if the relaySessionBlockHeight is bounded by (latestSessionBlockHeight - tolerance ) <= x <= latestBlockHeight
```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
-	mockCtx.On("PrevCtx", int64(976)).Return(ctx, nil)
+	mockCtx.On("BlockHeight").Return(currentBlockHeight)
+	mockCtx.On("PrevCtx", relayProofSessionBlockHeight).Return(ctx, nil)
 	mockCtx.On("PrevCtx", keeper.GetLatestSessionBlockHeight(mockCtx)).Return(ctx, nil)
 	mockCtx.On("Logger").Return(ctx.Logger())
+	return
+}
+
+func TestKeeper_HandleRelay(t *testing.T) {
+
```
Olshansk:
```suggestion
```

x/pocketcore/keeper/service_test.go
```diff
@@ -2,20 +2,25 @@ package keeper
...
+	"gopkg.in/h2non/gock.v1"
 	"testing"
 
 	sdk "github.com/pokt-network/pocket-core/types"
 	appsKeeper "github.com/pokt-network/pocket-core/x/apps/keeper"
 	appsTypes "github.com/pokt-network/pocket-core/x/apps/types"
 	"github.com/pokt-network/pocket-core/x/pocketcore/types"
 	"github.com/stretchr/testify/assert"
-	"gopkg.in/h2non/gock.v1"
 )
```
Olshansk:
```suggestion
// TODO(@olshansky, @msmania): Improve the tests per the suggestions from @msmania here: https://github.com/pokt-network/pocket-core/pull/1536#issuecomment-1494964711

```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	assert.Nil(t, err, err)
+	assert.NotNil(t, resp)
+	assert.NotEmpty(t, resp)
+	assert.Equal(t, resp.Response, "bar")
+}
+
+func TestKeeper_HandleRelayWithinTolerance(t *testing.T) {
+	defer gock.Off() // Flush pending mocks after test execution
+
+	// TC - relay proof is within the same session block
```
Olshansk:
```suggestion
	// tc - relay proof is from the previous session block assuming a 4 session block size
```

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	tc := relayTc{
+		blockHeight:                  976,
+		relayProofSessionBlockHeight: 975,
```
Olshansk: Mind adding one for 951 as well (boundary) + comment that we're using `DefaultSessionblockHeight`

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+}
+
+func TestKeeper_HandleRelay(t *testing.T) {
+
+	defer gock.Off() // Flush pending mocks after test execution
+
+	// tc - height is same as session block
+	tc := relayTc{
+		blockHeight:                  976,
+		relayProofSessionBlockHeight: 976,
```
Olshansk: Mind adding one for 1000 as well (boundary) + comment that we're using `DefaultSessionblockHeight`

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	assert.Equal(t, resp.Response, "bar")
+}
+
+func TestKeeper_HandleRelayOutsideTolerance(t *testing.T) {
+	// Each session block is 25 blocks by default genesis params found in DefaultSessionBlocktime, pocket-core/x/nodes/types/params.go
+	tcs := []relayTc{
+		{
+			// tc 1: relay proof ahead of current block height
+			blockHeight:                  926,
+			relayProofSessionBlockHeight: 976,
```
Olshansk: Should we use 951 so we're testing the boundary edge case?

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +71,142 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+		Reply(200).
+		BodyString("bar")
+	keeper, mockCtx, validRelay := setupHandleRelayTest(t, successTc.blockHeight, successTc.relayProofSessionBlockHeight)
 	resp, err := keeper.HandleRelay(mockCtx, validRelay)
 	assert.Nil(t, err, err)
 	assert.NotNil(t, resp)
 	assert.NotEmpty(t, resp)
 	assert.Equal(t, resp.Response, "bar")
+
+	types.GlobalPocketConfig.ClientBlockSyncAllowance = clientBlockSyncAllowanceTmp
```
Olshansk: Move this into a `t.cleanup()` below the place where you create the `tmp.

Ditto above

Olshansk (changes requested):
@PoktBlade A few more minor comments / NITs + I asked to leave a TODO per @msmania's suggestion so him or I can do it in a separate PR.

Two questions:

1. Does the following diagram capture (visually) the changes made in a simple way in your opinion? Lmk if I should change anything

```mermaid
sequenceDiagram
    title 4 Blocks Per Session

    participant C as Client
    participant N as Node(h=100)

    alt 1. before change - ok
        C->>+N: Relay(100<=h<=103)
        N->>-C: Response + Reward
    else 2. before change - err
        C->>+N: Relay(h<100 || h > 103)
        N->>-C: Err
    else 3. after change - ok
        C->>+N: Relay(100<=h<=103)
        N->>-C: Response + Reward
        C->>+N: Relay(96<=h<=99)
        N->>-C: Response + ?? NoReward??
    else 4. before change - err
        C->>+N: Relay(h<96 || h > 103)
        N->>-C: Err                
    end
```

2. Am I correct to understand that in case (3) above, the servicer won't be rewarded for serving a relay from a previous session

3. I was looking at `x/pocketcore/types/service.go` and trying to understand why we don't need to change the relay meta validation. Could you help me understand this?

```go
func (m RelayMeta) Validate(ctx sdk.Ctx) sdk.Error {
	// ensures the block height is within the acceptable range
	if ctx.BlockHeight()+int64(GlobalPocketConfig.ClientBlockSyncAllowance) < m.BlockHeight || ctx.BlockHeight()-int64(GlobalPocketConfig.ClientBlockSyncAllowance) > m.BlockHeight {
		return NewOutOfSyncRequestError(ModuleName)
	}
	return nil
}
```

Olshansk:
@PoktBlade Okay to close this one out in favor of #1529 so we can finish it off there?

x/pocketcore/keeper/service_test.go
baaspoolsllc: this test was the original test before my additions / not meant to test anything but the happy route

x/pocketcore/keeper/service_test.go
nodiesBlade: I noticed the test values here were incorrect

Are you asking me to put a test for 950?

x/pocketcore/keeper/session.go
nodiesBlade: I've adjusted this to the max range to be `latestSessionHeight` which is what my tests reflects as well. There is no reason to believe that the relay session block height would ever be above latest session block height.

if relay proof session block height is above latest session block height, then the app sent in a wrong app session height or the app dispatcher is ahead of the servicer node, which should result in an error

nodiesBlade:
for your diagram, can you include the session block height for more clarity? P.S, I did change the max bounds back to `latest session block height`

2) all the relays are still via client side validation, so the client should still acknowledge that it is a valid relay if it's from a previous sessions within bounds and store it as part of its evidence storage.
The time to submit a claim is bounded by [ClaimSubmissionWindow](https://docs.pokt.network/learn/protocol-parameters/#claimsubmissionwindow), so as long as you submit your claim within this window, you will be rewarded

3) I'm not sure what the purpose of relay metadata is. Perhaps one of the reasons it is there to prevent nodes from serving a relay that has to do a large historical lookup (i.e block 1). I ended up making it its own client parameter. Both also serve a purpose to check if the node is in sync in its own ways.

x/pocketcore/keeper/service_test.go
nodiesBlade: 976 should be the new boundary since i've modified the bounds, (see my other comment in regards to this)

note: this test was the original test before my additions, it's not really used to test my changes directly,

x/pocketcore/keeper/service_test.go
nodiesBlade: i modified the changes and added more descriptive comments on each tc

x/pocketcore/keeper/service_test.go
```diff
@@ -107,40 +109,55 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
-			// tc 2: relay proof two sessions behind
-			blockHeight:                  926,
-			relayProofSessionBlockHeight: 900,
+			// tc 2: relay proof session height is not within tolerance (min boundary exceeded)
+			blockHeight:                  976,
+			relayProofSessionBlockHeight: 926,
 		},
 	}
 	clientBlockSyncAllowanceTmp := types.GlobalPocketConfig.ClientBlockSyncAllowance // hold the initial value before mutating, this should not matter anyways since tests concurrency = 1
 	types.GlobalPocketConfig.ClientBlockSyncAllowance = 100                          // Allows the client block sync tolerance to go above 100 blocks for test purposes
```
Olshansk: Why do we need it to go above 100 for test purposes even though we don't modify it in `TestKeeper_HandleRelayWithinTolerance`? Ditto for other tests.

x/pocketcore/keeper/service_test.go
```diff
@@ -66,25 +72,157 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+		resp, err := keeper.HandleRelay(mockCtx, validRelay)
+		assert.Nil(t, err, err)
+		assert.NotNil(t, resp)
+		assert.NotEmpty(t, resp)
+		assert.Equal(t, resp.Response, "bar")
+	}
+}
+
+func TestKeeper_HandleRelayOutsideTolerance(t *testing.T) {
+	// Each session block is 25 blocks by default genesis params found in DefaultSessionBlocktime, pocket-core/x/nodes/types/params.go
```
Olshansk: Is this true given that we're modifying `ClientBlockSyncAllowance` below? Ditto for other tests.

Olshansk (changes requested):
@PoktBlade Sorry for the back & forth. This PR is simply the one I'm most nervous about because of all the numbers and parameters we're changing, and due to my lack of understanding of the complete picture. I want to be ultra-sure we're not missing anything and avoiding "off by 1" errors.

> all the relays are still via client side validation, so the client should still acknowledge that it is a valid relay if it's from a previous sessions within bounds and store it as part of its evidence storage.
The time to submit a claim is bounded by [ClaimSubmissionWindow](https://docs.pokt.network/learn/protocol-parameters/#claimsubmissionwindow), so as long as you submit your claim within this window, you will be rewarded

Does this mean there is no protocol level validation that a servicer is submitting rewards only for the session where the relays came from?

---
@msmania Would appreciate your 👀  on this PR as well.

nodiesBlade:
> Does this mean there is no protocol level validation that a servicer is submitting rewards only for the session where the relays came from?

No, that's not what I am saying. HandleRelay validates a relay proof (checks if it's signed by AAT, the _correct session block height,_ etc), serves the relay request, then stores the relay proof inside a LevelDB database until it's ready to submit a claim. This entire function end to end is done locally and not broadcasted to the blockchain. The network is unaware until the claim/proof is submitted. The claim can be submitted after the session has ended, and then the servicer has up until the ClaimSubmissionWindow to submit the claim before it is considered _matured_ (invalid). 

The relay proof itself is validated to match the session it belongs in whenever it comes time for the proof tx, where a randomly selected leaf is selected, and the leaf's relay proof session block height itself is validated against the claim's session.


I think the best resources to get you up to speed on this is:
1. The [Claim/Proof TX](https://github.com/pokt-network/pocket-core/blob/staging/doc/specs/reward_protocol.md) documentation
2. The [code where a claim is considered mature](https://github.com/pokt-network/pocket-core/blob/2c33c65c11ae97cffe4d392866d18284c858f311/x/pocketcore/keeper/claim.go#L65) (passed the claim submission window)
3. The leaf validation code [1](https://github.com/pokt-network/pocket-core/blob/2c33c65c11ae97cffe4d392866d18284c858f311/x/pocketcore/keeper/proof.go#L163), [2](https://github.com/pokt-network/pocket-core/blob/staging/x/pocketcore/types/proof.go#L82) where it matches the session block height

x/pocketcore/keeper/service_test.go
nodiesBlade: Yes, session block intervals is an on chain parameter, whereas `ClientBlockSyncAllowance` is a client specific parameter.

x/pocketcore/keeper/service_test.go
nodiesBlade: It was just a conservative number, the default is 10. I can shrink this, but it does not really impact the relay test, unless the ClientBlockSyncAllowance is < 25.

msmania:
I updated the diagram.  @PoktBlade Can you confirm this is correct?

sequenceDiagram
    title 4 Blocks Per Session

    participant C as Client
    participant N as Node(h=100 sbh=97)

    alt w/o change or tolerance=0
        C->>+N: Relay for sbh=97
        N->>-C: Ok
    else w/o change or tolerance=0
        C->>+N: Relay for sbh!=97
        N->>-C: InvalidBlockHeightError
    else tolerance=2
        C->>+N: Relay for sbh=97
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*1
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*2
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*3
        N->>-C: InvalidBlockHeightError
        C->>+N: Relay for sbh=97+4*1
        N->>-C: InvalidBlockHeightError
    end

msmania:
I updated the diagram.  @PoktBlade Can you confirm this is correct?

```mermaid
sequenceDiagram
    title 4 Blocks Per Session

    participant C as Client
    participant N as Node(h=100 sbh=97)

    alt w/o change or tolerance=0
        C->>+N: Relay for sbh=97
        N->>-C: Ok
    else w/o change or tolerance=0
        C->>+N: Relay for sbh!=97
        N->>-C: InvalidBlockHeightError
    else tolerance=2
        C->>+N: Relay for sbh=97
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*1
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*2
        N->>-C: Ok
        C->>+N: Relay for sbh=97-4*3
        N->>-C: InvalidBlockHeightError
        C->>+N: Relay for sbh=97+4*1
        N->>-C: InvalidBlockHeightError
    end
```

*sbh = Session Block Height

x/pocketcore/keeper/service_test.go
msmania: I agree `ClientBlockSyncAllowance` is confusing.  As @PoktBlade says, it's client-specific and stored in config.json, which means a node runner can set it to a big number like 100 or 100000000 to make `RelayMeta.Validate` always pass.

But, it doesn't impact the Pocket network at all because as @PoktBlade mentions below, relays are handled just locally.  From app or the portal's perspective, the relay result depends on which node the relay is dispatched to, but it doesn't impact the reward mechanism until the node submits a claim tx.  And, once a claim tx is submitted, what the network cares about is whether a relay is processed by a valid node in the corresponding session or not.  It doesn't care whether the node was synced or not when the relay was processed.

Anyway, it makes sense to me that we eliminate the impact of `ClientBlockSyncAllowance` by setting it to a big number like 100.

nodiesBlade:
> I updated the diagram. @PoktBlade Can you confirm this is correct?
> 
> 
> *sbh = Session Block Height

Looks good!

Olshansk:
@PoktBlade @msmania My main question has to do with validation of the RelayProof.

Assume:
* `height=100`
* `sbh=97`
* `tolerance=2`
* `relayHeight = 97-4*1=93`

1. Node services the request
2. Node stores it in local memory
3. Node submits claim
4. Chain requires proof for leaf with relay of height 93
5. Node submits proof
6.`93 != 97` because rp.SeeionBlogHeight != sessionBlockHeight so even though the relay was serviced, the node will not get rewarded

What am I missing here?

```go
// "Validate" - Validates the relay proof object
func (rp RelayProof) Validate(appSupportedBlockchains []string, sessionNodeCount int, sessionBlockHeight int64) sdk.Error {
	// validate the session block height
	if rp.SessionBlockHeight != sessionBlockHeight {
		return NewInvalidBlockHeightError(ModuleName)
	}
```

msmania:
> @PoktBlade @msmania My main question has to do with validation of the RelayProof.

I think you mean [this check](https://github.com/pokt-network/pocket-core/blob/staging/x/pocketcore/keeper/proof.go#L163), where a proof tx handler validates a random-picked proof.  As you can see, the height check is against the sbh of the corresponding claim.  With session roll over, a node at height 100 accepts a relay for sbh=93 and puts it in an evidence of sbh=93, not 97.  After that, nobody knows that relay was a latecomer.

msmania:
Now we have two parameters `ClientBlockSyncAllowance` and `ClientSessionBlockSyncAllowance`.  The former is block, and the latter is session.  It's very confusing.  `ClientSessionBlockSyncAllowance` should be block too, or renamed to `ClientSessionSyncAllowance` or something.

Olshansk:
Capturing offline discussion.

## Code References

1. [Here](https://github.com/pokt-network/pocket-core/blob/staging/x/pocketcore/keeper/service.go#L65) is where we store the relay in the local DB.
	- Note that it uses the metadata of the relay, not the node

2. [Here](https://github.com/pokt-network/pocket-core/blob/f0ce1b86a76cd40b9047351149a0a8bb9332db39/x/pocketcore/module.go#L142) is where we submit the claim after which point no more relays for that session be handled.
	- Note that it is random to smooth out the time at which the claim is submitted so its not all at once
	- There is potential opportunity to make this client configurable in the future

3. For reference, [here](https://github.com/pokt-network/pocket-core/blob/staging/x/pocketcore/keeper/session.go#L59) is where we compute session heights
	- Important to note that the division implicitly rounds to an int

## Frame of thinking

On-chain limitations
	- Earliest time to submit claim is on-chain
	- Latest time to submit claim is on-chain

Processing
	- Multiple trees being built at once rather than one at a time
	- A session is a grouping (bounded by the time limitations above) rather than a timing mechanism

Olshansk (dismissed):
I'd prefer to rename the parameter @msmania's suggestion [here](https://github.com/pokt-network/pocket-core/pull/1536#issuecomment-1499463573), but it can be done in a followup PR and not a blocker.

nodiesBlade:
Squashed all my initial commits to rebase easily

* updated test cases for out of range and in range. 
* renamed `client_session_block_sync_allowance` to `client_session_sync_allowance`
* Updated the update config command to include a default allowance (a small detail I remembered from LP, not sure if this is even used)
* Updated quickstart read.me


I believe this is good to merge in. @Olshansk, merge in if you think so as well!

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	assert.Nil(t, resp)
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Codespace(), sdk.CodespaceType(types.ModuleName))
+	assert.Equal(t, err.Code(), sdk.CodeType(types.CodeInvalidBlockHeightError))
+
+	// TC2:
+	// Test a relay while two sessions behind and forward,
+	// while ClientSessionSyncAllowance = 1
+	// --> InvalidBlockHeightError on forward session block height
+	// --> InvalidBlockHeightError on forward session block height
```
msmania: duplicated line

msmania:
I'd move `IsProofSessionHeightWithinTolerance` to service.go, but I'm just fine.  This patch all looks good to me.  Thanks!

x/pocketcore/keeper/service_test.go
```diff
@@ -137,8 +142,11 @@ func TestKeeper_HandleRelay(t *testing.T) {
 	mockCtx.On("BlockHeight").Return(ctx.BlockHeight())
 	mockCtx.On("Logger").Return(ctx.Logger())
 	mockCtx.On("PrevCtx", nodeBlockHeight).Return(ctx, nil)
+
+	AllSessionRangesTests := 4 // The range of block heights we will mock
```
Olshansk: s/allSessionRangesTests/

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+		t,
+		mockCtx,
+		keeper,
+		latestSessionHeight-blocksPerSesssion*int64(SessionRangeTc),
+		clientPrivateKey,
+		appPrivateKey,
+		nodePubKey,
+		chain,
+	)
+	assert.Nil(t, err, err)
```
Olshansk: NIT: The error will show up if it's not nil

```suggestion
	assert.Nil(t, err)
```

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+
+	// TC2:
+	// Test a relay while one session behind and forward,
+	// while ClientSessionSyncAllowance = 1
+	// --> Success on behind session block height
+	// --> InvalidBlockHeightError on forward session block height
+	SessionRangeTc := 1
+	types.GlobalPocketConfig.ClientSessionSyncAllowance = int64(SessionRangeTc)
+
+	// First test the minimum boundary
```
Olshansk, line 213:
```suggestion
	// First test the minimum boundary - Success
```

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+		appPrivateKey,
+		nodePubKey,
+		chain,
+	)
+	assert.Nil(t, err, err)
+	assert.NotNil(t, resp)
+	assert.NotEmpty(t, resp)
+	assert.Equal(t, resp.Response, "bar")
+
+	// Second test the maximum boundary
```
Olshansk:
```suggestion
	// Second test the maximum boundary - Error
```

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+
+	// TC2:
+	// Test a relay while two sessions behind and forward,
+	// while ClientSessionSyncAllowance = 1
+	// --> InvalidBlockHeightError on forward session block height
+	// --> InvalidBlockHeightError on forward session block height
+	SessionRangeTc = 2
+	types.GlobalPocketConfig.ClientSessionSyncAllowance = 1
+
+	resp, err = testRelayAt(
```
Olshansk, line 253: // First test the two sessions back - Error

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+		appPrivateKey,
+		nodePubKey,
+		chain,
+	)
+	assert.Nil(t, resp)
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Codespace(), sdk.CodespaceType(types.ModuleName))
+	assert.Equal(t, err.Code(), sdk.CodeType(types.CodeInvalidBlockHeightError))
+
+	// Second test the maximum boundary
```
Olshansk:
```suggestion
	// Second test the maximum boundary - Error
```

x/pocketcore/keeper/service_test.go
```diff
@@ -191,4 +201,83 @@ func TestKeeper_HandleRelay(t *testing.T) {
...
+	)
+	assert.Nil(t, resp)
+	assert.NotNil(t, err)
+	assert.Equal(t, err.Codespace(), sdk.CodespaceType(types.ModuleName))
+	assert.Equal(t, err.Code(), sdk.CodeType(types.CodeInvalidBlockHeightError))
+
+	// TC2:
+	// Test a relay while two sessions behind and forward,
+	// while ClientSessionSyncAllowance = 1
+	// --> InvalidBlockHeightError on forward session block height
```
Olshansk:
```suggestion
	// --> InvalidBlockHeightError on two sessions back block height
```

x/pocketcore/keeper/service_test.go
Olshansk:
```suggestion
	// --> InvalidBlockHeightError on two sessions forward block height
```

Olshansk (changes requested):
Just some nits related to comments. Will approve after.

---

Unrelated and optional: @PoktBlade, sharing personal experience: Rebasing has worked well for me in other environments (e.g. big companies) and platforms (e.g. gerrit). But I've found that because of a lot of factors (OSS, github, PR flow, squash & merging to main, etc...), I just always `"merge with main" and resolve conflicts.

nodiesBlade:
Resolved the nits, good to go again.

msmania (approved):
Thank you for your patience!

Olshansk (approved):
Thanks @PoktBlade. Feel free to squash & merge into `staging`!

I'll also take a stab at documenting all of the great discussions & notes we have here and will send your way to review by EOW. 🙏

POKT-Discourse:

https://forum.pokt.network/t/into-the-gateway-verse-bootstrapping-a-multi-gateway-ecosystem/4532/1

PR #1543, opened 2023-04-07
Explicitly set TestMode=0 in TestValidatorStateChange_OutputAddressEdit

msmania:
We cannot trust the current state as it may have been modified by previous tests and not reverted to the original value.

### DIFF
--- a/x/nodes/keeper/reward_test.go
+++ b/x/nodes/keeper/reward_test.go
@@ -1,9 +1,9 @@
 package keeper
 
 import (
-	"github.com/pokt-network/pocket-core/codec"
 	"testing"
 
+	"github.com/pokt-network/pocket-core/codec"
 	sdk "github.com/pokt-network/pocket-core/types"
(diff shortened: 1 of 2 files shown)


Olshansk:
@msmania Seems like it's still failing: https://app.circleci.com/pipelines/github/pokt-network/pocket-core/2811/workflows/fdf09d63-cf8f-45a2-bb3f-4a1fc54d0a00/jobs/5260/parallel-runs/0/steps/0-102

msmania:
> @msmania Seems like it's still failing: https://app.circleci.com/pipelines/github/pokt-network/pocket-core/2811/workflows/fdf09d63-cf8f-45a2-bb3f-4a1fc54d0a00/jobs/5260/parallel-runs/0/steps/0-102

Maybe adding a single test broke the testing order..?  Probably the root cause is the same.  Let's fix them all.

msmania:
All passed!

Olshansk (approved):
Seems like through these reviews we have found a hack for global variables in tests when `p=1` :)

PR #1556, opened 2023-06-07
Improve version string comparison

msmania:
This patch changes the logic to compare the version strings to see if the network has been upgraded or not.  The simple string comparison does not work when the network is upgraded from "0.9" to "0.10".  The proposed fix is to convert a version string into a slice of integers before comparison.

### Summary generated by Reviewpad on 08 Jun 23 00:08 UTC
This pull request includes two patches. 
The first patch improves the version string comparison in `types/utils.go` by adding a new function called `CompareVersionStrings`. This function compares two version strings as if they are integers and returns negative, positive, or zero if the first string is earlier, later, or equal to the second string respectively. It also includes a new unit test for the function.
The second patch addresses some of the previous comments and adds a new test case for the `CompareVersionStrings` function.

### DIFF
--- a/types/utils.go
+++ b/types/utils.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	dbm "github.com/tendermint/tm-db"
@@ -153,4 +154,46 @@ func ContainsString(s []string, e string) bool {
 		}
 	}
 	return false
+}
+
+func CompareVersionStrings(verStr1, verStr2 string) (int, error) {
+	ver1 := strings.Split(verStr1, ".")
+	ver2 := strings.Split(verStr2, ".")
+	lenVer1 := len(ver1)
+	lenVer2 := len(ver2)
+
+	numChunks := lenVer1
+	if lenVer2 < numChunks {
+		numChunks = lenVer2
+	}
+
+	for i := 0; i < numChunks; i++ {
+		verNum1, err := strconv.Atoi(ver1[i])
+		if err != nil {
+			return 0, err
+		}
+
+		verNum2, err := strconv.Atoi(ver2[i])
+		if err != nil {
+			return 0, err
+		}
+
+		if verNum1 < verNum2 {
+			return -1, nil
+		}
+
+		if verNum1 > verNum2 {
+			return 1, nil
+		}
+	}
+
+	if lenVer2 > numChunks {
+		return -1, nil
+	}
+
+	if lenVer1 > numChunks {
+		return 1, nil
+	}
+
+	return 0, nil
 }
--- a/types/utils.go
+++ b/types/utils.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


types/utils_test.go
```diff
@@ -64,3 +65,26 @@ func TestTimeFormatAndParse(t *testing.T) {
 		require.Equal(t, timeFromRFC.Format(SortableTimeFormat), tc.SDKSortableTimeStr)
 	}
 }
+func Test_compareVersionStrings(t *testing.T) {
+	comp, err := CompareVersionStrings("0.9.1.1", "0.10.0")
```
Olshansk, line 69: Please add one test to assert that "0.10.0" > "0.9.2"

x/gov/module.go
```diff
@@ -135,8 +135,18 @@ func (am AppModule) BeginBlock(ctx sdk.Ctx, req abci.RequestBeginBlock) {
...
 	u := am.keeper.GetUpgrade(ctx)
 
+	verComp, err := sdk.CompareVersionStrings(ctx.AppVersion(), u.Version)
+	if err != nil {
+		ctx.Logger().Error("INVALID VERSION STRING",
+			"ctx", ctx.AppVersion(),
+			"upgrade", u.Version,
+			"err", err.Error(),
+		)
+		os.Exit(1)
```
Olshansk, line 145: can we panic or do a Fatal log above instead of os.Exit(1)?

types/utils.go
```diff
@@ -154,3 +155,45 @@ func ContainsString(s []string, e string) bool {
 	}
 	return false
 }
+
+func CompareVersionStrings(verStr1, verStr2 string) (int, error) {
```
Olshansk, line 166: Please add a comment on what this does and let the reader know to look at the tests for reference

types/utils.go
```diff
@@ -154,3 +155,45 @@ func ContainsString(s []string, e string) bool {
...
+	ver1 := strings.Split(verStr1, ".")
+	ver2 := strings.Split(verStr2, ".")
+	lenVer1 := len(ver1)
+	lenVer2 := len(ver2)
+
+	numChunks := lenVer1
```
Olshansk, line 172: can we just do a math.Min operation?

types/utils.go
```diff
@@ -154,3 +155,45 @@ func ContainsString(s []string, e string) bool {
...
+	if lenVer2 < numChunks {
+		numChunks = lenVer2
+	}
+
+	for i := 0; i < numChunks; i++ {
```
Olshansk, line 177: Very clear implementation. Ty!

Olshansk (changes requested):
A few minor requests but nothing blocking

x/gov/module.go
msmania: `ctx.Logger()` is tendermint's logger that does not have `Fatal`.  The combination of `Error` and `os.Exit(1)` is consistent with the below cases when `os.FindProcess` or `p.Signal` fails.

types/utils.go
msmania: `math.Min` is for float64.  I think golang does not have a built-in min function for integers.

Olshansk (approved):

PR #1557, opened 2023-06-07
[CI/Containers] Add image build workflow

okdas:
## Description

With our decision to switch to GitHub Packages as the Container Registry for v0, it becomes easier to use the workflow to push the image.

To ensure a seamless transition, I have used an existing multi-arch build as a base. This means that existing deployments should not be affected or break due to the changes.

To validate the compatibility, I conducted tests on my own node by replacing the `poktnetwork/pocket-core:RC-0.9.2` image with the one generated by this workflow. The tests indicated that everything functioned as expected, without any issues.

Registry link: https://github.com/pokt-network/pocket-core/pkgs/container/pocket-v0

### Summary generated by Reviewpad on 07 Jun 23 22:17 UTC
This pull request adds an image build workflow, with a Dockerfile, build-images.yaml, and entrypoint.sh. The workflow handles build and push of images to GitHub Container Registry, while staging is set as the main branch. Additionally, the patch checks if it works with PR events.

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,18 @@
+# Based on a previous implementation to make sure we don't break existing deployments.
+# https://github.com/pokt-network/pocket-core-deployments/blob/staging/docker/Dockerfile
+
+FROM golang:1.17-alpine as build
+RUN apk add --no-cache ca-certificates
+WORKDIR /build
+ADD . .
+RUN go build -o pocket app/cmd/pocket_core/main.go
+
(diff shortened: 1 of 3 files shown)


Olshansk (approved):
Thanks @okdas!

Did not review the code in detail

okdas:
Thank you for the review. This should work just fine for release tags as well, but there's no guarantee due to the difference in the type of the GitHub Actions event. In the worst case scenario, I'll need to open another PR to fix any potential issues. :)

PR #1574, opened 2023-07-17
Add Proof tx submission enhancement/bug

nodiesBlade:
## Description

Whenever the evidence does not match the total proofs, the node tries to delete the evidence and submits a claim. It should never submit a proof whenever there is LESS than the number of relay proofs as that can cause an array index of out-bounds exception if the randomly selected index is more than what we have in the store. This can cause a node crash.

**Example: Servicer starts off with 10 relays in the evidence store**
```
1. Session ends
2. Servicer submits claim for 10 relays on chain
4. Servicer stops node and restarts evidence.db & evidence.db gets corrupted
5. Servicer will attempt to submit a proof, and psuedo-random selected index for merkle proof generation can result in AIOB
```

**Solution:**
If the number of relay proofs in the evidence store is less than the submitted claims total relay proofs then delete the evidence and don't submit a proof.

---
There is a potential race condition where the evidence is not sealed while submitting a claim, allowing for more relays to enter the evidence claim. Whenever a proof is generated via `GenerateMerkleRoot`, it will result in an incorrect Merkle proof from being generated due to a mismatch of how many relays are in the store and how many was submitted on chain. This can inadvertently happen more often as we accept session rollover relays

**Example: Servicer starts off with 10 relays in the evidence store**
```
1. Session ends
2. Servicer handles a relay, while also submitting the claim (RACE Condition, no proper locking mechanism here)
3. Servicer submits claim for 10 relays on chain
4. Servicer finishes handling the relay and evidence store now contains 11 relays
5. Servicer (after waiting a couple of blocks) submits a proof tx with a merkle proof for 11 relays, not 10
6. Validators will mark the TX as invalid due to invalid merkle proof.
```
**Solution**
- In order to allow for a best-effort process to submit proof for the claim the servicer submitted, we will calculate the `maxRelays` based off:
  1. Use the claim's total proof if the proof count doesn't exceed the maximum relays per session.
  6. Otherwise, fallback to the maximum relays per session.
  
The `GenerateMerkleRoot` already accepts a `maxRelays` parameter which was introduced to fix the Chocalate Rain/Overservicing vulnerability. It discards relays above maxRelays with the following implementation
```go
	if int64(len(ev.Proofs)) > maxRelays {
		ev.Proofs = ev.Proofs[:maxRelays]
		ev.NumOfProofs = maxRelays
	}
```  
---
Note: The more ideal solution is to only submit the claim whenever we are done servicing for sure, but this involves locks and potential side effects that Pocket core was never really built with in mind

Note: Something tells me that we were missing a `continue` statement in the old code whenever the evidence does not match the total proofs in the claim. It shouldn't submit if there is a mismatch / if we deleted the evidence. My solution adds a `continue` for less than, and tries a best effort to submit proof if we do have enough relay proofs in our evidence store.

### Summary generated by Reviewpad on 17 Jul 23 19:44 UTC
This pull request adds better proof validation to the `proof.go` file in the `x/pocketcore/keeper` package. It includes changes to handle a potential race condition where the evidence is not sealed while submitting a claim, allowing for more relays to enter the evidence claim. It also generates a merkle proof for the claim's total proof count if it doesn't exceed the maximum number of relays per session, otherwise it falls back to the max relays per session. Additionally, it includes validation of the level count on the claim by the total relays.

### DIFF
--- a/x/pocketcore/keeper/proof.go
+++ b/x/pocketcore/keeper/proof.go
@@ -50,13 +50,16 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 				ctx.Logger().Error(fmt.Sprintf("could not delete evidence is not sealed, could cause a relay leak: %s", err.Error()))
 			}
 		}
-		if evidence.NumOfProofs != claim.TotalProofs {
+
+		if evidence.NumOfProofs < claim.TotalProofs {
 			err := pc.DeleteEvidence(claim.SessionHeader, claim.EvidenceType, node.EvidenceStore)
 			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak"))
 			if err != nil {
 				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak: %s", err.Error()))
 			}
+			continue
 		}
+
 		// get the session context
 		sessionCtx, err := ctx.PrevCtx(claim.SessionHeader.SessionBlockHeight)
 		if err != nil {
@@ -73,8 +76,24 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 		if !found {
 			ctx.Logger().Error(fmt.Sprintf("an error occurred creating the proof transaction with app %s not found with evidence %v", evidence.ApplicationPubKey, evidence))
 		}
+
+		// There is a potential race condition where the evidence is not sealed while submitting a claim
+		// Allowing for more relays to enter the evidence claim, and whenever a proof is generated, it will
+		// result in an incorrect merkle proof from being generated.
+		// In order to allow for best-effort process to submit a proof for the claim the servicer submitted
+		// 1. Generate a merkle proof for claim's total proof if proof count doesn't exceed max relays per session
+		// 2. Otherwise, fall back to max relays per session
+		relaysCompleted := claim.TotalProofs
+
+		maxSessionRelays := pc.MaxPossibleRelays(app, k.SessionNodeCount(sessionCtx)).Int64()
+
+		// If the relays completed is more than what protocol will allow, set it to the max.
+		if relaysCompleted > maxSessionRelays {
+			relaysCompleted = maxSessionRelays
+		}
+
 		// get the merkle proof object for the pseudorandom index
-		mProof, leaf := evidence.GenerateMerkleProof(claim.SessionHeader.SessionBlockHeight, int(index), pc.MaxPossibleRelays(app, k.SessionNodeCount(sessionCtx)).Int64())
+		mProof, leaf := evidence.GenerateMerkleProof(claim.SessionHeader.SessionBlockHeight, int(index), relaysCompleted)
 		// if prevalidation on, then pre-validate
 		if pc.GlobalPocketConfig.ProofPrevalidation {
 			// validate level count on claim by total relays
--- a/x/pocketcore/keeper/proof.go
+++ b/x/pocketcore/keeper/proof.go
@@ -51,15 +51,13 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 			}
 		}
 
+		// In the case of evidence.NumOfProofs < claim.TotalProofs, we don't submit a proof because we cannot generate a valid merkle proof with a smaller number of relays.
+		// In the case of evidence.NumOfProofs > claim.TotalProofs, we truncate proofs of the evidence (that happens inside GenerateMerkleProof) and submit a proof.
 		if evidence.NumOfProofs < claim.TotalProofs {
-			//  The node should never submit a proof whenever the number of stored relay proofs are
-			//  LESS than the number of relay proofs in the submitted claim as it can cause
-			//  a array index of out-bounds exception if the randomly selected index is more than what
-			//  we have in the store, causing a process crash.
 			err := pc.DeleteEvidence(claim.SessionHeader, claim.EvidenceType, node.EvidenceStore)
-			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak"))
+			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs is less than submitted claim's total proofs, can't generate merkle proof"))
 			if err != nil {
-				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak: %s", err.Error()))
+				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs is less than submitted claim's total proofs, can't generate merkle proof: %s", err.Error()))
 			}
 			// If this is the case, delete the evidence and continue iterating through the claims to submit.
 			continue
@@ -85,7 +83,7 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 		// There is a potential race condition where the evidence is not sealed while submitting a claim
 		// Allowing for more relays to enter the evidence claim, and whenever a proof is generated, it will
 		// result in an incorrect merkle proof from being generated.
-		// In order to allow for best-effort process to submit a proof for the claim the servicer submitted
+		// In order to allow for best-effort process to submit a proof for the claim the service submitted
 		// 1. Generate a merkle proof for claim's total proof if proof count doesn't exceed max relays per session
 		// 2. Otherwise, fall back to max relays per session
 		relaysCompleted := claim.TotalProofs
--- a/x/pocketcore/keeper/proof.go
+++ b/x/pocketcore/keeper/proof.go
@@ -59,7 +59,7 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 			if err != nil {
 				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs is less than submitted claim's total proofs, can't generate merkle proof: %s", err.Error()))
 			}
-			// If this is the case, delete the evidence and continue iterating through the claims to submit.
+			// // If this is the case, delete the evidence associated with this claim and continue iterating through the claims from other nodes that must submit a proof tx.
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,16 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 				ctx.Logger().Error(fmt.Sprintf("could not delete evidence is not sealed, could cause a relay leak: %s", err.Error()))
 			}
 		}
-		if evidence.NumOfProofs != claim.TotalProofs {
+
+		if evidence.NumOfProofs < claim.TotalProofs {
```
Olshansk, line 56: Can you add inline comments in the code related to

1. Why we check for `<` (your explanation in the PR should suffice)
2. Why we do a `continue` below (explain that reward will be lost)

x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,16 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 				ctx.Logger().Error(fmt.Sprintf("could not delete evidence is not sealed, could cause a relay leak: %s", err.Error()))
 			}
 		}
-		if evidence.NumOfProofs != claim.TotalProofs {
+
```
Olshansk, line 53: @PoktBlade Have you looked into the difficulty of adding a unit test for this? I realize it's not easy, but just wanted to make sure that we at least put effort into it.

x/pocketcore/keeper/proof.go
```diff
@@ -73,8 +76,24 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 		if !found {
 			ctx.Logger().Error(fmt.Sprintf("an error occurred creating the proof transaction with app %s not found with evidence %v", evidence.ApplicationPubKey, evidence))
 		}
+
+		// There is a potential race condition where the evidence is not sealed while submitting a claim
```
Olshansk: Do you have any idea why this wasn't a major concern in the past?

Olshansk (changes requested):
Left a couple comments - no blockers.

Appreciate the explanation in the issue. I understand it looks safe but trying to be "ultra careful" with any last minute changes.

Adding @msmania for a review as well.

x/pocketcore/keeper/proof.go
nodiesBlade: Generally whenever the claim is sealed, relays stop coming in and so there is some non deterministic timing in which the race condition could happen. If relays keep coming in (i.e session rollover), the problem becomes more prominent.

We did find some on chain tx's to show that this did periodically happen even without session rollover though.

x/pocketcore/keeper/proof.go
nodiesBlade: I can look into it some more for sanity purposes.

In regards to some other testing, @jorgecuesta did run it live in his fleet, and saw that there weren't any invalid proofs being submitted on chain. This was a dramatic decrease (0 invalid proofs) vs what he saw before the changes.

x/pocketcore/keeper/proof.go
nodiesBlade: Sounds good, done

x/pocketcore/keeper/proof.go
nodiesBlade: image
You can go [here ](https://www.poktscan.com/node_runner?domain=poktscan.cloud&tab=transfers)and filter the transactions by "Proof" and "Failed", and put the start date before ~ the July 10th and it will show some of the failed submissions even with session rollover disabled.

x/pocketcore/keeper/proof.go
nodiesBlade: Update: I wrote some integration tests last night based off `TestClaimProtoTx`, but it turns out the test for this was already broken.. The reason why we don't see them when running tests, these tests are skipped whenever -s flag is passed.

Anyhow.. will see if I can fix the initial broken tests (seems like a crashing error when submitting proofs, due to AIOB)

x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,21 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 				ctx.Logger().Error(fmt.Sprintf("could not delete evidence is not sealed, could cause a relay leak: %s", err.Error()))
 			}
 		}
-		if evidence.NumOfProofs != claim.TotalProofs {
+
+		if evidence.NumOfProofs < claim.TotalProofs {
+			//  The node should never submit a proof whenever the number of stored relay proofs are
```
msmania: A comment here should mention two cases for readability:

- In the case of `evidence.NumOfProofs < claim.TotalProofs`, we don't submit a proof because we cannot generate a valid merkle proof with a smaller number of relays.
- In the case of `evidence.NumOfProofs > claim.TotalProofs`, we truncate proofs of the evidence (that happens inside `GenerateMerkleProof`) and submit a proof.

I think the problem is not OOB.  If so, the fix would be to pass `evidence.NumOfProofs` to `getPseudorandomIndex`, or add a boundary check somewhere.  The problem is that a smaller evidence is not enough to generate a matching merkle proof.  That's why we don't submit a proof.

x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,21 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
...
+			//  LESS than the number of relay proofs in the submitted claim as it can cause
+			//  a array index of out-bounds exception if the randomly selected index is more than what
+			//  we have in the store, causing a process crash.
 			err := pc.DeleteEvidence(claim.SessionHeader, claim.EvidenceType, node.EvidenceStore)
 			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak"))
```
msmania: Can you update the error message as well?

msmania (changes requested):
The change looks good.  Can you update the comments for better readability?

x/pocketcore/keeper/proof.go
nodiesBlade: Sounds good, done

x/pocketcore/keeper/proof.go
nodiesBlade: > The problem is that a smaller evidence is not enough to generate a matching merkle proof. That's why we don't submit a proof.

Asking for my knowledge - wouldn't you be able to generate a valid proof if the pseudo-selected random index is within bounds? 

Anyhow, updated the comment. THanks for the review!

x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,19 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
...
+		// In the case of evidence.NumOfProofs > claim.TotalProofs, we truncate proofs of the evidence (that happens inside GenerateMerkleProof) and submit a proof.
+		if evidence.NumOfProofs < claim.TotalProofs {
 			err := pc.DeleteEvidence(claim.SessionHeader, claim.EvidenceType, node.EvidenceStore)
-			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak"))
+			ctx.Logger().Error(fmt.Sprintf("evidence num of proofs is less than submitted claim's total proofs, can't generate merkle proof"))
 			if err != nil {
-				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs does not equal claim total proofs... possible relay leak: %s", err.Error()))
+				ctx.Logger().Error(fmt.Sprintf("evidence num of proofs is less than submitted claim's total proofs, can't generate merkle proof: %s", err.Error()))
 			}
+			// If this is the case, delete the evidence and continue iterating through the claims to submit.
```
Olshansk:
```suggestion
			// If this is the case, delete the evidence associated with this claim and continue iterating through the  claims from other nodes that must submit a proof tx.
```

x/pocketcore/keeper/proof.go
Olshansk: @PoktBlade Bump on existing tests and/or new tests.

x/pocketcore/keeper/proof.go
Olshansk: Was running into some issues but reached out to poktscan here: https://discord.com/channels/854406364931686400/950799595171086377/1131367539667107872

x/pocketcore/keeper/proof.go
```diff
@@ -73,8 +79,24 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 		if !found {
 			ctx.Logger().Error(fmt.Sprintf("an error occurred creating the proof transaction with app %s not found with evidence %v", evidence.ApplicationPubKey, evidence))
 		}
+
+		// There is a potential race condition where the evidence is not sealed while submitting a claim
+		// Allowing for more relays to enter the evidence claim, and whenever a proof is generated, it will
+		// result in an incorrect merkle proof from being generated.
```
Olshansk:
```suggestion
		// Allowing for more relays than the maximum allowed in a certain session will result
		// in the generation of an incorrect Merkle proof.
```

x/pocketcore/keeper/proof.go
```diff
@@ -73,8 +79,24 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
...
+		// In order to allow for best-effort process to submit a proof for the claim the service submitted
```
Olshansk:
```suggestion
		// In order to allow for best-effort submission of proofs for each existing claim:
```

x/pocketcore/keeper/proof.go
```diff
@@ -50,13 +50,19 @@ func (k Keeper) SendProofTx(ctx sdk.Ctx, n client.Client, node *pc.PocketNode, p
 				ctx.Logger().Error(fmt.Sprintf("could not delete evidence is not sealed, could cause a relay leak: %s", err.Error()))
 			}
 		}
-		if evidence.NumOfProofs != claim.TotalProofs {
+
+		// In the case of evidence.NumOfProofs < claim.TotalProofs, we don't submit a proof because we cannot generate a valid merkle proof with a smaller number of relays.
```
Olshansk, line 54: Can you add examples in the comment on when > and < happens?

Olshansk (changes requested):

x/pocketcore/keeper/proof.go
nodiesBlade: That's what this PR is for.

Git blame, reference back here. I rather not include such explicitness in the code for a guardrail.

nodiesBlade: added a reference back to this PR instead

nodiesBlade:
Added all the comments as suggested! Thanks for the review.

In regards to the tests, I haven't made any progress to them - don't think I will have the bandwidth moving forward to diagnose / fix the integration tests. The changes here are minimal and only guard rails for proof tx, not consensus. If needed, we can push an additional hot fix without the need for consensus upgrades.

Olshansk (changes requested):
Going to block this PR until tests are fixed & added per this comment: https://github.com/pokt-network/pocket-core/pull/1574#issuecomment-1644197144

@PoktBlade Could you do a handoff (whether it's to myself, a core team member or someone from the community) with regard to the status of the tests you investigated? For example:
- Which tests exist for it (e.g. names & links)
- Which ones are broken (e.g. names & links)
- Insight/ideas into why certain tests are broken (e.g. a few bullet points)
- Which ones would you add if you had capacity (e.g. a list)

nodiesBlade:
I'm handing it off as is. The PR is already merged into Poktscan's fork.

nodiesBlade:
Since I don't have time to dedicate more to this issue, please cherry pick or fork my changes and move forward. 

Closing since it won't be merged in / blocked.

jorgecuesta:
@Olshansk all the request you ask about test are not fair to be handle by the community people trying to help if pocket has people that get a salary for this. 
This is why then people hesitate to help on the project. I think u guys may need to figure out a way to work on v0 because has a lot of things that are not working and should be working like TESTs and that is pocket responsibility, not the community indeed.

Olshansk:
> @Olshansk all the request you ask about test are not fair to be handle by the community people trying to help if pocket has people that get a salary for this. This is why then people hesitate to help on the project. I think u guys may need to figure out a way to work on v0 because has a lot of things that are not working and should be working like TESTs and that is pocket responsibility, not the community indeed.

Thanks for the feedback.

I will work on adding the tests myself this time and reach out about getting a budget for it in the future.

PR #1581, opened 2023-09-07
[v0.11 feature] Built-in Reward Share

msmania:
## Description

This patch implements the Built-in Reward Share feature which lets the network itself to distribute the relay/block rewards into multiple addresses.  The change consists of the following parts.

- Add `RewardDelegators` in `Validator`
- Add `RewardDelegators` in `MsgStake`
- Add a new command `pocket nodes stakeNew`
- Distributes the fee of claim/proof transaction to a node's operator address
- Add a new config `prevent_negative_reward_claim` not to claim a potential loss evidence.

This change is consensus-breaking.  The new behavior is put behind a new feature key `RewardDelegator`.  The new field is kept unavailable until activation.

The new structure of `Validator` or `MsgStake` is backward/forward compatible, meaning the new binary can still unmarshal data marshaled by an older binary, and vice versa.  In other words, the network before `RewardDelegator` activation accepts an `MsgStake` transaction, ignoring the `RewardDelegators` field.  And the new `Validator` structure can handle all historical states from genesis.  Therefore this patch does not introduce a structure like `10.0Validaor` as the NCUST patch did before.

### Summary generated by Reviewpad on 26 Dec 23 01:21 UTC
This pull request includes changes to multiple files. Here is a summary of the diff:

1. The file `common_test.go` was modified to replace the import of `math/rand` with `crypto/rand`. Additionally, the comment `// : deadcode unused` was removed.

2. The file `x/nodes/keeper/abci_test.go` was modified to add and remove import statements, as well as comment out unnecessary code related to state conversion.

3. The file `x/nodes/types/validator.go` was modified to add an import, add a new field to the `Validator` struct, add a new function to create a validator from a message, modify several methods to include a new field in the output, and add a new struct and comment.

4. The file `x/nodes/types/validator_test.go` was modified to add import statements and a new test function.

5. The file `msg_test.go` was modified to add and remove import statements, add new test functions, and update existing test functions.

6. The file `keeper_test.go` was modified to add import statements, modify existing test functions, and add new test functions.

7. The file `go.mod` was modified to add and update package requirements.

8. The file `handler.go` was modified to add import statements and modify function implementations.

9. The file `nodes.proto` was modified to remove an import statement and add a new field to a message.

10. The file `msg.go` was modified to add import statements, add a new struct and function, and modify existing methods.

11. The file `genesis_test.go` was modified to add import statements and modify existing test functions.

12. The file `rpc_test.go` was modified to add and remove import statements, modify function implementations, and add test cases.

13. The file `expectedKeepers.go` was modified to remove comments and add a new method.

14. The file `config.go` was modified to add a new field to a struct.

15. The file `msg.proto` was modified to add a new field to a message.

16. The file `LegacyValidator.go` was modified to add a new method and update existing methods.

17. The file `errors.go` was modified to add new error codes and functions to handle them.

18. The file `reward_test.go` was modified to add import statements, add and update test functions.

19. The file `util_test.go` was modified to rearrange import statements and add new test functions.

Please review these changes and provide any necessary feedback. Let me know if you need more information or if there's anything else I can assist you with.


POKT-Discourse:

https://forum.pokt.network/t/unleashing-the-potential-of-pocket/4720/1

POKT-Discourse:

https://forum.pokt.network/t/unleashing-the-potential-of-non-custodial-node-running/4796/1

Olshansk:
@msmania Please note that I updated the base to `tokikuch/RTTM2` so its easier to review. We HAVE TO make sure to keep it up to date with that branch and MAKE SURE to update it back to `main` before meging.

app/cmd/cli/node.go
```diff
@@ -116,3 +117,73 @@ Will prompt the user for the <fromAddr> account passphrase.`,
 		fmt.Println(resp)
 	},
 }
+
+var stakeNewCmd = &cobra.Command{
+	Use:   `stakeNew <OperatorPublicKey> <OutputAddress> <SignerAddress> <Stake> <ChainIDs> <ServiceURL> <Delegators> <NetworkID> <Fee> [Memo]`,
```
Olshansk:
```suggestion
	Use:  "stakeNew <OperatorPublicKey> <OutputAddress> <SignerAddress> <Stake> <ChainIDs> <ServiceURL> <Delegators> <NetworkID> <Fee> [Memo]",
```

app/cmd/cli/node.go
```diff
@@ -116,3 +117,73 @@ Will prompt the user for the <fromAddr> account passphrase.`,
...
+  OperatorPublicKey Public key to use as the node's operator account
+  OutputAddress     Address to use as the node's output account
+  SignerAddress     Address to sign the transaction
+  Stake             Amount to stake in uPOKT
+  ChainIDs          Comma-separated chain IDs to host on the node
+  ServiceURL        Relay endpoint of the node.  Must include the port number.
+  Delegators        Delegator addresses to share rewards
+  NetworkID         Network ID to submit a transaction to e.g. mainnet or testnet
+  Fee               Transaction fee in uPOKT
+  Memo              Text to include in the transaction.  No functional effect.
```
Olshansk: Can you add an example command at the bottom of the `Long` description, just so everyone can see what it looks like?

app/cmd/cli/node.go
```diff
@@ -116,3 +117,73 @@ Will prompt the user for the <fromAddr> account passphrase.`,
 		fmt.Println(resp)
 	},
 }
+
+var stakeNewCmd = &cobra.Command{
```
Olshansk, line 123:
```suggestion
// stakeNew is an upgraded version of `nodesCmd` that captures newer on-chain functionality in a cleaner way
var stakeNewCmd = &cobra.Command{
```

app/cmd/cli/node.go
Olshansk:
```suggestion
  Memo              Optional. Text to include in the transaction. No functional effect.
```

app/cmd/rpc/rpc_test.go
```diff
@@ -1636,3 +1634,53 @@ func NewValidChallengeProof(t *testing.T, privateKeys []crypto.PrivateKey) (chal
 	}
 	return proof
 }
+
+func TestMsgStake_Marshaling_BackwardCompatibility(t *testing.T) {
+	// Tx 3640B15041998FE800C2F61FC033CBF295D9282B5E7045A16F754ED9D8A54AFF in Mainnet
```
Olshansk: Please add a comment w/ the CLI command you used to retrieve the hex encoded tx below.

app/cmd/rpc/rpc_test.go
```diff
@@ -1636,3 +1634,53 @@ func NewValidChallengeProof(t *testing.T, privateKeys []crypto.PrivateKey) (chal
...
+			"QTKyYiRdhqrGqeqTIMVSckAe8RIEMDAwMxIEMDAwNBIEMDAwNRIEMDAwORIEMDAy" +
+			"MRIEMDAyNxIEMDAyOBIEMDA0NhIEMDA0NxIEMDA0ORIEMDA1MBIEMDA1NhIEMDA2" +
+			"NhIEMDA3MhIEMDNERhoMMTQwMDAwMDAwMDAwIiNodHRwczovL3ZhbDE2NjcwMDUy" +
+			"MDYuYzBkM3Iub3JnOjQ0MyoU6By0i9H9b2jibqTioCbqBdSFO3USDgoFdXBva3QS" +
+			"BTEwMDAwGmQKIHN80LkGpRfoCjD3vrxBMrJiJF2Gqsap6pMgxVJyQB7xEkDOrzwH" +
+			"w68+vl2z9nC+zYz3u4J7Oe3ntBOVP+cYHO5+lLuc8nH0OaG6pujXEPo19F5qW4Zh" +
+			"NBEgtChJp+QhYVgIIiBDdXN0b2RpYWwgdG8gTm9uLUN1c3RvZGlhbCBhZ2FpbijS" +
+			"CQ=="
+
+	StakeTxAfterDelegatorsUpgrade :=
```
Olshansk, line 1706: Please add a comment on how you generated this.

types/config.go
```diff
@@ -19,37 +19,38 @@ type SDKConfig struct {
...
+	ABCILogging                     bool   `json:"abci_logging"`
+	RelayErrors                     bool   `json:"show_relay_errors"`
+	DisableTxEvents                 bool   `json:"disable_tx_events"`
+	Cache                           bool   `json:"-"`
+	IavlCacheSize                   int64  `json:"iavl_cache_size"`
+	ChainsHotReload                 bool   `json:"chains_hot_reload"`
+	GenerateTokenOnStart            bool   `json:"generate_token_on_start"`
+	LeanPocket                      bool   `json:"lean_pocket"`
+	LeanPocketUserKeyFileName       string `json:"lean_pocket_user_key_file"`
+	NotClaimPossiblyNegativeRewards bool   `json:"not_claim_possibly_negative_rewards"`
```
Olshansk: Wdyt of `prevent_negative_reward_claim` instead of `not_claim_possibly_negative_rewards`?

If we go with this, need to update all corresponding types, functions, vars, comments, etc...

app/cmd/rpc/rpc_test.go
```diff
@@ -1636,3 +1634,53 @@ func NewValidChallengeProof(t *testing.T, privateKeys []crypto.PrivateKey) (chal
...
+	})
+
+	// Choose Proto marshaler
+	heightForProto := int64(-1)
+	// Simulate post-NCUST
+	codec.UpgradeFeatureMap[codec.NonCustodialUpdateKey] = -1
+	// Initialize app.cdc
+	app.Codec()
+
+	stdTx, err := app.UnmarshalTxStr(StakeTxBeforeDelegatorsUpgrade, heightForProto)
```
Olshansk, line 1729:
```suggestion
	// Validate that an old stake messages DOES NOT have delegators
	stdTx, err := app.UnmarshalTxStr(StakeTxBeforeDelegatorsUpgrade, heightForProto)
```

app/cmd/rpc/rpc_test.go
```diff
@@ -1636,3 +1634,53 @@ func NewValidChallengeProof(t *testing.T, privateKeys []crypto.PrivateKey) (chal
...
+	assert.Nil(t, err)
+	msgStake, ok := stdTx.Msg.(*types2.MsgStake)
+	assert.True(t, ok)
+	assert.Nil(t, msgStake.Delegators)
+	assert.Nil(t, msgStake.ValidateBasic())
+
+	stdTx, err = app.UnmarshalTxStr(StakeTxAfterDelegatorsUpgrade, heightForProto)
```
Olshansk, line 1737:
```suggestion
	// Validate that an old stake messages DOES have delegators
	stdTx, err = app.UnmarshalTxStr(StakeTxAfterDelegatorsUpgrade, heightForProto)
```

proto/x/nodes/msg.proto
```diff
@@ -23,6 +23,10 @@ message MsgProtoStake {
 		(gogoproto.jsontag) = "output_address,omitempty",
 		(gogoproto.moretags) = "yaml:\"output_address\""
 	];
+	map<string, uint32> Delegators = 6 [
```
Olshansk: Please add a comment on what this is. For example: `// Mapping from delegated-to addresses to a fraction (value between 0 and 100) of rewards`.

proto/x/nodes/nodes.proto
```diff
@@ -12,6 +11,23 @@ message ProtoValidator {
...
+	int32 status = 4 [(gogoproto.jsontag) = "status"];
+	repeated string Chains = 5 [(gogoproto.jsontag) = "chains"];
+	string ServiceURL = 6 [(gogoproto.jsontag) = "service_url"];
+	string StakedTokens = 7 [(gogoproto.customtype) = "github.com/pokt-network/pocket-core/types.BigInt", (gogoproto.jsontag) = "tokens", (gogoproto.nullable) = false];
+	google.protobuf.Timestamp UnstakingCompletionTime = 8 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true, (gogoproto.jsontag) = "unstaking_time", (gogoproto.moretags) = "yaml:\"unstaking_time\""];
+	bytes OutputAddress = 9 [(gogoproto.casttype) = "github.com/pokt-network/pocket-core/types.Address", (gogoproto.jsontag) = "output_address,omitempty", (gogoproto.moretags) = "yaml:\"output_address\""];
+	map<string, uint32> Delegators = 10 [(gogoproto.jsontag) = "delegators,omitempty", (gogoproto.moretags) = "yaml:\"delegators\""];
+}
+
+message ProtoValidatorV8 {
```
Olshansk: 1. Why did we create a new proto instead of updating the old one?
2. Why is it called V8?

x/nodes/handler.go
```diff
@@ -59,10 +60,14 @@ func handleStake(ctx sdk.Ctx, msg types.MsgStake, k keeper.Keeper, signer crypto
 		}
 	}
 
-	pk := msg.PublicKey
-	addr := pk.Address()
-	// create validator object using the message fields
-	validator := types.NewValidator(sdk.Address(addr), pk, msg.Chains, msg.ServiceUrl, sdk.ZeroInt(), msg.Output)
+	if !k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) &&
+		msg.Delegators != nil {
+		msg.Delegators = nil
```
Olshansk: Do you not think this should be an error or a panic?

I feel like nullifying it could create some indirection for the user/developer, or potentially some more complex scenario.

x/nodes/handler.go
```diff
@@ -59,10 +60,14 @@ func handleStake(ctx sdk.Ctx, msg types.MsgStake, k keeper.Keeper, signer crypto
...
+	}
+
+	validator := types.NewValidatorFromMsg(msg)
+	// StakedTokens is set through StakeValidator.  Resetting to 0 for now.
```
Olshansk: Not completely clear to me why we're doing this.

Olshansk (changes requested):
Uploading a partial review (not complete yet)

Screenshot 2023-12-13 at 4 41 33 PM

codec/codec.go
```diff
@@ -58,6 +58,7 @@ const (
 	OutputAddressEditKey         = "OEDIT"
 	ClearUnjailedValSessionKey   = "CRVAL"
 	PerChainRTTM                 = "PerChainRTTM"
+	DelegatorsKey                = "DELE"
```
Olshansk: NIT: We've already moved away from the 5 char key with `PerChainRTTM` so I suggest renaming this as well.

`RewardsDelegatorsKey = "RewardDelegator"`

- `DELE` is too much like `delete`
- `Delegator` has lots of meanings in the blockchain world so I'd rather we be verbose

x/nodes/keeper/params.go
```diff
@@ -160,6 +164,25 @@ func (k Keeper) NodeReward(ctx sdk.Ctx, reward sdk.BigInt) (nodeReward sdk.BigIn
 	return
 }
 
+// Split feeCollector's cut into DAO and Propower
```
Olshansk:
```suggestion
// Split feeCollector's cut into the DAO's cut and the Proposer's cut
```

x/nodes/keeper/params.go
```diff
@@ -144,7 +144,11 @@ func (k Keeper) ServicerStakeFloorMultiplierExponent(ctx sdk.Ctx) (res sdk.BigDe
 	return
 }
 
-func (k Keeper) NodeReward(ctx sdk.Ctx, reward sdk.BigInt) (nodeReward sdk.BigInt, feesCollected sdk.BigInt) {
+// Split rewards into node's cut and feeCollector's cut (= DAO + Proposer)
```
Olshansk:
```suggestion
// Split block rewards into the node's cut and the feeCollector's (DAO + Proposer) cut
```

x/nodes/keeper/params.go
```diff
@@ -160,6 +164,25 @@ func (k Keeper) NodeReward(ctx sdk.Ctx, reward sdk.BigInt) (nodeReward sdk.BigIn
...
+	daoAllocation := sdk.NewDec(k.DAOAllocation(ctx))
+	proposerAllocation := sdk.NewDec(k.ProposerAllocation(ctx))
+
+	// get the new percentages of `dao / (dao + proposer)`
+	daoAllocation = daoAllocation.Quo(daoAllocation.Add(proposerAllocation))
+
+	// dao cut calculation truncates int ex: 1.99uPOKT = 1uPOKT
+	daoCut = feesCollected.ToDec().Mul(daoAllocation).TruncateInt()
+
+	// proposer is whatever is left
```
Olshansk:
```suggestion
	// proposer gets whatever is left after the DAO's truncated rewards are taken out
```

x/nodes/keeper/reward_test.go
```diff
@@ -87,6 +90,28 @@ func TestMint(t *testing.T) {
...
+	if !acc.Coins.IsEqual(expectedCoins) {
+		fmt.Println(
+			"Balance mismatch",
+			address,
+			"actual=",
+			acc.Coins,
+			"expected=",
+			expectedCoins,
+		)
+	}
```
Olshansk: 1. If we get an error, I'm not 100% sure if this goes to the next line.
2. Regardless of (1), I think this (pseudocode below) might be more idiomatic:

```go
errorStr = fmt.Sprintf("Balance mistmatch ...", arg1, arg2, ...)
assert.True(t, acc.Coins.IsEqual(expectedCoins), errorStr)
```

x/nodes/types/msg.go
```diff
@@ -141,16 +143,17 @@ func (msg MsgSend) GetFee() sdk.BigInt {
...
-	Chains     []string         `json:"chains" yaml:"chains"`
-	Value      sdk.BigInt       `json:"value" yaml:"value"`
-	ServiceUrl string           `json:"service_url" yaml:"service_url"`
-	Output     sdk.Address      `json:"output_address,omitempty" yaml:"output_address"`
+	PublicKey  crypto.PublicKey  `json:"public_key" yaml:"public_key"`
+	Chains     []string          `json:"chains" yaml:"chains"`
+	Value      sdk.BigInt        `json:"value" yaml:"value"`
+	ServiceUrl string            `json:"service_url" yaml:"service_url"`
+	Output     sdk.Address       `json:"output_address,omitempty" yaml:"output_address"`
+	Delegators map[string]uint32 `json:"delegators,omitempty" yaml:"delegators"`
```
Olshansk: 1. Please add a comment above explaining what kind of delegators.
2. Optional: consider renaming to `RewardDelegators`

x/nodes/types/msg.go
```diff
@@ -229,6 +233,18 @@ func (msg MsgStake) ValidateBasic() sdk.Error {
 	if err := ValidateServiceURL(msg.ServiceUrl); err != nil {
 		return err
 	}
+	if msg.Delegators != nil {
+		totalShares := uint32(0)
+		for addrStr, share := range msg.Delegators {
```
Olshansk: s/share/rewardShare

x/nodes/types/msg.go
```diff
@@ -229,6 +233,18 @@ func (msg MsgStake) ValidateBasic() sdk.Error {
...
+			if _, err := sdk.AddressFromHex(addrStr); err != nil {
+				return ErrInvalidDelegators(DefaultCodespace, err.Error())
+			}
+			totalShares = totalShares + share
+			if totalShares > 100 {
+				return ErrInvalidDelegators(DefaultCodespace, "Total share exceeds 100")
```
Olshansk:
```suggestion
				return ErrInvalidDelegators(DefaultCodespace, fmt.Sprintf("Total share %d exceeds 100", totalShares))
```

x/nodes/types/validator_legacy.go
```diff
@@ -2,13 +2,16 @@ package types
...
 	"fmt"
+	"time"
+
 	"github.com/pokt-network/pocket-core/codec"
 	"github.com/pokt-network/pocket-core/crypto"
 	sdk "github.com/pokt-network/pocket-core/types"
-	"time"
 )
 
+// Make sure these structs implement ProtoMarshaler
```
Olshansk:
```suggestion
// Compilation level enforcement to validate that structs implement ProtoMarshaler
```

x/nodes/types/validator_legacy.go
```diff
@@ -21,6 +24,17 @@ type LegacyValidator struct {
 	UnstakingCompletionTime time.Time        `json:"unstaking_time" yaml:"unstaking_time"` // if unstaking, min time for the validator to complete unstaking
 }
 
+func (v LegacyValidator) ExactEqualsTo(v2 LegacyValidator) bool {
```
Olshansk: 1. s/Exact/Deep?
2. Can you add a comment on why we're implementing a custom `Equals` operation?

x/nodes/types/validator_legacy.go
```diff
@@ -176,3 +218,137 @@ func (v LegacyValidator) ToProto() LegacyProtoValidator {
 		UnstakingCompletionTime: v.UnstakingCompletionTime,
 	}
 }
+
+type LegacyValidator8 struct {
```
Olshansk: What does the 8 represent?

x/nodes/types/validator.go
```diff
@@ -4,27 +4,27 @@ import (
...
+	Jailed                  bool              `json:"jailed" yaml:"jailed"`                           // has the validator been jailed from staked status?
+	Status                  sdk.StakeStatus   `json:"status" yaml:"status"`                           // validator status (staked/unstaking/unstaked)
+	Chains                  []string          `json:"chains" yaml:"chains"`                           // validator non native blockchains
+	ServiceURL              string            `json:"service_url" yaml:"service_url"`                 // url where the pocket service api is hosted
+	StakedTokens            sdk.BigInt        `json:"tokens" yaml:"tokens"`                           // tokens staked in the network
+	UnstakingCompletionTime time.Time         `json:"unstaking_time" yaml:"unstaking_time"`           // if unstaking, min time for the validator to complete unstaking
+	OutputAddress           sdk.Address       `json:"output_address,omitempty" yaml:"output_address"` // the custodial output address of the validator
+	Delegators              map[string]uint32 `json:"delegators,omitempty" yaml:"delegators"`
```
Olshansk: 1. Add comment on what kind of delegators these are
2. Optional: consider renaming per some of the previous commits.

x/nodes/types/validator.go
```diff
@@ -248,19 +292,21 @@ func (v Validator) ToProto() ProtoValidator {
...
+	PublicKey               string            `json:"public_key" yaml:"public_key"`         // the consensus public key of the validator; hex encoded in JSON
+	Jailed                  bool              `json:"jailed" yaml:"jailed"`                 // has the validator been jailed from staked status?
+	Status                  sdk.StakeStatus   `json:"status" yaml:"status"`                 // validator status (staked/unstaking/unstaked)
+	Chains                  []string          `json:"chains" yaml:"chains"`                 // validator non native blockchains
+	ServiceURL              string            `json:"service_url" yaml:"service_url"`       // url where the pocket service api is hosted
+	StakedTokens            sdk.BigInt        `json:"tokens" yaml:"tokens"`                 // tokens staked in the network
+	UnstakingCompletionTime time.Time         `json:"unstaking_time" yaml:"unstaking_time"` // if unstaking, min time for the validator to complete unstaking
+	OutputAddress           sdk.Address       `json:"output_address" yaml:"output_address"` // custodial output address of tokens
+	Delegators              map[string]uint32 `json:"delegators" yaml:"delegators"`
```
Olshansk: Ditto

x/nodes/types/util_test.go
```diff
@@ -24,3 +25,26 @@ func TestValidateServiceURL(t *testing.T) {
 	assert.NotNil(t, ValidateServiceURL(invalidURLBadPort), "invalid bad port")
 	assert.NotNil(t, ValidateServiceURL(invalidURLBad), "invalid bad url")
 }
+
+func TestCompareStringMaps(t *testing.T) {
```
Olshansk, line 41: Should we add a unit test for `CompareSlices` too?

x/nodes/types/util_test.go
```diff
@@ -24,3 +25,26 @@ func TestValidateServiceURL(t *testing.T) {
...
+
+	m2 = nil
+	assert.False(t, CompareStringMaps(m1, m2))
+	assert.False(t, CompareStringMaps(nil, m1))
+
+	m1 = nil
+	assert.True(t, CompareStringMaps(m1, m2))
+
+	assert.True(t, CompareStringMaps(nil, map[string]int{}))
+}
```
Olshansk:
```suggestion

	// m1 is non-empty and m2 is empty
	m1["a"] = 10
	m1["b"] = 100
	assert.False(t, CompareStringMaps(m1, m2))

	// m1 and m2 are not empty and identical
	m2["b"] = 100
	m2["a"] = 10
	assert.True(t, CompareStringMaps(m2, m1))

	// m1 is non-empty and m2 is nil
	m2 = nil
	assert.False(t, CompareStringMaps(m1, m2))
	assert.False(t, CompareStringMaps(nil, m1))

	// m1 and m2 are both nil
	m1 = nil
	assert.True(t, CompareStringMaps(m1, m2))

	// Empty and nil maps are identical
	assert.True(t, CompareStringMaps(nil, map[string]int{}))
}
```

x/nodes/keeper/reward.go
```diff
@@ -7,13 +7,53 @@ import (
...
+	relays sdk.BigInt,
+	stake sdk.BigInt,
+) (sdk.BigInt, sdk.BigInt) {
+	// feature flags
+	isAfterRSCAL := k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey)
+	multiplier := k.GetChainSpecificMultiplier(ctx, chain)
+
+	var coins sdk.BigInt
+	//check if PIP22 is enabled, if so scale the rewards
+	if isAfterRSCAL {
```
Olshansk, line 69: 1. NIT for readability:

```go
if !isAfterRSCAL {
	coins := multiplier.Mul(relays)
         return k.splitRewards(ctx, coins)
}

// normal business logic
...
```

2. I compare this to the previous code and there were no changes, right?

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
-		//Convert from tokens to a BIN number
-		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
-		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
-		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), Pip22ExponentDenominator).Quo(k.ServicerStakeWeightMultiplier(ctx))
-		coinsDecimal := multiplier.ToDec().Mul(relays.ToDec()).Mul(weight)
-		//truncate back to int
-		coins = coinsDecimal.TruncateInt()
-	} else {
-		coins = multiplier.Mul(relays)
+	if k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) {
```
Olshansk:
```suggestion
	// Check if block rewards need to be split / shared across all delegators
	if k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) {
```

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+		rewardCost := k.AccountKeeper.GetFee(ctx, pcTypes.MsgClaim{}).
+			Add(k.AccountKeeper.GetFee(ctx, pcTypes.MsgProof{}))
+		if toNode.LT(rewardCost) {
+			rewardCost = toNode
+		}
+		k.mint(ctx, rewardCost, validator.Address)
```
Olshansk: Why did you remove the `IsPositive` check? I think we should return it.

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
-		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
-		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
-		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), Pip22ExponentDenominator).Quo(k.ServicerStakeWeightMultiplier(ctx))
-		coinsDecimal := multiplier.ToDec().Mul(relays.ToDec()).Mul(weight)
-		//truncate back to int
-		coins = coinsDecimal.TruncateInt()
-	} else {
-		coins = multiplier.Mul(relays)
+	if k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) {
+		rewardCost := k.AccountKeeper.GetFee(ctx, pcTypes.MsgClaim{}).
```
Olshansk: 1. I think we should move `rewardCost` into a helper (you use it in multiple places)
2. This is new and I don't fully understand the reasoning / goal behind it.

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+			Add(k.AccountKeeper.GetFee(ctx, pcTypes.MsgProof{}))
+		if toNode.LT(rewardCost) {
```
Olshansk, line 131: Should we not just return or default to zero at this positive and skip the mint altogether?

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
 	}
 	return toNode
 }
 
+// Splits rewards into the primary recipient and delegator addresses and
+// invokes a callback per share.
+func SplitNodeRewards(
+	rewards sdk.BigInt,
+	primaryRecipient sdk.Address,
+	delegators map[string]uint32,
```
Olshansk, line 176:
```suggestion
	delegators map[string]uint32, // address to share (< 100) mapp
```

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+	callback func(sdk.Address, sdk.BigInt),
```
Olshansk:
```suggestion
	shareRewardsCallback func(addr sdk.Address, coins sdk.BigInt), // callback to send coins of total rewards to addr
```

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+) {
+	if !rewards.IsPositive() {
+		return
+	}
+
+	totalShare := int64(0)
```
Olshansk: OPTIONAL NIT: move this to a helper called something like `validateRewardShareMap`

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+	for _, share := range delegators {
+		totalShare = totalShare + int64(share)
+		if totalShare > 100 {
+			// If the total shares for delegators exceeds 100,
+			// all rewards go to the primary recipient.
+			delegators = nil
+			break
```
Olshansk: I think this shold be an error/panic, not a break. 

This is on-chain code and should never happen.

x/nodes/keeper/reward.go
```diff
@@ -45,41 +89,84 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
+		percentage := sdk.NewDecWithPrec(int64(share), 2)
+		allocation := rewards.ToDec().Mul(percentage).TruncateInt()
+		if allocation.IsPositive() {
+			callback(addr, allocation)
+		}
+		remains = remains.Sub(allocation)
+	}
+
+	if remains.IsPositive() {
+		callback(primaryRecipient, remains)
```
Olshansk: Please add a log statement or emit an event if this is false.

x/nodes/keeper/reward.go
```diff
@@ -97,38 +184,62 @@ func (k Keeper) GetChainSpecificMultiplier(ctx sdk.Ctx, chain string) sdk.BigInt
...
+		if !k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) {
+			validator.Delegators = nil
 		}
+
+		SplitNodeRewards(
+			proposerCut,
+			k.GetOutputAddressFromValidator(validator),
+			validator.Delegators,
+			func(recipient sdk.Address, share sdk.BigInt) {
+				err = k.AccountKeeper.SendCoins(
```
Olshansk, line 267: You know how you have a function called `rewardFee`. 

I feel like there could be an edge case where (for example) we delegate to 1000 different addresses and run out of funds by running `Send` to everyone.

Wdyt of adding a function (in pseudocode terms) that does `numDelegators * costPerSendTx` and make sure we have enough?

x/pocketcore/keeper/claim.go
```diff
@@ -18,6 +18,10 @@ import (
 func (k Keeper) SendClaimTx(ctx sdk.Ctx, keeper Keeper, n client.Client, node *pc.PocketNode, claimTx func(pk crypto.PrivateKey, cliCtx util.CLIContext, txBuilder auth.TxBuilder, header pc.SessionHeader, totalProofs int64, root pc.HashRange, evidenceType pc.EvidenceType) (*sdk.TxResponse, error)) {
 	// get the private val key (main) account from the keybase
 	address := node.GetAddress()
+	validator := k.posKeeper.Validator(ctx, address)
+	// The cost to earn relay rewards from an evidence
+	rewardCost := k.authKeeper.GetFee(ctx, pc.MsgClaim{}).
```
Olshansk: per some of my previous comments:

1. Consider creating a shared helper
2. Documenting why this is the reward cost
3. Open question about accounting for `delegatorCount * sendTxFee`?

x/nodes/keeper/reward.go
```diff
@@ -7,13 +7,53 @@ import (
...
 
+// CalculateRelayReward - Calculate the amount of rewards based on the given
+// number of relays and the staked tokens.  The returned reward amount includes
+// DAO & Proposer portions.
+func (k Keeper) CalculateRelayReward(
+	ctx sdk.Ctx,
+	chain string,
+	relays sdk.BigInt,
+	stake sdk.BigInt,
+) (sdk.BigInt, sdk.BigInt) {
```
Olshansk: Can you named the return params so it's clear if it is `(daoReward, proposerReward)` or vice versa?

x/pocketcore/keeper/claim.go
```diff
@@ -45,6 +49,39 @@ func (k Keeper) SendClaimTx(ctx sdk.Ctx, keeper Keeper, n client.Client, node *p
 			}
 			continue
 		}
+		if validator != nil && pc.GlobalPocketConfig.NotClaimPossiblyNegativeRewards {
+			rewardExpected, _ := k.posKeeper.CalculateRelayReward(
```
Olshansk, line 58: The godoc for this function is:
```
// CalculateRelayReward - Calculate the amount of rewards based on the given
// number of relays and the staked tokens.  The returned reward amount includes
// DAO & Proposer portions.
```

Thoughts on renaming `rewardExpected` to a more semantic representation?

x/pocketcore/keeper/claim.go
```diff
@@ -45,6 +49,39 @@ func (k Keeper) SendClaimTx(ctx sdk.Ctx, keeper Keeper, n client.Client, node *p
...
+				ctx,
+				evidence.Chain,
+				sdk.NewInt(evidence.NumOfProofs),
+				validator.GetTokens(),
+			)
+			if rewardExpected.LTE(rewardCost) {
+				// If the expected amount of relay rewards from this evidence is less
+				// than the cost of claiming/proofing the evicence, claining the
```
Olshansk:
```suggestion
				// than the cost of claiming/proofing the evidence, claiming the
```

x/pocketcore/keeper/claim.go
```diff
@@ -45,6 +49,39 @@ func (k Keeper) SendClaimTx(ctx sdk.Ctx, keeper Keeper, n client.Client, node *p
...
+				// evidece is a potential loss.
```
Olshansk:
```suggestion
				// evidence is a potential loss.
```

x/pocketcore/keeper/claim.go
```diff
@@ -45,6 +49,39 @@ func (k Keeper) SendClaimTx(ctx sdk.Ctx, keeper Keeper, n client.Client, node *p
...
+				// calculated when the network processes a proof transaction. It's
+				// possible this evidence is profitable if RTTM is increased and/or
+				// the node's stake is increased to an upper bin.
+				ctx.Logger().Info("Discarding an evidence not worth claiming",
+					"addr", address,
+					"sbh", evidence.SessionBlockHeight,
+					"chain", evidence.Chain,
+					"proofs", evidence.NumOfProofs,
+					"rewardExpected", rewardExpected,
+					"cost", rewardCost)
```
Olshansk:
```suggestion
					"rewardCost", rewardCost)
```

x/nodes/types/msg_test.go
```diff
@@ -663,6 +664,47 @@ func TestMsgStake_ValidateBasic(t *testing.T) {
...
+	}
+	assert.Nil(t, msg.ValidateBasic())
+
+	invalidAddr := "1234"
+	msg.Delegators[invalidAddr] = 10
+	err := msg.ValidateBasic()
+	assert.NotNil(t, err)
+	assert.Equal(t, CodeInvalidaDelegators, err.Code())
+
+	// [0]
```
Olshansk:
```suggestion
	// rewardShareMap: {delegator1: 0}
```

x/nodes/types/msg_test.go
```diff
@@ -663,6 +664,47 @@ func TestMsgStake_ValidateBasic(t *testing.T) {
...
+	delete(msg.Delegators, invalidAddr)
+	msg.Delegators[delegator1.PublicKey().Address().String()] = 0
+	assert.Nil(t, msg.ValidateBasic())
+
+	// [100]
```
Olshansk:
```suggestion
	// rewardShareMap: {delegator1: 100}
```

x/nodes/types/msg_test.go
```diff
@@ -663,6 +664,47 @@ func TestMsgStake_ValidateBasic(t *testing.T) {
...
+	msg.Delegators[delegator1.PublicKey().Address().String()] = 100
+	assert.Nil(t, msg.ValidateBasic())
+
+	// [100, 1]
```
Olshansk:
```suggestion
	// rewardShareMap: {delegator1: 100. delegator2: 1}
```

x/nodes/types/msg_test.go
```diff
@@ -663,6 +664,47 @@ func TestMsgStake_ValidateBasic(t *testing.T) {
...
+	msg.Delegators[delegator2.PubKey().Address().String()] = 1
+	err = msg.ValidateBasic()
+	assert.NotNil(t, err)
+	assert.Equal(t, CodeInvalidaDelegators, err.Code())
+
+	// [99, 1]
```
Olshansk:
```suggestion
	// rewardShareMap: {delegator1: 99, delegator2: 1}
```

x/nodes/types/msg_test.go
```diff
@@ -663,6 +664,47 @@ func TestMsgStake_ValidateBasic(t *testing.T) {
 	}
 }
 
+func TestMsgStake_Delegators(t *testing.T) {
```
Olshansk, line 667: Wdyt of having an upper bound on the number of delegators? I think you can imagine the on-chain bloat attacks this could cause.

x/nodes/keeper/valStateChanges.go
```diff
@@ -261,6 +261,15 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentValidator, newValidtor typ
 				return types.ErrUnequalOutputAddr(k.Codespace())
 			}
 		}
+
+		// Delegators can be set/edited only if 1) the feature has been activated
+		// AND 2) the message is signed by the operator address.
```
Olshansk:
```suggestion
		// Delegators can be set/edited only if:
		// 1) The feature has been activated AND
		// 2) the message is signed by the operator address.
```

x/nodes/keeper/valStateChanges.go
```diff
@@ -261,6 +261,15 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentValidator, newValidtor typ
...
+		if k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) &&
+			!types.CompareStringMaps(currentValidator.Delegators, newValidtor.Delegators) &&
+			!signer.Equals(currentValidator.Address) {
+			return types.ErrDisallowedOutputAddressEdit(k.Codespace())
```
Olshansk: Should this be a `DelegatorEdit` error?

Olshansk (changes requested):
Still not a complete review but posting my additional comments so far.

Screenshot 2023-12-14 at 4 40 38 PM

x/nodes/handler.go
msmania: I think ignoring unmarshaled fields is a general behavior.

x/nodes/handler.go
msmania: This is to keep the original behavior that `NewValidator` does not set the `StakedTokens` field.  Updated the leading comment.

x/nodes/keeper/reward.go
msmania: This is brand-new logic.  I think `IsPositive` is not needed because `rewardCost` is from `GetFee`, but having it doesn't do harm.  Let's add it.

x/nodes/keeper/valStateChanges.go
msmania: I was lazy not to introduce a new error.  Added it.

proto/x/nodes/nodes.proto
msmania: V8 means 0.8.  At first I followed the same as the version 0.8 commit ([this](https://github.com/pokt-network/pocket-core/commit/32157db43c88fcd37d5c68978f0256c8631db833)) introduced `LegacyProtoValidator`, but I realized just appending a field was enough as commented in another place.  Let's cut `ProtoValidatorV8`.

x/pocketcore/keeper/claim.go
msmania: The first variable returned here is the expected amount of the servicer's reward, so I think this name is correct.  Updated the leading comment of `CalculateRelayReward`.

x/pocketcore/keeper/claim.go
msmania: seems I was rushing here :)

x/nodes/types/validator_legacy.go
msmania: I forgot why I named it `ExactEqualsTo`.  Renamed it to just `Equals`.

> Can you add a comment on why we're implementing a custom Equals operation?

Isn't it clear why we need this?

x/nodes/types/validator_legacy.go
msmania: It meant v0.8.  Dropped this struct.

x/nodes/types/msg_test.go
msmania: Good point.  Instead of directly limiting the number of delegators, I decided to disallow zero share delegators.  With that, the number of delegators is limited to 100 because the total share cannot exceed 100.

x/nodes/keeper/reward.go
msmania: `rewardCost` is the cost a servicer needs to pay to earn relay rewards, which is the total of the transaction fee of claim and proof (= 20000 upokt).  With this patch, the network distributes this `rewardCost` to the servicer before distributing portions to the output address and delegators so that node operators don't have to maintain the balance of nodes' operator wallets.

x/nodes/keeper/reward.go
msmania: > 1. NIT for readability:

Instead of having `splitRewards` in two places, I introduced `calculateRewardRewardPip22` to put the PIP-22 logic in.

> 2. I compare this to the previous code and there were no changes, right?

Right, no change in `CalculateRelayReward`.  I don't like horizontal scroll, so I split long lines.

x/nodes/keeper/reward.go
msmania: We can, but I'd prefer this way.  Let's not have additional `if` blocks to keep the function simplified in this case.

x/nodes/keeper/reward.go
msmania: Each delegator share is a percentage, so we will not run out of funds.  In addition, I decided to disallow zero share in the `delegators` field, meaning the maximum case is the delegators field has 100 entries where each delegator gets 1% share.

x/nodes/keeper/reward_test.go
msmania: For 1), it does.  Assert doesn't stop the test.

I agree with 2) that looks smarter.  Will change it.  Practically, however, having this kind of false condition is much easier to debug because we can set a breakpoint there.

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   # TODO: Make builds for all platforms where Pocket Core is expected to run
   build:
     docker:
-      - image: cimg/go:1.18
+      - image: cimg/go:1.21
```
Olshansk: I realize you probably added this to make ci pass, but configuration changes like this can be risky and I think we should do it in a separate PR. I'm sure you've seen stuff like this in the past.

We can likely merge it quickly BEFORE this if we split it out.

CONTRIBUTING.md
```diff
@@ -38,7 +38,7 @@ Within your fork you are free to work however you want, but keep in mind that in
 
 #### Setting up the Go Environment
 
-Please follow the [Official Installation Guide](https://go.dev/doc/install) to complete this step. Pocket Core uses `go 1.18` so make sure to install the appropiate version before beginning development.
+Please follow the [Official Installation Guide](https://go.dev/doc/install) to complete this step. Pocket Core uses `go 1.21` so make sure to install the appropiate version before beginning development.
```
Olshansk: Again, I think 1.18 -> 1.21 is outside the scope of this already huge change.

app/cmd/rpc/rpc_test.go
```diff
@@ -1636,3 +1634,102 @@ func NewValidChallengeProof(t *testing.T, privateKeys []crypto.PrivateKey) (chal
...
+	if err != nil {
+		return "", err
+	}
+	outputAddr, err := types.AddressFromHex("fe818527cd743866c1db6bdeb18731d04891df78")
+	if err != nil {
+		return "", err
+	}
+	msg := &types2.MsgStake{
+		PublicKey:  privKey.PublicKey(),
+		Chains:     []string{"DEAD", "BEEF"},
```
Olshansk, line 1658: giphy

x/nodes/handler.go
```diff
@@ -59,10 +60,20 @@ func handleStake(ctx sdk.Ctx, msg types.MsgStake, k keeper.Keeper, signer crypto
...
+		}
+	} else if msg.Delegators != nil {
+		// Ignore the delegators field before the upgrade
+		msg.Delegators = nil
+	}
+
+	validator := types.NewValidatorFromMsg(msg)
+	// We used to use NewValidator here to initialize the `validator`, and
+	// it does not set the field StakedTokens.  To keep the same behavior,
+	// we reset this field to 0.  StakedTokens is set through StakeValidator.
```
Olshansk: I don't fully understand this comment. 

When we say `it does not set the field StakeTokens`, are you referring to the existing behaviour or the `NewValidatorFromMsg` function?

1. If it's the existing (previous) behaviour, we should update the comment since it doesn't make sense outside the scope of this PR.
2. If it's the function itself, then I don't understand why we need line 76.

#PUC

x/nodes/keeper/reward.go
```diff
@@ -1,19 +1,82 @@
...
 // RewardForRelays - Award coins to an address using the default multiplier
 func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Address) sdk.BigInt {
 	return k.RewardForRelaysPerChain(ctx, "", relays, address)
 }
 
+func (k Keeper) calculateRewardRewardPip22(
+	ctx sdk.Ctx,
+	relays, stake, multiplier sdk.BigInt,
+) sdk.BigInt {
+	//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
```
Olshansk:
```suggestion
	// floorstake to the lowest bin multiple or take ceiling, whichever is smaller
```

x/nodes/keeper/reward.go
```diff
@@ -1,19 +1,82 @@
...
+	flooredStake := sdk.MinInt(
+		stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))),
+		k.ServicerStakeWeightCeiling(ctx).
+			Sub(k.ServicerStakeWeightCeiling(ctx).
+				Mod(k.ServicerStakeFloorMultiplier(ctx))),
+	)
+	//Convert from tokens to a BIN number
```
Olshansk: NIT: White space at the beginning of line here and below.

x/nodes/keeper/validator_test.go
```diff
@@ -148,3 +149,66 @@ func Test_sortNoLongerStakedValidators(t *testing.T) {
...
+	// Validator --> []byte --> Validator
+	marshaled, err = Marshal(&val_1)
+	assert.Nil(t, err)
+	assert.NotNil(t, marshaled)
+	val_2.Reset()
+	err = Unmarshal(marshaled, &val_2)
+	assert.Nil(t, err)
+	assert.True(t, val_2.ToLegacy().Equals(val_1.ToLegacy()))
+	assert.True(t, val_2.OutputAddress.Equals(val_1.OutputAddress))
+	assert.True(t, types.CompareStringMaps(val_2.Delegators, val_1.Delegators))
```
Olshansk: Can you add an assertion to make sure the delegators are also not empty?

x/nodes/keeper/reward.go
```diff
@@ -45,41 +112,99 @@ func (k Keeper) RewardForRelaysPerChain(ctx sdk.Ctx, chain string, relays sdk.Bi
...
-		//truncate back to int
-		coins = coinsDecimal.TruncateInt()
-	} else {
-		coins = multiplier.Mul(relays)
+	// After the delegator upgrade, we compensate a servicer's operator wallet
+	// for the transaction fee of claim and proof.
+	if k.Cdc.IsAfterDelegatorUpgrade(ctx.BlockHeight()) {
+		rewardCost := k.GetRewardCost(ctx)
+		if toNode.LT(rewardCost) {
+			rewardCost = toNode
```
Olshansk, line 137: Following up on this.

If rewardCost > actualReward, we reward the node with a bigger amount. 

Is this correct? If so, #PUC

x/nodes/keeper/valStateChanges_test.go
```diff
@@ -498,6 +493,97 @@ func TestValidatorStateChange_OutputAddressEdit(t *testing.T) {
...
+	// Attempt to set a delegators before the upgrade --> The field is ignored
+	assert.Nil(t, runStake(operatorPubKey1, singleDelegator, outputPubKey))
+	validatorCur, found := k.GetValidator(ctx, operatorAddr1)
+	assert.True(t, found)
+	assert.Nil(t, validatorCur.Delegators)
+
+	// Enable DELE
+	codec.UpgradeFeatureMap[codec.RewardsDelegatorKey] = -1
+
+	// Attempt to change the delegators with output's signature --> Fail
```
Olshansk, line 559: Was this part of the proposal passed? Output can't change delegators but the operator can?

Olshansk (changes requested):
Reviewed everything except for `x/nodes/keeper/reward_test.go` and I think we're really close!

I don't know about you, but I think this reads SO much better than the previous iteration. Thanks for all the effort!

Screenshot 2023-12-20 at 9 12 20 PM

x/nodes/keeper/valStateChanges_test.go
msmania: Yes.  See this https://forum.pokt.network/t/pip32-unleashing-the-potential-of-non-custodial-node-running/4796.

> Another topic we need to discuss is “editing”. The upgrade to RC-0.10.4 enabled the new feature Output Address Edit [3], which allows an MsgStake transaction signed by the current output address to change the output address to another. Now who should be able to change delegators? We propose that only the owner of the node operator should have the authority to make changes as follows:

.circleci/config.yml
msmania: Yes, build failed in a docker due to generic.  I created a separate PR https://github.com/pokt-network/pocket-core/pull/1587.

x/nodes/handler.go
msmania: It's 1).  The new `NewValidatorFromMsg` sets `StakedTokens` while the existing `NewValidator` doesn't.  The line:76 is to fill that gap.  Ideally I should update `NewValidator`, but I wanted to keep the patch smaller.  Updated the comment.

x/nodes/keeper/reward.go
msmania: The actual amount of relay rewards (= servicer's portion 85%) is `toNode`.  If `rewardCost > toNode`, we run `rewardCost = toNode` and reward the servicer with all of the servicer's portion, meaning no rewards for output address and delegators.  Updated the comment.

x/nodes/keeper/valStateChanges_test.go
Olshansk: Can you #PUC with the link?

x/nodes/types/errors.go
```diff
@@ -37,7 +37,7 @@ const (
 	CodeUnauthorizedSigner          CodeType          = 125
 	CodeNilSigner                   CodeType          = 126
 	CodeDisallowedOutputAddressEdit CodeType          = 127
-	CodeInvalidaDelegators          CodeType          = 128
+	CodeInvalidDelegators           CodeType          = 128
```
Olshansk: NIT: RewardDelegator 

Update downstream functions & comments too.

Olshansk (changes requested):
@msmania I took another look and I think this is as good of a review as I can do too (reaching PR exhaustion).

I added one small NIT, one #PUC request, but otherwise the only blocker is to get the CI green after we merge the other PR.

msmania:
> @msmania I took another look and I think this is as good of a review as I can do too (reaching PR exhaustion).
> 
> I added one small NIT, one #PUC request, but otherwise the only blocker is to get the CI green after we merge the other PR.

Thank you!  I created https://github.com/pokt-network/pocket-core/pull/1589 to get the CI green.

msmania:
> @msmania I took another look and I think this is as good of a review as I can do too (reaching PR exhaustion).
> 
> I added one small NIT, one #PUC request, but otherwise the only blocker is to get the CI green after we merge the other PR.

Comparing [the test results](https://app.circleci.com/pipelines/github/pokt-network/pocket-core/2979/workflows/547aac82-2b84-442b-adbf-9a0c02c17906/jobs/5618) of #1589 to the [results](https://app.circleci.com/pipelines/github/pokt-network/pocket-core/2967/workflows/2cdb732f-10ed-4195-a5b5-f4a9799e5faa/jobs/5594) of #1587, the following tests failed in both of the test runs.

- TestValidateGenesis/Test_ValidateGenesis_8
- TestDefaultParams

The difference is TestRPC_QueryUnconfirmedTxs, which I think is flaky and can be ignored as it failed on #1582 too as mentioned [here](https://github.com/pokt-network/pocket-core/pull/1582#issuecomment-1841021778).  So I think this is good to be merged.  What do you think, @Olshansk?

Olshansk (dismissed):
@msmania Going to approve this since we've iterated quite a bit and it's the last day before the holidays. 

However, three requests:
1. Please make sure to get a 🟢  in #1589
2. Please use `squash and merge` and merging into main.
3. Please `git merge` instead of `git rebase` in future PRs. It makes it MUCH easier for the reviewer to leverage the "changes since last review" feature.

msmania:
#1589 got a green light.

msmania:
@Olshansk Looks like a force-push revoked your approval.  Can you sign off again?

app/cmd/rpc/rpc_test.go
```diff
@@ -219,71 +215,73 @@ type RPCResultUnconfirmedTxsResponse struct {
...
-		amount, _ := resTX.StdTx.Msg.Value.Amount.Int64()
-		assert.Equal(t, amount, int64(1000))
-		assert.Equal(t, strings.ToLower(resTX.StdTx.Msg.Value.FromAddress), strings.ToLower(cb.GetAddress().String()))
-		assert.Equal(t, strings.ToLower(resTX.StdTx.Msg.Value.ToAddress), strings.ToLower(kp.GetAddress().String()))
-	}
-
-	cleanup()
-	stopCli()
-}
+// Disabling the test because this is flaky.
```
Olshansk: Uncomment this and use `t.skip("TODO: Figure out why this test is flaky")` instead.

Olshansk (dismissed):
@msmania Left one small nit / comment, so please tend to that before merging.

Note: No matter how many times I review this, I'm still a bit worried (i.e. chain halt) given the size and scope of the change, so let's make sure to do our due diligence in testing once we're in TestNet.

**--- Regarding Git Flow ---** 

Bullet points with extra context below.

- I left a comment related to this in one of the PRs a couple weeks ago with a bit more details [1]
- For some extra *"cred"*, even Mitchell Hashimoto understood the value of using this GitHub workflow
- We documented this in the `pocket` repo but not the `pocket-core` one unfortunately. I'll make sure we transfer it into the  `poktroll` repo
- I want to emphasize that this is not a `how Pocket does thing` but a `How GitHub expects users to use Git`. A friend of mine is a PM there and I've brought this up a few times.

[1] https://github.com/pokt-network/pocket-core/pull/1581#pullrequestreview-1795308586
[2] https://x.com/olshansky/status/1708314683738820998?s=20
[3] https://github.com/pokt-network/pocket/blob/main/docs/development/CODE_REVIEW_GUIDELINES.md

Olshansk (approved):

POKT-Discourse:

PR #1580, opened 2023-09-07
[v0.11 feature] Per-chain Relays To Token Multiplier

msmania:
## Description
This patch introduces a new network parameter `pos/RelaysToTokensMultiplierMap` so that we can set a custom RTTM per chain.  The existing parameter `pos/RelaysToTokensMultiplier` serves the default RTTM if a custom RTTM is not set.

This change is consensus-breaking.  The new behavior is put behind a new feature key `RTTM2`.  The new parameter is kept unavailable until activation.

The patch also contains two tests:
- `Test_RTTM2_ChangeParamValue` to make sure the new parameter is kept unavailable until activating the `RTTM2` feature key.
- `TestKeeper_RewardForRelaysPerChain` to make sure the amount of relay rewards are calculated with a custom RTTM in the new parameter if it's set.

### Summary generated by Reviewpad on 07 Sep 23 00:52 UTC
This pull request contains the following changes:

1. The addition of a new test function called `TestKeeper_RewardForRelaysPerChain`. This test function verifies the rewards generated for different scenarios related to relays per chain.

2. Modifications to the file "expectedKeepers.go" were made. A new function called "RewardForRelaysPerChain" was added, which calculates and returns a value of type `sdk.BigInt`.

3. Changes were made to the file "module.go" related to activating additional parameters and setting a minimum signed per window value.

4. The file "keeper_test.go" includes the addition of a new test function called "Test_RTTM2_ChangeParamValue", which tests the change of a parameter value after a certain activation height.

5. A new file "proof.go" was modified to enhance the functionality and reliability of the `ExecuteProof` function in handling relay proofs.

6. The file "param.go" now includes additional tracking of parameters introduced after genesis and skips writing them to the database until the corresponding feature activation height is reached.

7. The file "nodes.go" includes changes to the `AwardCoinsForRelays` function, enabling it to award coins for relays completed per chain.

8. Modifications were made to the `PosKeeper` interface in the file "expectedKeepers.go", adding a new function called `RewardForRelaysPerChain`.

9. The file "service_test.go" includes changes to the `RewardForRelays` method in the `MockPosKeeper` struct, allowing coins to be awarded for relays completed per chain.

10. The file "abci.go" in the `x/nodes/keeper` package includes import statement additions and changes to variable names.

11. The go.sum file shows updates to dependencies related to the "github.com/pokt-network/tendermint" package.

These changes enhance functionality, introduce new features, and improve reliability in various parts of the codebase.

PR #1579, opened 2023-09-06
Fix infinite recursion in Session.HashString

cristaloleg:
## Description

Fixes #1578 

### Summary generated by Reviewpad on 06 Sep 23 07:14 UTC
This pull request fixes an issue with infinite recursion in the `HashString` method of the `Session` struct. The patch adds a fix to the method by returning `s.SessionHeader.HashString()` instead of calling `s.HashString()` recursively. Additionally, there are some code formatting changes and comments added for clarification.

PR #1582, opened 2023-10-18
feat: enforce max chains during new session

### LINKED ISSUE #1584: [Session] MaxChains Param Check at Start
During the 09/12/2023 V0 Builder Call - Shane | Nodepilot suggested a session start qualification check of the MaxChains parameter to make sure a Servicer was not violating the current limit (set to 15 as per gov)

Current session start checks include:
Jailed
Stake Value
Geo

Goal
Set up enforcement of the MaxChains parameter that is currently missing
Future proofs any changes to the actual parameter value (if gov were to change it, it would only be enforced if a Servicer updated their stake on the network)

h5law:
### Human Summary

This PR addresses the blocker for the [Gandalf](https://forum.pokt.network/t/pre-proposal-gandalf-decrease-maximumchains/4648) proposal. It **does not** implement any of the changes proposed in this proposal but instead enforces the `MaxChains` parameter which is currently unforced to an extent. It would prevent a node from joining a session if they are staked to more chains than allowed. This enables the future changes to this parameter to be enforceable on-chain by making sure any node joining a session is abiding by this parameter.

Fixes #1584 

### Summary generated by Reviewpad on 06 Dec 23 01:50 UTC
This pull request includes several changes that affect multiple files:

1. In the "claim.go" file:
   - An import statement has been added for "github.com/tendermint/tendermint/rpc/client".
   - The function "SendClaimTx" has been modified to include multiple arguments on separate lines for better readability.
   - An additional condition has been added to the if statement inside the "SendClaimTx" function.
   - Several log statements inside the "SendClaimTx" function have been modified to use chaining for better readability.
   - The function "ValidateClaim" has been modified to add a new error condition for checking the maximum number of chains an app can be staked to.
   - The "ValidateClaim" function now checks the length of the app's chains and compares it to the maximum allowed chains.
   - Formatting changes have been made to the log statements inside the "ValidateClaim" function for better readability.
   - The variable declaration for "msg" inside the "DeleteExpiredClaims" function has been modified to use the shorthand syntax.

2. In the "run" job:
   - The test command timeout has been increased to 15 minutes by adding the argument "-timeout=15m". This change ensures that the tests have enough time to complete.

3. In the "go.mod" file:
   - The version of the "github.com/google/go-cmp" package has been updated from "v0.5.6" to "v0.5.8".
   - The version of the "golang.org/x/sys" package has been updated from "v0.1.0" to "v0.14.0".

4. In the "x/pocketcore/types/service.go" file:
   - The import statements have been rearranged and reorganized.
   - The "Validate" function of the "Relay" struct has been modified to accept additional parameters and has been formatted with proper line breaks for better readability.
   - An additional validation has been added to check if the number of chains an app is staked to exceeds the permitted limit.
   - Some variable assignments and conditional statements have been modified for clarity and consistency.
   - The "sessionEndCtx" variable has been corrected to be properly assigned based on the block height.
   - Some comments within the code have been updated for better clarity.

5. In the "codec.go" file:
   - A new constant "EnforceMaxChainsUpdateKey" has been added.
   - A new method "IsAfterEnforceMaxChainsUpgrade" has been added to check if the height is after a specific upgrade.
   - The "IsOnNonCustodialUpgrade" method has been updated to include the height in the check.
   - The "IsAfterNamedFeatureActivationHeight" method has been updated to include the height in the check.
   - The "SliceToExistingMap" function has been updated to use shorthand declaration for the "fmap" variable.
   - The "SliceToMap" function has been updated to use shorthand declaration for the "fmap" variable.
   - The "MapToSlice" function has been updated to use shorthand declaration for the "fslice" variable.

6. In the "expectedKeepers.go" file:
   - The function "MaxChains" has been added to the "PosKeeper" interface.
   - The function "MaxChains" has been added to the "AppsKeeper" interface.

7. In the "error_codes.go" file:
   - A new error code "CodeChainsOverLimitError" with a value of 91 has been added.
   - A new error "ChainsOverLimitError" with a message "the number of staked chains is over the limit" has been added.
   - A new function "NewChainsOverLimitError" has been added that creates an error with the "CodeChainsOverLimitError" code and a formatted error message that includes the number of chains that were received and the maximum allowed chains.

8. In the "session.go" file:
   - In the "Session" struct, a new import statement for the "log" package has been added. Additionally, the "HashString" method has been modified to return "s.SessionHeader.HashString()" instead of "s.HashString()".
   - The "NewSession" function has been modified to add new parameters and adjust the parameter alignment. The added parameters are "sessionNodesCount" and "blockHash".
   - The "NewSessionNodes" function has been modified to add new parameters and adjust the parameter alignment. The added parameters are "sessionKey" and "sessionNodesCount".
   - The "NewSessionNodes" function also includes additional code related to retrieving values from the "ctx" and "keeper", as well as checking node conditions and adding nodes to the session.
   - The "MaxPossibleRelays" function includes a comment correction and a rounding operation for the returned value.

9. In the "service_test.go" file:
   - Lines 26 and 27: The variables "ethereum" and "bitcoin" are modified to use octal escape sequences instead of hexadecimal ones. The value "01" is now "0o1" and the value "02" is now "0o2".
   - Lines 148 and 152: The variable "ethereum" is modified to use an octal escape sequence instead of a hexadecimal one. The value "01" is now "0o1".
   - Lines 197 and 201: The variable "ethereum" is modified to use an octal escape sequence instead of a hexadecimal one. The value "01" is now "0o1".
   - Lines 326 and 330: A new method "MaxChains" is added to the "MockAppsKeeper" struct. This method returns the value 15.
   - Lines 397 and 401: A new method "MaxChains" is added to the "MockPosKeeper" struct. This method returns the value 15.

These changes aim to improve various aspects of the code, such as readability, maintainability, error handling, and usage of newer package versions. Please review these changes thoroughly.

### DIFF
--- a/x/nodes/keeper/reward.go
+++ b/x/nodes/keeper/reward.go
@@ -21,7 +21,7 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	// get validator
 	validator, found := k.GetValidator(ctx, address)
 
-	// adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
+	//adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
 	if !found && (isAfterRSCAL || isNonCustodialActive) {
 		ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
 		return sdk.ZeroInt()
@@ -47,17 +47,17 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 
 	var coins sdk.BigInt
 
-	// check if PIP22 is enabled, if so scale the rewards
+	//check if PIP22 is enabled, if so scale the rewards
 	if isAfterRSCAL {
 		stake := validator.GetTokens()
-		// floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
 		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), k.ServicerStakeWeightCeiling(ctx).Sub(k.ServicerStakeWeightCeiling(ctx).Mod(k.ServicerStakeFloorMultiplier(ctx))))
-		// Convert from tokens to a BIN number
+		//Convert from tokens to a BIN number
 		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
-		// calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
+		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
 		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), Pip22ExponentDenominator).Quo(k.ServicerStakeWeightMultiplier(ctx))
 		coinsDecimal := k.RelaysToTokensMultiplier(ctx).ToDec().Mul(relays.ToDec()).Mul(weight)
-		// truncate back to int
+		//truncate back to int
 		coins = coinsDecimal.TruncateInt()
 	} else {
 		coins = k.RelaysToTokensMultiplier(ctx).Mul(relays)
@@ -142,10 +142,11 @@ func (k Keeper) GetPreviousProposer(ctx sdk.Ctx) (addr sdk.Address) {
 	if b == nil {
 		k.Logger(ctx).Error("Previous proposer not set")
 		return nil
-		// os.Exit(1)
+		//os.Exit(1)
 	}
 	_ = k.Cdc.UnmarshalBinaryLengthPrefixed(b, &addr, ctx.BlockHeight())
 	return addr
+
 }
 
 // SetPreviousProposer -  Store proposer public key for this block
--- a/x/nodes/keeper/reward.go
+++ b/x/nodes/keeper/reward.go
@@ -14,7 +14,6 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	// feature flags
 	isAfterRSCAL := k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey)
 	isNonCustodialActive := k.Cdc.IsAfterNonCustodialUpgrade(ctx.BlockHeight())
-	isEnforceMaxChains := k.Cdc.IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight())
 
 	// The conditions of the original non-custodial issue:
 	isDuringFirstNonCustodialIssue := isNonCustodialActive && isAfterRSCAL && ctx.BlockHeight() <= codec.NonCustodial1RollbackHeight
@@ -25,13 +24,6 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	// adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
 	if !found && (isAfterRSCAL || isNonCustodialActive) {
 		ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
-		return sdk.ZeroInt()
-	}
-
-	// check if the node is overstaked for max chains and if so  and if so dont reward them
-	// this ensures that nodes can still participate in the network but are not rewarded for relays
-	// if they are overstaked for max chains, ensuring that the network functions properly
-	if isEnforceMaxChains && int64(len(validator.GetChains())) > k.GetNodeMaxChains(ctx) {
 		return sdk.ZeroInt()
 	}
 
--- a/x/nodes/keeper/reward.go
+++ b/x/nodes/keeper/reward.go
@@ -14,6 +14,7 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	// feature flags
 	isAfterRSCAL := k.Cdc.IsAfterNamedFeatureActivationHeight(ctx.BlockHeight(), codec.RSCALKey)
 	isNonCustodialActive := k.Cdc.IsAfterNonCustodialUpgrade(ctx.BlockHeight())
+	isEnforceMaxChains := k.Cdc.IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight())
 
 	// The conditions of the original non-custodial issue:
 	isDuringFirstNonCustodialIssue := isNonCustodialActive && isAfterRSCAL && ctx.BlockHeight() <= codec.NonCustodial1RollbackHeight
@@ -21,9 +22,16 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 	// get validator
 	validator, found := k.GetValidator(ctx, address)
 
-	//adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
+	// adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
 	if !found && (isAfterRSCAL || isNonCustodialActive) {
 		ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
+		return sdk.ZeroInt()
+	}
+
+	// check if the node is overstaked for max chains and if so  and if so dont reward them
+	// this ensures that nodes can still participate in the network but are not rewarded for relays
+	// if they are overstaked for max chains, ensuring that the network functions properly
+	if isEnforceMaxChains && int64(len(validator.GetChains())) > k.GetNodeMaxChains(ctx) {
 		return sdk.ZeroInt()
 	}
 
@@ -47,17 +55,17 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
 
 	var coins sdk.BigInt
 
-	//check if PIP22 is enabled, if so scale the rewards
+	// check if PIP22 is enabled, if so scale the rewards
 	if isAfterRSCAL {
 		stake := validator.GetTokens()
-		//floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
+		// floorstake to the lowest bin multiple or take ceiling, whicherver is smaller
 		flooredStake := sdk.MinInt(stake.Sub(stake.Mod(k.ServicerStakeFloorMultiplier(ctx))), k.ServicerStakeWeightCeiling(ctx).Sub(k.ServicerStakeWeightCeiling(ctx).Mod(k.ServicerStakeFloorMultiplier(ctx))))
-		//Convert from tokens to a BIN number
+		// Convert from tokens to a BIN number
 		bin := flooredStake.Quo(k.ServicerStakeFloorMultiplier(ctx))
-		//calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
+		// calculate the weight value, weight will be a floatng point number so cast to DEC here and then truncate back to big int
 		weight := bin.ToDec().FracPow(k.ServicerStakeFloorMultiplierExponent(ctx), Pip22ExponentDenominator).Quo(k.ServicerStakeWeightMultiplier(ctx))
 		coinsDecimal := k.RelaysToTokensMultiplier(ctx).ToDec().Mul(relays.ToDec()).Mul(weight)
-		//truncate back to int
+		// truncate back to int
 		coins = coinsDecimal.TruncateInt()
 	} else {
 		coins = k.RelaysToTokensMultiplier(ctx).Mul(relays)
@@ -142,11 +150,10 @@ func (k Keeper) GetPreviousProposer(ctx sdk.Ctx) (addr sdk.Address) {
 	if b == nil {
 		k.Logger(ctx).Error("Previous proposer not set")
 		return nil
-		//os.Exit(1)
+		// os.Exit(1)
 	}
 	_ = k.Cdc.UnmarshalBinaryLengthPrefixed(b, &addr, ctx.BlockHeight())
 	return addr
-
 }
 
 // SetPreviousProposer -  Store proposer public key for this block
--- a/x/nodes/keeper/validator.go
+++ b/x/nodes/keeper/validator.go
@@ -234,7 +234,8 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 	return validators
 }
 
-// GetNodeMaxChains - Retrieve the maximum number of chains a node can have
+// GetNodeMaxChains - Retrieve the maximum number of chains a node can have at the height
+// associated with the context
 func (k Keeper) GetNodeMaxChains(ctx sdk.Ctx) int64 {
 	return k.MaxChains(ctx)
 }
--- a/x/nodes/keeper/validator.go
+++ b/x/nodes/keeper/validator.go
@@ -188,7 +188,8 @@ func (k Keeper) ClearSessionCache() {
 
 // IterateAndExecuteOverVals - Goes through the validator set and executes handler
 func (k Keeper) IterateAndExecuteOverVals(
-	ctx sdk.Ctx, fn func(index int64, validator exported.ValidatorI) (stop bool)) {
+	ctx sdk.Ctx, fn func(index int64, validator exported.ValidatorI) (stop bool),
+) {
 	store := ctx.KVStore(k.storeKey)
 	iterator, _ := sdk.KVStorePrefixIterator(store, types.AllValidatorsKey)
 	defer iterator.Close()
@@ -231,6 +232,11 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 		validators = append(validators, validator)
 	}
 	return validators
+}
+
+// ValidatorHasLessOrEqualMaxChains - Check if the validator has less or equal than the maximum chains param
+func (k Keeper) ValidatorHasLessOrEqualMaxChains(ctx sdk.Ctx, val exported.ValidatorI) bool {
+	return int64(len(val.GetChains())) <= k.GetParams(ctx).MaximumChains
 }
 
 // map of validator addresses to serialized power
--- a/x/nodes/keeper/validator.go
+++ b/x/nodes/keeper/validator.go
@@ -234,12 +234,6 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 	return validators
 }
 
-// GetNodeMaxChains - Retrieve the maximum number of chains a node can have at the height
-// associated with the context
-func (k Keeper) GetNodeMaxChains(ctx sdk.Ctx) int64 {
-	return k.MaxChains(ctx)
-}
-
 // map of validator addresses to serialized power
 type valPowerMap map[[sdk.AddrLen]byte][]byte
 
--- a/x/nodes/keeper/validator.go
+++ b/x/nodes/keeper/validator.go
@@ -188,8 +188,7 @@ func (k Keeper) ClearSessionCache() {
 
 // IterateAndExecuteOverVals - Goes through the validator set and executes handler
 func (k Keeper) IterateAndExecuteOverVals(
-	ctx sdk.Ctx, fn func(index int64, validator exported.ValidatorI) (stop bool),
-) {
+	ctx sdk.Ctx, fn func(index int64, validator exported.ValidatorI) (stop bool)) {
 	store := ctx.KVStore(k.storeKey)
 	iterator, _ := sdk.KVStorePrefixIterator(store, types.AllValidatorsKey)
 	defer iterator.Close()
--- a/x/nodes/keeper/validator.go
+++ b/x/nodes/keeper/validator.go
@@ -234,9 +234,9 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 	return validators
 }
 
-// ValidatorHasLessOrEqualMaxChains - Check if the validator has less or equal than the maximum chains param
-func (k Keeper) ValidatorHasLessOrEqualMaxChains(ctx sdk.Ctx, val exported.ValidatorI) bool {
-	return int64(len(val.GetChains())) <= k.GetParams(ctx).MaximumChains
+// GetNodeMaxChains - Retrieve the maximum number of chains a node can have
+func (k Keeper) GetNodeMaxChains(ctx sdk.Ctx) int64 {
+	return k.MaxChains(ctx)
 }
 
 // map of validator addresses to serialized power
--- a/x/nodes/types/validator.go
+++ b/x/nodes/types/validator.go
@@ -4,10 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/pokt-network/pocket-core/codec"
 	"strings"
 	"time"
 
-	"github.com/pokt-network/pocket-core/codec"
 	"github.com/pokt-network/pocket-core/crypto"
 
 	sdk "github.com/pokt-network/pocket-core/types"
--- a/x/nodes/types/validator.go
+++ b/x/nodes/types/validator.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/pokt-network/pocket-core/codec"
-
 	"github.com/pokt-network/pocket-core/crypto"
 
 	sdk "github.com/pokt-network/pocket-core/types"
--- a/x/nodes/types/validator.go
+++ b/x/nodes/types/validator.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/pokt-network/pocket-core/codec"
 	"strings"
 	"time"
+
+	"github.com/pokt-network/pocket-core/codec"
 
 	"github.com/pokt-network/pocket-core/crypto"
 
--- a/x/pocketcore/keeper/claim.go
+++ b/x/pocketcore/keeper/claim.go
@@ -5,13 +5,14 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/tendermint/tendermint/rpc/client"
+
 	"github.com/pokt-network/pocket-core/codec"
 	"github.com/pokt-network/pocket-core/crypto"
 	sdk "github.com/pokt-network/pocket-core/types"
 	"github.com/pokt-network/pocket-core/x/auth"
 	"github.com/pokt-network/pocket-core/x/auth/util"
 	pc "github.com/pokt-network/pocket-core/x/pocketcore/types"
-	"github.com/tendermint/tendermint/rpc/client"
 )
 
 // "SendClaimTx" - Automatically sends a claim of work/challenge based on relays or challenges stored.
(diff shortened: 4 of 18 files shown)


Olshansk:
@msmania Could you take a look at this review when you have a chance?

x/pocketcore/types/session.go
```diff
@@ -124,17 +125,21 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
-		// if not found or jailed, don't add to session and continue
-		if node == nil || node.IsJailed() || !NodeHasChain(chain, node) || sessionNodes.Contains(node.GetAddress()) {
+		// if not found or jailed or staked chains is too large, don't add to session and continue
+		if node == nil ||
+			node.IsJailed() ||
+			!NodeHasChain(chain, node) ||
+			sessionNodes.Contains(node.GetAddress()) ||
+			!keeper.ValidatorHasLessOrEqualMaxChains(ctx, node) {
```
msmania: This is a consensus breaking change.  Can you add a feature flag in codec/codec.go and apply this new condition only if the height >= the activation height?

x/nodes/keeper/validator.go
```diff
@@ -233,6 +234,11 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 	return validators
 }
 
+// ValidatorHasLessOrEqualMaxChains - Check if the validator has less or equal than the maximum chains param
+func (k Keeper) ValidatorHasLessOrEqualMaxChains(ctx sdk.Ctx, val exported.ValidatorI) bool {
```
msmania: I'd introduce a function `GetNodeMaxChains()` that just returns `k.MaxChains(ctx)` here
and do comparison in `NewSessionNodes`.  That's more straightforward and easier to understand than having a negated condition of a function named `ValidatorHasLessOrEqualMaxChains`.  What do you think?

x/nodes/types/validator.go
```diff
@@ -4,10 +4,11 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/pokt-network/pocket-core/codec"
 	"strings"
 	"time"
 
+	"github.com/pokt-network/pocket-core/codec"
+
 	"github.com/pokt-network/pocket-core/crypto"
```
msmania: Let's remove these empty lines between "github.com/..." at line 11 and 13.

msmania:
Can you add some description to explain why we need this?  Reviewpad covers *what* this patch does, but it doesn't cover *why* part.  And if this is to unblock [Gandalf](https://forum.pokt.network/t/pre-proposal-gandalf-decrease-maximumchains/4648), please mention it too.

msmania (changes requested):

x/pocketcore/types/session.go
h5law: How should I determine the height of activation?

Olshansk: Look here for some examples: https://github.com/pokt-network/pocket-core/blob/staging/codec/codec.go#L287

msmania: You don't need to determine the block height.  You just define a feature name as a string.  The height is determined when we submit `gov enable` command to activate the feature.  Here's a PR defining a new feature: https://github.com/pokt-network/pocket-core/pull/1534/files#diff-8b594c387517838ab44fd9d9f7a5f8c1e3efa32486a1ba4b923858e4e6538955

x/pocketcore/types/session.go
```diff
@@ -283,3 +288,7 @@ func NodeHasChain(chain string, node exported.ValidatorI) bool {
 	}
 	return hasChain
 }
+
+func InvalidMaxChains(ctx sdk.Ctx, keeper PosKeeper, node exported.ValidatorI) bool {
+	return keeper.Codec().IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight()) && keeper.GetNodeMaxChains(ctx) > int64(len(node.GetChains()))
```
msmania: Can we just use the global variable [`ModuleCdc`](https://github.com/pokt-network/pocket-core/blob/staging/x/pocketcore/types/codec.go#L12)?

x/pocketcore/types/session.go
```diff
@@ -124,17 +125,21 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
-		// if not found or jailed, don't add to session and continue
-		if node == nil || node.IsJailed() || !NodeHasChain(chain, node) || sessionNodes.Contains(node.GetAddress()) {
+		// if not found or jailed or staked chains is too large, don't add to session and continue
+		if node == nil ||
+			node.IsJailed() ||
+			!NodeHasChain(chain, node) ||
+			sessionNodes.Contains(node.GetAddress()) ||
+			InvalidMaxChains(ctx, keeper, node) {
```
msmania: It's efficient to evaluate `InvalidMaxChains` before `NodeHasChain` or `sessionNodes.Contains`.

msmania: In this case, defining `InvalidMaxChains` does not improve readability.  The condition is simple enough and if we have a separate function, people need to go to the function to see what it does.

And, calling `IsAfterEnforceMaxChainsUpgrade` and `GetNodeMaxChains` in the loop is not necessary.  These functions are lightweight, but still I'd prefer doing it once before the loop.  So the change would be:

Before the loop
```go
enforceMaxChain := ModuleCdc.IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight())
nodeMaxChains := keeper.GetNodeMaxChains(ctx)
```

In the loop
```go
// After the EnforceMaxChains upgrade, don't select a node with too many
// chains.
if enforceMaxChain && nodeMaxChains > int64(len(node.GetChains())) {
	continue
}
```

msmania: This may need some discussion.  Should we use `ctx` to retrieve the MaxChains parameter, or `sessionCtx`?

I think `sessionCtx` is a safer choice.  If we use `ctx`, selected nodes may change every block in the same session.  Using `sessionCtx` makes sure the same nodes are selected throughout a session.  The downside is a new value of MaxChains is not activated until the next session, but I believe it's acceptable.

What do you think?

msmania:
Thank you for adding the description.  Given that this upgrade is *EnforceMaxChains*, probably we may want to enforce app's MaxChains too as an opportunistic change?  The change would be to add a similar check in `Relay.Validate` and `ValidateClaim`.

RossiNYC:
@h5law - can you link this PR to https://github.com/orgs/pokt-network/projects/143?pane=issue&itemId=38406491

I think you're solving for it.

POKT-Discourse:

https://forum.pokt.network/t/pip-32-fix-maxchains-paramater-fmp/4805/1

h5law:
@msmania I have updated the PR to check the app has not exceeded max chains when validating relays and during claim validation. Is this what you had in mind?

msmania:
> @msmania I have updated the PR to check the app has not exceeded max chains when validating relays and during claim validation. Is this what you had in mind?

Yes, the logic is something like that, but you need to use `MaxChains` in the AppKeeper ([here](https://github.com/pokt-network/pocket-core/blob/RC-0.10.4/x/apps/types/params.go#L45)).  There are two parameters `pos/MaximumChains` and `application/MaximumChains`.

x/nodes/keeper/validator.go
```diff
@@ -233,6 +234,11 @@ func (k Keeper) AllValidators(ctx sdk.Ctx) (validators []exported.ValidatorI) {
 	return validators
 }
 
+// GetNodeMaxChains - Retrieve the maximum number of chains a node can have
```
Olshansk:
```suggestion
// GetNodeMaxChains - Retrieve the maximum number of chains a node can have at the height
// associated with the context
```

x/pocketcore/keeper/claim.go
```diff
@@ -131,6 +131,11 @@ func (k Keeper) ValidateClaim(ctx sdk.Ctx, claim pc.MsgClaim) (err sdk.Error) {
 			return pc.NewOverServiceError(pc.ModuleName)
 		}
 	}
+	// Ensure that the app is not staked to more than the permitted number of chains
```
Olshansk, line 147: NIT: Move `int64(len(app.GetChains())` into a local var for readability

x/pocketcore/keeper/claim.go
```diff
@@ -131,6 +131,11 @@ func (k Keeper) ValidateClaim(ctx sdk.Ctx, claim pc.MsgClaim) (err sdk.Error) {
...
+	if pc.ModuleCdc.IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight()) &&
+		int64(len(app.GetChains())) > k.posKeeper.GetNodeMaxChains(sessionContext) {
```
Olshansk: Why do you use the `posKeeper` in the check but the `appKepper` in the error msgt?

x/pocketcore/types/service.go
```diff
@@ -56,6 +57,11 @@ func (r *Relay) Validate(ctx sdk.Ctx, posKeeper PosKeeper, appsKeeper AppsKeeper
 	if !found {
 		return sdk.ZeroInt(), NewAppNotFoundError(ModuleName)
 	}
+	// Ensure that the app is not staked to more than the permitted number of chains
+	if ModuleCdc.IsAfterEnforceMaxChainsUpgrade(ctx.BlockHeight()) &&
```
Olshansk, line 70: Ditto my comments elsewhere

Olshansk (changes requested):

RossiNYC:
Just FYI @Olshansk - during the last builder call PNF decided this PR needs a DAO vote to get included into the next build - Shane is aware and is working to publish the PIP that will x-ref this.

msmania (changes requested):

x/pocketcore/types/session.go
h5law: I think you are right using the session context gives us more consistency

h5law: @msmania do you think using the session context for only the param check. I have it this way but using the current context for the blockheight (for the flag check only)

msmania (commented):
I should have noticed this earlier.  Does this approach, changing `NewSessionNodes`, cause service disruption?

I think most of nodes are hosting 15 chains right now.  If we reduce it to 3 or any small number, the loop in `NewSessionNodes` needs more iterations.  In the worst case, it enters an infinite loop if no nodes meet the `MaxChains` condition.

Even in a transition phase before all node runners adapt their nodes to a new `MaxChains`, we should keep the service quality as is.  To do that, I think the right approach would be a node with excessive chains still serve relays, but it won't earn service rewards.  What do you think?

h5law:
@msmania 

> I should have noticed this earlier.  Does this approach, changing `NewSessionNodes`, cause service disruption?

So to be clear this change will **not** cause service distruption as no nodes are able to stake higher than the 15 limit. However if the DAO decides to change this parameter in future then PNF should release a statement and people will have to change their max chains stakes to the new value before it happens on chain or else they will not be able to join the session.

> Even in a transition phase before all node runners adapt their nodes to a new `MaxChains`, we should keep the service quality as is.  To do that, I think the right approach would be a node with excessive chains still serve relays, but it won't earn service rewards.  What do you think?

I think that blocking service rewards is a good way to go. I think this would just be a check in the claim/proof lifecycle code where funds are distributed which just disables the sending of funds if the node is "overstaked". I can find this and add this in soon

msmania:
> So to be clear this change will **not** cause service distruption as no nodes are able to stake higher than the 15 limit. However if the DAO decides to change this parameter in future then PNF should release a statement and people will have to change their max chains stakes to the new value before it happens on chain or else they will not be able to join the session.

What I'm saying is, if all nodes hosting a chain have staked 15 chains, after the parameter is changed to a value smaller than 15, that chain stops working because a dispatch request enters an infinite loop.  Even if a half of nodes staked have staked excessive chains, a dispatch request gets slower because of more iterations in the loop.

Of course we release a statement and encourage node runners to restake, and of course they're incentivized to do so, but still, we cannot expect all node runners are quick and responsive.  QoS shouldn't be relay on this kind of off-chain human activity.

x/nodes/keeper/reward.go
```diff
@@ -14,19 +14,27 @@ func (k Keeper) RewardForRelays(ctx sdk.Ctx, relays sdk.BigInt, address sdk.Addr
...
+	// adding "&& (isAfterRSCAL || isAfterNonCustodial)" to sync from scratch as weighted stake and non-custodial introduced this requirement
 	if !found && (isAfterRSCAL || isNonCustodialActive) {
 		ctx.Logger().Error(fmt.Errorf("no validator found for address %s; at height %d\n", address.String(), ctx.BlockHeight()).Error())
 		return sdk.ZeroInt()
 	}
 
+	// check if the node is overstaked for max chains and if so  and if so dont reward them
+	// this ensures that nodes can still participate in the network but are not rewarded for relays
+	// if they are overstaked for max chains, ensuring that the network functions properly
+	if isEnforceMaxChains && int64(len(validator.GetChains())) > k.GetNodeMaxChains(ctx) {
```
h5law: @msmania So i removed the check to select nodes for a session and replaced it with a check in the rewarding of the node for a session. If they are overstaked they get no rewards. Is this more in line with your idea?

msa6867:
I have been out of the loop for a while, so I apologize for the very late input.   The updated approach deviates from the approach that was voted on (enforce at session).  I recommend going back to a session-enforced approach,  but rather than doing an enforcement check as to whether len (chain list) is LE MaxChain, simply ignore any entry in the node’s chain ID list beyond the first MaxChain entries when selecting nodes for a session.  That should solve most of the concerns raised by @msmania (about transition stability if not enough nodes drop chain list below maxchain ahead of a parameter update) without moving enforcement to rewards.  Enforcing at rewards is problematic for reasons discussed by myself and @RawthiL on the forum thread.

Please note that if you decide to go with this approach (ie, ignore entries beyond first MaxChain), then it will be good to have full awareness of how the database is ordered.... is the list ordered by data entry, or is it stored in numerically-ascending order of values, etc.  Expected behavior both in testnest and mainnet will be slightly different depending on the answer to this question.

POKT-Discourse:

https://forum.pokt.network/t/pip-34-fix-maxchains-paramater-fmp/4805/8

moatus:
> A much simpler approach would be the following. Pass the current MaxChain parameter to session selection code and update the session selection code so as to effectively ignore any chain ID in a node's list of chain IDs beyond the first MaxChain entries. The only downside is that enforcement would be done at the level of portal rather than protocol, so in a multi-portal environment, you would need to make sure each portal enforces this.

@msa6867 You suggest that enforcement of MaxChains should be on the gateways themselves, but that is much harder to enforce. Systems would have to be created to audit sessions and gateway traffic to ensure that gateways are enforcing a network wide rule. The network already has a consensus layer, to ensure that rules are being followed, so it makes way more sense to utilize it for MaxChains, as was it's intended design.

I personally felt that the network's own consensus layer made more sense then creating an off-chain mechanism of enforcement (which is essentially and off-chain consensus mechanism). This is why I did not pursue the gateway enforcement route and believe in this current route. The DAO has already voted on a consensus level change, so I would suggest you take alternative suggestions to the DAO/forum first. This PR is specifically focused on a consensus level fix so any discussion on an alternative gateway level mechanism should happen over there.

POKT-Discourse:

https://forum.pokt.network/t/pip-34-fix-maxchains-paramater-fmp/4805/10

RawthiL:
Really? why have you changed the mechanism after the vote started @h5law ?

Enabling an starvation attack and adding complexity to the RTTM calculation is not the right call. 
Enforcing on the session generation is the right way to go.
I think that the issue brought up by @msmania  can be addressed in an other way. The chances that no node runner will stake on a chain are really low. Why would this happen if the node runner already has the working blockchain node before the change? Also, we have Community Chains and bootstrap nodes to quickly fix this. These "abandoned" chains will mean constant and uncontested POKT for the first node runner that stakes.

msmania:
> @msmania do you think using the session context for only the param check. I have it this way but using the current context for the blockheight (for the flag check only)

Not in `RewardForRelays`.  I was thinkig of rejecting a claim in `ValidateClaim` if the sender has excessive chains.  It's very simple.

> Further, I am not sure why we are focused on consensus-breaking changes when this should be fixable without breaking consensus.

Updating the session selection code is consensus-breaking anyway.

h5law:
@RawthiL Ideally I would leave the change to node selection, however I think the claim validation check is better to avoid a (unlikely) bad state.

@msmania I have moved the chain check to the `ValidateClaim` method.

For anyone else the PR is small I would appreciate if you have comments about a specific line or area of the code could you leave your comment on that line, if you can, or at least link to it - this would be appreciated on my end just to make addressing things easier when context switching from Shannon.

@msa6867 I think your idea about gateway level enforcement is not ideal, this should be enforced at the protocol level. Anything that changes: session selection, rewards or claims, most things in regards to what we are talking about will be consensus breaking. About this touching numerous files a lot of the changes are simply my LSP formatting the code as it wasnt formatted properly before (gopls) this is in fact a very small PR.

@moatus @RawthiL I realise I made the change during the vote that moved away from session selection, I am following @msmania lead here. I think the vote was more focused on enforcing the parameter at the protocol level. Whether enforced through joining a session or being able to claim for it, the result is the same (no change at the moment). If the param were to change then we would want no issues. Enforcement at the session level opens a potential risk to QoS levels, enforcement at claim validation seems more appropriate for both apps and servicers.

Any feedback is welcomed this change should be vetted and the best approach should be taken

msa6867:
I have edited my comment above to clean it up and get rid of the negative tone.  Sorry about that.  Also  Ignore what I said about enforcing at gateway... still at session but with a methodology of ignoring all but first MaxChain entries... This mod should take care most of the QoS state concern raised by @msmania while allowing enforcement to still be at session.  

@h5law, yeah I saw upon going back that most of this was formatting, lol.

h5law:
@msmania @moatus @msa6867 @RawthiL WDYT of the first n selection logic?

RawthiL:
@h5law I think that the problem (which is not a real problem IMO) will persist, as the order of the chains in the stake command hardly follows any order. All the chains that are not on the "N" first positions will suffer the same problem as if the node was unstaked. The "problem" is not solved, its impact is only reduced.

Also, it would be interesting to know if the order of the vector elements is preserved when written and read during the marshaling/unarshaling of the json data (I'm a complete ignorant here, but I heard some concerns). If the read/write order is not strict than this "read n first chains" is not possible to implement.

My concern is that even with picking the "N" top chains, we will keep carrying nodes staked for more chains that what is allowed, creating non-compliances in the data for a reason that is not relevant enough. We should go for the original commit, there was no problem there, lazy node runners are not to be trusted for high QoS.

kinesis-agent:
First of all, thank you all for considering and discussing the concern I raised.  I believe it's necessary discussion.  And I really appreciate your diligence and patience @h5law.

> @msmania @moatus @msa6867 @RawthiL WDYT of the first n selection logic?

For this logic, I agree with @RawthiL.  Besides his points, I think it even deincentivizes node runners to restake their nodes.  If their 15-chains nodes still earn rewards, there is no reason for them to pay fees and reduce the chains.

I like this idea in  [RawthiL's post](https://forum.pokt.network/t/pip-34-fix-maxchains-paramater-fmp/4805/21).

> We can plan the enforcement of the max chain parameter. The same way we do with any other protocol update, we observe the chain data, when we are sure that we have at least 24 nodes on each chain with only one chain staked, we make the change.

When we decrease the param, before submitting a gov change tx, we can wait until the majority of nodes are restaked and make sure `NewSessionNodes` returns within a reasonable time.  Having this off-chain operation is not ideal, but we do if it's needed. With that, I agree with the first approach, not picking nodes with excessive chains in `NewSessionNodes`.

msmania:
First of all, thank you all for considering and discussing the concern I raised.  I believe it's necessary discussion.  And I really appreciate your diligence and patience @h5law.

> @msmania @moatus @msa6867 @RawthiL WDYT of the first n selection logic?

For this logic, I agree with @RawthiL.  Besides his points, I think it even deincentivizes node runners to restake their nodes.  If their 15-chains nodes still earn rewards, there is no reason for them to pay fees and reduce the chains.

I like this idea in  [RawthiL's post](https://forum.pokt.network/t/pip-34-fix-maxchains-paramater-fmp/4805/21).

> We can plan the enforcement of the max chain parameter. The same way we do with any other protocol update, we observe the chain data, when we are sure that we have at least 24 nodes on each chain with only one chain staked, we make the change.

When we decrease the param, before submitting a gov change tx, we can wait until the majority of nodes are restaked and make sure `NewSessionNodes` returns within a reasonable time.  Having this off-chain operation is not ideal, but we do if it's needed. With that, I agree with the first approach, not picking nodes with excessive chains in `NewSessionNodes`.

msa6867:
I agree with what @RawthiL is saying.   We have a pretty tight-knit community of node runners that are well connected to the foundation and each other.  It should not at all be a problem to signal the intention to lower the MaxChain value and get voluntary movement to the new level ahead of the actual parameter change, at least to sufficient level to ensure a smooth transition for apps, even if a long-tail of less-connected node runners see their nodes blocked at transition..

moatus:
Thanks for the update @h5law. I'm not a coder myself so I can't comment on the actual code, but if the logic follows what has been stated, then I believe this is good in my book.

As @msmania mentioned, we can have a coordinated approach prior to reducing chains. Granted this should happen only happen once (maybe twice depending on the DAO's decision) so making sure it's coordinated shouldn't be difficult. It's in the best interest of node runners to make sure their nodes are staked in the right order.

At the end of the day, any node runner wanting returns will need to update their stake, so no reason to try and work around that fact. If a provider is going to neglect their order for some reason, that is on them. We shouldn't need to put any more guard rails that what we already have.

I think this approach address all past concerns. Thanks for the update 👍

h5law:
Final update 🤞🏼 In a bid to keep things simple I reverted back to the original commit's logic. It seems we are in agreement that this doesn't pose as much a problem as previously thought, as we will coordinate the release as we do all others.

I chose to revert the logic simply because it makes the code simpler. Simple is easier to understand and this logic really makes the most sense I believe, I don't see a need to overcomplicate it.

I appreciate all your guys' (@msmania @RawthiL @moatus @msa6867) inputs and hope this final form is satisfactory.

RawthiL:
@h5law It would be nice if you comment this rollback in the forum too, so we have the subject settled for all audiences (not everyone reads the PR comments)

POKT-Discourse:

https://forum.pokt.network/t/pip-34-fix-maxchains-paramater-fmp/4805/24

x/pocketcore/types/session.go
```diff
@@ -17,11 +18,17 @@ func (s Session) IsSealable() bool {
 }
 
 func (s Session) HashString() string {
-	return s.HashString()
+	return s.SessionHeader.HashString()
```
msmania, line 21: This change should be https://github.com/pokt-network/pocket-core/pull/1579.

msmania (changes requested):
The change looks good.

One more ask.  Can you exclude files if the change there is only for linting.  Upon quick look, I think the following files have no actual logic change.

- app/cmd/rpc/rpc_test.go
- app/tx_test.go
- x/apps/keeper/application.go
- x/nodes/keeper/keeper.go
- x/nodes/keeper/reward.go
- x/nodes/keeper/validator.go
- x/nodes/types/validator.go

x/pocketcore/types/session.go
h5law: That pr is closed can we leave it in here?

Olshansk: Agreed. Given that the other author didn't merge it in, let's just keep it here.

x/pocketcore/types/service.go
```diff
@@ -56,6 +57,12 @@ func (r *Relay) Validate(ctx sdk.Ctx, posKeeper PosKeeper, appsKeeper AppsKeeper
 	if !found {
 		return sdk.ZeroInt(), NewAppNotFoundError(ModuleName)
 	}
+	// Ensure that the app is not staked to more than the permitted number of chains
+	lenAppChains := int64(len(app.GetChains()))
```
Olshansk: s/lenAppChains/numAppChains

`len` is the function we use to get the `num`

x/pocketcore/types/session.go
```diff
@@ -124,17 +139,22 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
 		m[n.String()] = struct{}{}
 
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
-		// if not found or jailed, don't add to session and continue
-		if node == nil || node.IsJailed() || !NodeHasChain(chain, node) || sessionNodes.Contains(node.GetAddress()) {
+		lenNodeChains := int64(len(node.GetChains()))
+		// if not found or jailed or is overstaked to chains
+		if node == nil ||
+			isEnforceMaxChains && lenNodeChains > keeper.MaxChains(ctx) ||
```
Olshansk: I'm not 100% sure but I think this could result in a bug. Better safe than sorry

```suggestion
			(isEnforceMaxChains && lenNodeChains > keeper.MaxChains(ctx)) ||
```

Olshansk (changes requested):
@h5law Can you add a unit test that verifies this before and after the feature flag is enabled? 

@msmania has a great example in #1580 if you need a reference

h5law:
Screenshot 2023-12-02 at 00 10 16

Added new tests but CI fails for another reason, I have no idea why this error occurs and its fully out of scope of this PR to attend to it. Expanding the timeout fixed it previously but this behaviour is strange and should not occur. Besides that @Olshansk @msmania this PR is ready to go

x/pocketcore/types/session_test.go
```diff
@@ -35,216 +43,462 @@ func TestSessionKey_Validate(t *testing.T) {
...
+		Jailed:                  false,
+		Status:                  sdk.Staked,
+		Chains:                  []string{ethereum},
+		ServiceURL:              "https://www.google.com:443",
+		StakedTokens:            sdk.NewInt(100000),
+		UnstakingCompletionTime: time.Time{},
+	}
+	node1 := nodesTypes.Validator{
+		Address:                 sdk.Address(fakePubKey1.Address()),
+		PublicKey:               (fakePubKey1),
```
Olshansk:
```suggestion
		PublicKey:               fakePubKey1,
```

x/pocketcore/types/session_test.go
```diff
@@ -35,216 +43,462 @@ func TestSessionKey_Validate(t *testing.T) {
...
+	allNodes[6] = node6
+	allNodes[7] = node7
+	allNodes[8] = node8
+	allNodes[9] = node9
+	allNodes[10] = node10
+	allNodes[11] = node11
+	sessionCtx := newContext(t, false)
+	ctx := newContext(t, false)
+	k := MockPosKeeper{Validators: allNodes}
+	sessionNodes, err := NewSessionNodes(
```
Olshansk, line 254: If you just add a comment saying "context before upgrade" and "context after upgrade". Are we going to need the copy-paste of all the preparation?

Olshansk (changes requested):

x/pocketcore/types/session.go
```diff
@@ -124,17 +139,22 @@ func NewSessionNodes(sessionCtx, ctx sdk.Ctx, keeper PosKeeper, chain string, se
...
 		m[n.String()] = struct{}{}
 
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
-		// if not found or jailed, don't add to session and continue
-		if node == nil || node.IsJailed() || !NodeHasChain(chain, node) || sessionNodes.Contains(node.GetAddress()) {
+		lenNodeChains := int64(len(node.GetChains()))
+		// if not found or jailed or is overstaked to chains
+		if node == nil ||
+			(isEnforceMaxChains && lenNodeChains > keeper.MaxChains(ctx)) ||
```
msmania: Use `sessionCtx` instead of using `ctx` as discussed in https://github.com/pokt-network/pocket-core/pull/1582#discussion_r1369336590 and have the param value as a local variable like `maxNodeChains` outside the loop.  It's ok to use `ctx` for `IsAfterEnforceMaxChainsUpgrade`.

msmania:
> Added new tests but CI fails for another reason, I have no idea why this error occurs and its fully out of scope of this PR to attend to it. Expanding the timeout fixed it previously but this behaviour is strange and should not occur. Besides that @Olshansk @msmania this PR is ready to go

Hmm, what failed is `TestRPC_QueryUnconfirmedTxs`, where I think one transaction was processed before the test ran `UnconfirmedTxs`, which has nothing to do with this change anyway.  I'm fine with ignoring irrelevant test failures.

msmania (changes requested):

msmania (approved):

Olshansk (approved):

POKT-Discourse:

PR #1585, opened 2023-10-24
Allow an AppMsgStake tx to change the address of a staked app

msmania:
The patch implements [PIP-35](https://forum.pokt.network/t/pip-35-introduce-a-secure-way-to-transfer-a-staked-app-to-a-new-account/4806), allowing the existing AppMsgStake transaction to change the address of a staked app.  This enables us to *transfer* the existing app slot from one to a new account without unstaking.  To make this operation easier, this patch also introduces a new command `app transfer`.

### Summary generated by Reviewpad on 15 Dec 23 11:53 UTC
This pull request includes changes in multiple files. Here is a summary of the changes:

1. The file `appStateChanges_test.go` has changes related to application state changes and transfers. The changes include the addition of new import statements, test functions, and helper functions.

2. The file `types.go` has changes related to the `MsgStake` type. The changes involve adding a new method `IsValidTransfer` to check for a special case in the `MsgStake` type where ownership transfer is being done.

3. The file `codec.go` has changes related to the support of an "App Transfer" feature. The changes include adding a constant and a new function for checking if a certain upgrade has occurred.

4. The file `expectedKeepers.go` has changes related to the addition of a new interface `AppKeeper`.

5. The file `keeper.go` has changes that involve adding a new field to the `Keeper` struct.

6. The file `baseapp.go` has changes that include formatting changes in the package comment and modifications in the `DeliverTx` function.

7. The file `app.go` has changes related to the assignment of a field in the `app.accountKeeper` object.

8. The file `auth.go` has changes that include adding comments and conditions for non-custodial and output address editor upgrades.

9. The file `common_test.go` has changes that involve importing packages, renaming a package, and adding/modifying functions.

10. The file `appStateChanges.go` has changes related to the validation and transfer functionality of applications.

11. The file `txUtil.go` has changes that add a new function for transferring an application.

12. The file `app/cmd/cli/app.go` has changes related to the addition of a new command for transferring the ownership of a staked app.

13. The file `keeper.go` has changes that add a new method for checking if a message is for transferring ownership.

14. The file `handler.go` has changes that include import statements, function parameter modifications, and logic for transferring application ownership.

These are the summaries of the changes in each file. Let me know if you have any specific questions or need further information regarding these changes.

### DIFF
--- a/app/cmd/cli/app.go
+++ b/app/cmd/cli/app.go
@@ -121,7 +121,7 @@ var appTransferCmd = &cobra.Command{
 	Use:   "transfer <fromAddr> <newAppPubKey> <networkID> <fee> [memo]",
 	Short: "Transfer the ownership of a staked app from one to another",
 	Long: `Submits a transaction to transfer the ownership of a staked app from
-<fromAddr> to a neww account specified as <newAppPubKey> without unstaking
+<fromAddr> to a new account specified as <newAppPubKey> without unstaking
 any app.  In other words, this edits the address of a staked app.  To run this
 command, you must have the private key of the current staked app <fromAddr>
 `,
--- a/app/cmd/cli/app.go
+++ b/app/cmd/cli/app.go
@@ -19,6 +19,7 @@ func init() {
 	rootCmd.AddCommand(appCmd)
 	appCmd.AddCommand(appStakeCmd)
 	appCmd.AddCommand(appUnstakeCmd)
+	appCmd.AddCommand(appTransferCmd)
 	appCmd.AddCommand(createAATCmd)
 }
 
@@ -112,6 +113,65 @@ Prompts the user for the <fromAddr> account passphrase.`,
 			fmt.Println(err)
 			return
 		}
+		fmt.Println(resp)
+	},
+}
+
+var appTransferCmd = &cobra.Command{
+	Use:   "transfer <fromAddr> <newAppPubKey> <networkID> <fee> [memo]",
+	Short: "Transfer the ownership of a staked app from one to another",
+	Long: `Submits a transaction to transfer the ownership of a staked app from
+<fromAddr> to a neww account specified as <newAppPubKey> without unstaking
+any app.  In other words, this edits the address of a staked app.  To run this
+command, you must have the private key of the current staked app <fromAddr>
+`,
+	Args: cobra.MinimumNArgs(4),
+	Run: func(cmd *cobra.Command, args []string) {
+		app.InitConfig(datadir, tmNode, persistentPeers, seeds, remoteCLIURL)
+
+		currentAppAddr := args[0]
+		newAppPubKey := args[1]
+		networkId := args[2]
+		feeStr := args[3]
+		memo := ""
+		if len(args) >= 5 {
+			memo = args[4]
+		}
+
+		fee, err := strconv.ParseInt(feeStr, 10, 64)
+		if err != nil {
+			fmt.Println("Invalid fee:", err)
+			return
+		}
+
+		fmt.Printf("Enter passphrase to unlock %s: ", currentAppAddr)
+		passphrase := app.Credentials(pwd)
+
+		rawTx, err := TransferApp(
+			currentAppAddr,
+			newAppPubKey,
+			passphrase,
+			networkId,
+			fee,
+			memo,
+		)
+		if err != nil {
+			fmt.Println("Failed to build a transaction:", err)
+			return
+		}
+
+		rawTxBytes, err := json.Marshal(rawTx)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+
+		resp, err := QueryRPC(SendRawTxPath, rawTxBytes)
+		if err != nil {
+			fmt.Println("Failed to submit a transaction:", err)
+			return
+		}
+
 		fmt.Println(resp)
 	},
 }
--- a/baseapp/baseapp.go
+++ b/baseapp/baseapp.go
@@ -1,10 +1,10 @@
-///*
-//Package baseapp contains data structures that provide basic data storage
-//functionality and act as a bridge between the ABCI interface and the SDK
-//abstractions.
+// /*
+// Package baseapp contains data structures that provide basic data storage
+// functionality and act as a bridge between the ABCI interface and the SDK
+// abstractions.
 //
-//BaseApp has no state except the CommitMultiStore you provide upon init.
-//*/
+// BaseApp has no state except the CommitMultiStore you provide upon init.
+// */
 package baseapp
 
 import (
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"github.com/pokt-network/pocket-core/codec/types"
 	"github.com/pokt-network/pocket-core/crypto"
-	types2 "github.com/pokt-network/pocket-core/x/apps/types"
 	"github.com/pokt-network/pocket-core/x/auth"
 	"github.com/tendermint/tendermint/evidence"
 	"github.com/tendermint/tendermint/node"
@@ -810,9 +809,8 @@ func (app *BaseApp) DeliverTx(req abci.RequestDeliverTx) (res abci.ResponseDeliv
 		msg := tx.GetMsg()
 		messageType = msg.Type()
 		recipient = msg.GetRecipient()
-		if signerPK == nil || messageType == types2.MsgAppStakeName {
-			signers := msg.GetSigners()
-			if len(signers) >= 1 {
+		if signerPK == nil {
+			if signers := msg.GetSigners(); len(signers) >= 1 {
 				signer = signers[0]
 			}
 		} else {
--- a/codec/codec.go
+++ b/codec/codec.go
@@ -58,6 +58,7 @@ const (
 	OutputAddressEditKey         = "OEDIT"
 	ClearUnjailedValSessionKey   = "CRVAL"
 	PerChainRTTM                 = "PerChainRTTM"
+	AppTransferKey               = "AppTransfer"
 )
 
 func GetCodecUpgradeHeight() int64 {
@@ -283,6 +284,12 @@ func (cdc *Codec) IsAfterOutputAddressEditorUpgrade(height int64) bool {
 func (cdc *Codec) IsAfterPerChainRTTMUpgrade(height int64) bool {
 	return (UpgradeFeatureMap[PerChainRTTM] != 0 &&
 		height >= UpgradeFeatureMap[PerChainRTTM]) ||
+		TestMode <= -3
+}
+
+func (cdc *Codec) IsAfterAppTransferUpgrade(height int64) bool {
+	return (UpgradeFeatureMap[AppTransferKey] != 0 &&
+		height >= UpgradeFeatureMap[AppTransferKey]) ||
 		TestMode <= -3
 }
 
--- a/x/apps/handler.go
+++ b/x/apps/handler.go
@@ -2,15 +2,16 @@ package pos
 
 import (
 	"fmt"
+	"reflect"
+
 	"github.com/pokt-network/pocket-core/crypto"
 	sdk "github.com/pokt-network/pocket-core/types"
 	"github.com/pokt-network/pocket-core/x/apps/keeper"
 	"github.com/pokt-network/pocket-core/x/apps/types"
-	"reflect"
 )
 
 func NewHandler(k keeper.Keeper) sdk.Handler {
-	return func(ctx sdk.Ctx, msg sdk.Msg, _ crypto.PublicKey) sdk.Result {
+	return func(ctx sdk.Ctx, msg sdk.Msg, signer crypto.PublicKey) sdk.Result {
 		ctx = ctx.WithEventManager(sdk.NewEventManager())
 		// convert to value for switch consistency
 		if reflect.ValueOf(msg).Kind() == reflect.Ptr {
@@ -18,7 +19,7 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
 		}
 		switch msg := msg.(type) {
 		case types.MsgStake:
-			return handleStake(ctx, msg, k)
+			return handleStake(ctx, msg, signer, k)
 		case types.MsgBeginUnstake:
 			return handleMsgBeginUnstake(ctx, msg, k)
 		case types.MsgUnjail:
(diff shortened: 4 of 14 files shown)


POKT-Discourse:

https://forum.pokt.network/t/pip-35-introduce-a-secure-way-to-transfer-a-staked-app-to-a-new-account/4806/1

POKT-Discourse:

https://forum.pokt.network/t/era-cycle-report-cycle-2/4828/1

app/cmd/cli/app.go
```diff
@@ -116,6 +117,65 @@ Prompts the user for the <fromAddr> account passphrase.`,
 	},
 }
 
+var appTransferCmd = &cobra.Command{
+	Use:   "transfer <fromAddr> <newAppPubKey> <networkID> <fee> [memo]",
+	Short: "Transfer the ownership of a staked app from one to another",
+	Long: `Submits a transaction to transfer the ownership of a staked app from
+<fromAddr> to a neww account specified as <newAppPubKey> without unstaking
```
Olshansk:
```suggestion
<fromAddr> to a new account specified as <newAppPubKey> without unstaking
```

baseapp/baseapp.go
```diff
@@ -810,9 +809,8 @@ func (app *BaseApp) DeliverTx(req abci.RequestDeliverTx) (res abci.ResponseDeliv
 		msg := tx.GetMsg()
 		messageType = msg.Type()
 		recipient = msg.GetRecipient()
-		if signerPK == nil || messageType == types2.MsgAppStakeName {
-			signers := msg.GetSigners()
-			if len(signers) >= 1 {
+		if signerPK == nil {
```
Olshansk, line 812: Do you know why we used to have `messageType == types2.MsgAppStakeName`?

codec/codec.go
```diff
@@ -57,6 +57,7 @@ const (
 	VEDITKey                     = "VEDIT"
 	OutputAddressEditKey         = "OEDIT"
 	ClearUnjailedValSessionKey   = "CRVAL"
+	AppTransferKey               = "AppTransfer"
```
Olshansk, line 61: NIT: s/AppTransfer/APPTR?

x/apps/handler.go
```diff
@@ -30,25 +31,38 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
-	err := k.StakeApplication(ctx, application, msg.Value)
-	if err != nil {
-		return err.Result()
+	// check if the msg is to transfer an app first
+	if curApp, err := k.ValidateApplicationTransfer(ctx, signer, msg); err == nil {
+		if err := k.TransferApplication(ctx, curApp, msg.PubKey); err != nil {
+			return err.Result()
+		}
+	} else {
+		// then check if they can stake
```
Olshansk:
```suggestion
		// otherwise check if the msg is to stake an application
```

x/apps/handler.go
```diff
@@ -30,25 +31,38 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
+		if err := k.ValidateApplicationStaking(ctx, application, msg.Value); err != nil {
+			ctx.Logger().Error(fmt.Sprintf("Validate App Can Stake Error, at height: %d with address: %s", ctx.BlockHeight(), sdk.Address(addr).String()))
```
Olshansk, line 57:
```suggestion
			ctx.Logger().Error(fmt.Sprintf("ERROR: ValidateApplicationStaking at height: %d with address: %s", ctx.BlockHeight(), sdk.Address(addr).String()))
```

x/apps/handler.go
```diff
@@ -30,25 +31,38 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
+		}
+
+		ctx.Logger().Info("Change App state to Staked " + sdk.Address(addr).String())
+		// change the application state to staked
+		if err := k.StakeApplication(ctx, application, msg.Value); err != nil {
+			return err.Result()
+		}
 	}
 	// create the event
+	senderAddrStr := sdk.Address(signer.Address()).String()
```
Olshansk: s/senderAddrStr/signedAddrStr

x/apps/handler.go
```diff
@@ -30,25 +31,38 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
-		return err.Result()
-	}
-	ctx.Logger().Info("Change App state to Staked " + sdk.Address(addr).String())
-	// change the application state to staked
-	err := k.StakeApplication(ctx, application, msg.Value)
-	if err != nil {
-		return err.Result()
+	// check if the msg is to transfer an app first
+	if curApp, err := k.ValidateApplicationTransfer(ctx, signer, msg); err == nil {
+		if err := k.TransferApplication(ctx, curApp, msg.PubKey); err != nil {
```
Olshansk: Can you add a log line about here similar to the one you have below for consistency?

x/apps/types/msg.go
```diff
@@ -62,6 +66,24 @@ func (msg MsgStake) ValidateBasic() sdk.Error {
...
+		return ErrNilApplicationAddr(DefaultCodespace)
+	}
+
+	// The stake amount must be zero for transfer
+	if !msg.Value.IsZero() {
+		return ErrBadStakeAmount(DefaultCodespace)
+	}
+
+	// The chains must be empty for transfer
+	if !msg.Value.IsZero() {
```
Olshansk: This is the same check as on line 75

x/apps/keeper/appUtil.go
```diff
@@ -30,3 +30,33 @@ func (k Keeper) AllApplications(ctx sdk.Ctx) (apps []exported.ApplicationI) {
 	}
 	return apps
 }
+
+func (k Keeper) IsMsgAppTransfer(
```
Olshansk, line 35: Please add a godoc comment

x/apps/types/msg.go
```diff
@@ -45,6 +45,10 @@ func (msg MsgStake) GetSignBytes() []byte {
 
 // ValidateBasic quick validity check for staking an application
 func (msg MsgStake) ValidateBasic() sdk.Error {
+	if err := msg.IsValidTransfer(); err == nil {
```
Olshansk, line 50: Can you add a comment explaining that we're reusing the `MsgStake` command for transfers

x/apps/keeper/appStateChanges.go
```diff
@@ -77,6 +101,40 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentApp types.Application, amo
...
 
+func (k Keeper) ValidateApplicationTransfer(
+	ctx sdk.Ctx,
+	signer crypto.PublicKey,
+	msg types.MsgStake,
+) (curApp types.Application, err sdk.Error) {
+	if !ctx.IsAfterUpgradeHeight() ||
+		!k.Cdc.IsAfterAppTransferUpgrade(ctx.BlockHeight()) {
+		err = types.ErrApplicationStatus(k.codespace)
+		return
```
Olshansk: NIT: Can we use explicit return types (not named). It's harder to follow this way.

x/apps/keeper/appStateChanges.go
```diff
@@ -77,6 +101,40 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentApp types.Application, amo
...
+	// The signer must be a staked app
+	var found bool
+	curApp, found = k.GetApplication(ctx, sdk.Address(signer.Address()))
+	if !found || !curApp.IsStaked() {
+		err = types.ErrApplicationStatus(k.codespace)
+		return
+	}
+
+	err = ensurePubKeyTypeSupported(ctx, msg.PubKey, k.Codespace())
+	if err != nil {
```
Olshansk: Single line

x/apps/keeper/appStateChanges.go
```diff
@@ -77,6 +101,40 @@ func (k Keeper) ValidateEditStake(ctx sdk.Ctx, currentApp types.Application, amo
...
+		return
+	}
+
+	// The pubkey in the message must not exist
```
Olshansk: We explicitly set the `PubKey` in `MsgStake` when doing a transfer like so:

```go
	msg := appsType.MsgStake{PubKey: newAppPubKey}
	if err = msg.ValidateBasic(); err != nil {
		return nil, err
	}
```

A bit confused by this comment.

Do you mean, the `The pubKey of msgStake must not be a pre-existing application`?

x/auth/ante.go
```diff
@@ -72,6 +72,17 @@ func ValidateTransaction(ctx sdk.Ctx, k Keeper, stdTx types.StdTx, params Params
...
 	}
 
+	if ctx.IsAfterUpgradeHeight() &&
+		k.Cdc.IsAfterAppTransferUpgrade(ctx.BlockHeight()) {
+		msgSigner := sdk.Address(stdTx.Signature.Address())
+		// AppMsgStake has a special case of AppTransfer.  In that case, we accept
+		// a message where the pubkey is different from the signer, delegating
+		// most of validation work to the app's message handler after fee deduction.
+		if k.AppKeeper.IsMsgAppTransfer(ctx, msgSigner, stdTx.Msg) {
+			validSigners = append(validSigners, msgSigner)
```
Olshansk, line 88: Is `msgSigner` not part of `validSigners` by default?

x/apps/keeper/appStateChanges_test.go
```diff
@@ -371,3 +374,137 @@ func TestAppStateChange_BeginUnstakingApplication(t *testing.T) {
...
+	t.Cleanup(func() {
+		codec.UpgradeHeight = originalUpgradeHeight
+		codec.TestMode = originalTestMode
+		codec.UpgradeFeatureMap[codec.AppTransferKey] = originalFeatKey
+	})
+	codec.UpgradeHeight = -1
+
+	ctx, _, keeper := createTestInput(t, true)
+
+	apps := make([]types.Application, 3)
```
Olshansk, line 428:
```suggestion
	// Create and stake 3 apps
	apps := make([]types.Application, 3)
```

x/apps/keeper/appStateChanges_test.go
```diff
@@ -371,3 +374,137 @@ func TestAppStateChange_BeginUnstakingApplication(t *testing.T) {
...
+		amount := sdk.NewInt(int64(i) * 10000000)
+		fundAccount(t, &ctx, &keeper, apps[i].Address, amount)
+		assert.Nil(t, keeper.StakeApplication(ctx, apps[i], amount))
+	}
+	pubKeys[3] = getRandomPubKey()
+	addrs[3] = sdk.Address(pubKeys[3].Address())
+
+	// apps[0]: staked
+	// apps[1]: unstaking
+	// apps[2]: staked and jailed
```
Olshansk, line 444:
```suggestion
	// apps[2]: staked and jailed
	// apps[3]: not an application (see pubKeys[3]) and will be used for the transfer
```

x/apps/keeper/appStateChanges_test.go
```diff
@@ -371,3 +374,137 @@ func TestAppStateChange_BeginUnstakingApplication(t *testing.T) {
...
+	err = transferApp(t, &ctx, &keeper, pubKeys[0], pubKeys[3])
+	assert.Nil(t, err)
+
+	// transfer: apps[3]-->apps[2](unstaking) - fail
+	err = transferApp(t, &ctx, &keeper, pubKeys[3], pubKeys[2])
+	assert.NotNil(t, err)
+	assert.Equal(t, keeper.Codespace(), err.Codespace())
+	assert.Equal(t, types.CodeInvalidStatus, err.Code())
+
+	// transfer: apps[3]-->apps[0](previous owner) success
```
Olshansk, line 471: 1. Why do we expect this to succeed?
2. If we were to wait for a new block, would this fail since the previous app is never unstaked?

x/apps/keeper/appStateChanges_test.go
```diff
@@ -371,3 +374,137 @@ func TestAppStateChange_BeginUnstakingApplication(t *testing.T) {
...
+	stakedApp, found = keeper.GetApplication(ctx, addrs[1])
+	assert.True(t, found)
+	assert.True(t, stakedApp.IsUnstaking())
+	stakedApp, found = keeper.GetApplication(ctx, addrs[2])
+	assert.True(t, found)
+	assert.True(t, stakedApp.IsStaked())
+	assert.True(t, stakedApp.IsJailed())
+
+	// transfer a jailed app - success
+	err = transferApp(t, &ctx, &keeper, pubKeys[2], pubKeys[3])
```
Olshansk, line 505: pubKeys[3] was already transfered to above. If we were to wait for a new block, would this fail?

Olshansk (changes requested):
Nothing blocking, but just a couple nits and questions.

baseapp/baseapp.go
msmania: Unfortunately no.  This was introduced by the large, combined commit  32157db43c88fcd37d5c68978f0256c8631db833 of v0.8, but this part does nothing meaningful.  MsgAppStake's `GetSigners` always returns a single element array and its value is `msg.PubKey` which was guaranteed to be the signer, so it doesn't make any sense to have this.

codec/codec.go
msmania: This may be bikeshedding.  Do we need to keep feature names cryptic codenames like those?  In my understanding, there is no limitation in case sensitivity or the length.  I think having more descriptive names makes the result of `/v1/query/upgrade` easier to understand.

x/apps/handler.go
msmania: This error log is not new.  Let's leave it as is, though it has strange capitalization.

x/apps/handler.go
msmania: changed it to `signerAddrStr`

x/apps/types/msg.go
msmania: Good catch. Changed it to `len(msg.Chains) > 0`.

x/apps/keeper/appStateChanges.go
msmania: Yes, you're right.  What I meant was "not exist in the world state".  Applying your suggestion.

x/apps/handler.go
msmania: I realized `TransferApplication` never fails.  Removed the return value and moved info logging from `TransferApplication` to this function.

x/apps/keeper/appStateChanges_test.go
msmania: `transferApp` directly modifies the world state.  There is no block generation.  This means at line 468, the app has been transferred to `apps[3]` and `apps[0]` is not a staked app.  That's why we can transfer the ownership back to `apps[0]`.

x/apps/keeper/appStateChanges_test.go
msmania: Same here.  At line 502, the app ownership has been transferred back to `apps[0]`.  `apps[3]` is not a staked app.

x/auth/ante.go
msmania: Counterintuitively it's not.  `stdTx.GetSigners()` is simply redirected to `tx.GetMsg().GetSigners()`, not looking at the actual signer.  This is why every transaction message has a redundant field storing its signer.

x/apps/keeper/appStateChanges_test.go
Olshansk: Can you update the test to verify that IF we do wait for a block, the transfer back fails?

x/auth/ante.go
Olshansk: Can you #PUC (#PleaseUpdateComment) with this additional detail & context?

x/apps/types/msg.go
Olshansk: Can you add a unit test to catch this please?

codec/codec.go
Olshansk: Not adamant either way. Feel free to resolve.

Olshansk (changes requested):
Added a few followup comments to the previous review. PTAL

x/apps/keeper/appStateChanges_test.go
msmania: > Can you update the test to verify that IF we do wait for a block, the transfer back fails?

If a block was generated after the line 462, the test would still pass.  After the app ownership is transferred from A to B, the address A is the same as being unstaked or never being staked, meaning the address can be staked again.

baseapp/baseapp.go
Olshansk: Logically, I agree that you're right and I cannot prove otherwise. However, I'm really paranoid that there's some weird edge case and I don't want any of us to deal with a chain halt.

`Hypothesis`: When the message type is NOT an AppStake (e.g. delegated staking for suppliers / node providers), maybe we need to enforce to using the first signer?

Olshansk:
@msmania See the one linger comment I have [here](https://github.com/pokt-network/pocket-core/pull/1585#discussion_r1428569418).

--- 

Also, I tried to do a `Show changes since your last review` (see screenshots below) but I'm guessing there was a rebase/force-push/etc done to the branch which prevents that from happening.

I'm travelling on a trip this weekend and will aim to take a second look at this mid next week.

Screenshot 2023-12-15 at 3 18 11 PM

Screenshot 2023-12-15 at 3 19 06 PM

baseapp/baseapp.go
msmania: This patch does not change the logic in the case where the message type is not an AppStake.  Regardless of this patch, we enforce to using the first signer only if `signerPK` is null.  Moreover, this `signer` affects `abci.ResponseDeliverTx` which just goes to the transaction log.  I don't think it causes a chain halt.

Olshansk (approved):
Thanks @msmania!

POKT-Discourse:

PR #1588, opened 2023-12-21
Do not call `node.GetChains()` if node is nil

msmania:
In `NewSessionNodes`, the variable `node` can be `nil` if the node is unstaked between `sessionCtx` and `ctx`.  In such a case, `node.GetChains()` causes a panic.  This patch makes sure we don't call it if the `node` is `nil`.
This is a regression from #1582.

### Summary generated by Reviewpad on 21 Dec 23 10:23 UTC
This pull request fixes a regression introduced in #1582. The patch ensures that the function `node.GetChains()` is not called if the `node` is `nil`, preventing a panic in the `NewSessionNodes` function.

### DIFF
--- a/x/pocketcore/types/session.go
+++ b/x/pocketcore/types/session.go
@@ -150,10 +150,9 @@ func NewSessionNodes(
 
 		// cross check the node from the `new` or `end` world state
 		node = keeper.Validator(ctx, n)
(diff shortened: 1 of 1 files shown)


h5law (approved):
Good catch LGTM

Olshansk (approved):

PR #1591, opened 2024-01-14
Sort RewardDelegators when marshaling a validator

msmania:
## Description

This is a fix for a regression introduced by #1581.

When a node stores a validator with the new `RewardDelegators` field, its marshaled product is not deterministic because the order of map iteration is not deterministic.  As a result, the merkle tree of application.db can differ between nodes, leading to consensus failure.

The proposed fix is to sort the keys of `RewardDelegators` by using the stable marshaler.

### Summary generated by Reviewpad on 15 Jan 24 18:21 UTC
This pull request includes changes to the `nodes.proto` file and a file diff.

The changes in the `nodes.proto` file include:
- Added an option `stable_marshaler` with a value of `true`. This option is used to make the order of `RewardDelegators` deterministic.
- Updated the `Address` field to include additional options `(gogoproto.casttype)`, `(gogoproto.moretags)`, and `(gogoproto.jsontag)`.
- Updated the `PublicKey` field to include additional options `(gogoproto.moretags)` and `(gogoproto.jsontag)`.
- Updated the `jailed` field to include an additional option `(gogoproto.jsontag)`.

The file diff includes the following changes:
- Imported the "encoding/hex", "math/rand", "reflect", "strconv", and "time" packages.
- Added a test case for marshalling RewardDelegators.
- Added the "Test_Marshal_RewardDelegators" function.
- Created variables and constants within the test function.
- Generated random numbers to be used as reward delegator addresses.
- Created and initialized a map, "delegatorMap", with random numbers as keys and their corresponding modulo 10 values as values.
- Created a "Validator" struct with various fields and assigned values to them.
- Marshaled the validator struct and ensured that the resulting hash is the same for every iteration of the test.

Please let me know if there is anything specific you would like me to focus on.


Olshansk:
@msmania Friendly reminder to update ticket metadata
Screenshot 2024-01-15 at 11 22 56 AM

x/nodes/keeper/validator_test.go
```diff
@@ -150,6 +155,60 @@ func Test_sortNoLongerStakedValidators(t *testing.T) {
...
+	numOfDelegators := 100
+
+	chains := []string{"0001", "0002", "005A", "005B", "005C", "005D"}
+	url := "https://test.pokt.network:443"
+	stake := int64(18000000000)
+	pubKey, _ := crypto.NewPublicKey("0b787e54e66b3db3a3396c2322d8314287e08990f7696c490153b078bada7e94")
+	nodeAddr := sdk.Address(pubKey.PubKey().Address())
+	outputAddr, _ := sdk.AddressFromHex("42846261e1798fc08e1dfd97325af7b280f815b0")
+
+	randNums := make([]int, numOfDelegators)
```
Olshansk, line 173: Can you add a comment on what `randNums` is intended to be used for?

x/nodes/keeper/validator_test.go
```diff
@@ -150,6 +155,60 @@ func Test_sortNoLongerStakedValidators(t *testing.T) {
...
+	for j := 0; j < numOfDelegators; j++ {
+		randNums[j] = rand.Int()
+	}
+
+	var valHash string
+
+	for i := 0; i < numOfIterations; i++ {
+		rand.Shuffle(len(randNums), func(i, j int) {
```
Olshansk, line 189: #PUC (#PleaseUpdateComment) on why we are shuffling.

Context: Note that the reason I ask fo these things isn't because I can't figure it out myself, but because I take time to think through it to understand. When the author adds a comment while implementing it, it saves time for many devs (and reviewers) many times over.

x/nodes/keeper/validator_test.go
```diff
@@ -150,6 +155,60 @@ func Test_sortNoLongerStakedValidators(t *testing.T) {
...
+		}
+		valBytes, err := val.Marshal()
+		assert.Nil(t, err)
+
+		// Make sure the hash is always the same regardless of the order of randNums
+		if len(valHash) == 0 {
+			valHash = hex.EncodeToString(valBytes)
+		} else {
+			assert.Equal(t, hex.EncodeToString(valBytes), valHash)
+		}
```
Olshansk: Why is it not something like this?

```suggestion
		// First test iteration when valhash hasn't been set yet
		if len(valHash) == 0 {
			valHash = hex.EncodeToString(valBytes)
		}

		// Make sure the hash is always the same regardless of the order of randNums
		assert.Equal(t, hex.EncodeToString(valBytes), valHash)
```

x/nodes/types/nodes.pb.go
```diff
@@ -445,7 +446,14 @@ func (m *ProtoValidator) MarshalToSizedBuffer(dAtA []byte) (int, error) {
 	var l int
 	_ = l
 	if len(m.RewardDelegators) > 0 {
+		keys := make([]string, len(m.RewardDelegators))
```
Olshansk: Strong `no` towards this approach. This is an auto-generated proto file that has the following comment at the top.

```go
// Code generated by protoc-gen-gogo. DO NOT EDIT.
// source: x/nodes/nodes.proto
```

The way to go around this is to use add functionality to the (auto-generated) receiver in the same directory where it was created.

Lmk if you need help with this, but I would look at `*MsgProtoStake)` as an example:

Screenshot 2024-01-15 at 11 40 21 AM

---

Context: discussing whether we should commit `.pb.go` files to `main` is outside the scope of this discussion. Let's just keep that legacy decision around for now.

Olshansk (changes requested):

x/nodes/types/nodes.pb.go
msmania: I found the best solution for this.  Stay tuned..

x/nodes/keeper/validator_test.go
msmania: Good point

x/nodes/keeper/validator_test.go
msmania: Sure.  Added a comment about this.

msmania:
> @msmania Friendly reminder to update ticket metadata

So the process is to set a project AND fill the Release field in it?

x/nodes/types/nodes.pb.go
msmania: So I found [the stable marshaler](https://pkg.go.dev/github.com/gogo/protobuf/plugin/marshalto).  This does almost the same as the previous change except it adopts the descending order, which means the current world state of testnet is invalid.  We need to roll back the state and redo the test again...

Olshansk (approved):

PR #1598, opened 2024-02-07
[Docs] AAT related comments & documentation improvements

Olshansk:
This PR is the result of reviewing the codebase to answer AAT-related questions
for the backend & infra teams at grove. As a result, the following changes were made:

- Reformat (automatically) a handful of files using [goimports-reviser](https://github.com/incu6us/goimports-reviser)
- Update comments in some parts of the source code related to AATs
- Update the documentation related to generating and using AATs
- Build a reference upon which the foundation will update the official documentation at [pokt-network/docs-v2](https://github.com/pokt-network/docs-v2)

General quality-of-life improvements for documentation, comments and code health.


app/util.go
```diff
@@ -15,8 +15,12 @@ import (
 	pocketKeeper "github.com/pokt-network/pocket-core/x/pocketcore/keeper"
 )
 
-func GenerateAAT(appPubKey, clientPubKey string, key crypto.PrivateKey) (aatjson []byte, err error) {
-	aat, er := pocketKeeper.AATGeneration(appPubKey, clientPubKey, key)
+// GenerateAAT generates an AAT to be used for relay servicing.
+// - appPubKey is the public key of the application that's paying for on-chain service.
```
adshmh: This comment is very helpful: a link to a grove doc page that describes the relationship between apps, nodes, and relays, would be great.

doc/specs/application-auth-token.md
```diff
@@ -20,26 +35,79 @@ A semver string specifying the spec version under which this AAT needs to be int
 
 > type: `string`
 
-The application will sign a hash of the `message` property within this token with the specified `appPubKey` and
-corresponding private key.
+The application will sign a hash of the `message` property within this token with
```
adshmh: I find the `message` reference confusing: rephrasing this to say the application will sign a hash of all the other fields of the AAT (or even naming all the fields that are part of the signed bytes slice) would be easier to read.

There is a single reference to `message` in a diagram on this page, and that is SHA256 hash of the encoded AAT.

doc/specs/application-auth-token.md
```diff
@@ -20,26 +35,79 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+The hexadecimal public key allowing granular control of who can use the `AAT`.
 
 ## ECDSA ed25519 Signature Scheme
 
-The protocol wide ed25519 ECDSA will be used for any signatures and verifications that are used within this
-specification.
+The protocol wide ed25519 ECDSA will be used for any signatures and verifications
+that are used within this specification.
+
+## AAT Uses-Cases
```
adshmh: I believe AAT has a single use, i.e. allow Gateways access Service nodes on behalf of the application. `Generating AAT` seems like a better title here, with the two cases of `Application === Client` and `Application !== Client` as the two different ways to generate an AAT. The current `AAT Generation` section title can be then merged with these two cases to capture the common steps between them.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +116,45 @@ The proper way to sign the token is as follows:
...
     Version:              a.Version,
-}
+  }
+  AATBytes = JSON.Encode(AAT)
+  Message = SHA3_256(AATBytes)
+  AAT.ApplicationSignature = ED25519.Sign(Message)
 ```
 
-`AATBytes = JSON.Encode(AAT)`
+### Relay Generation
```
adshmh: Perhaps `Using an AAT to generate a Relay` is a better title.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +116,45 @@ The proper way to sign the token is as follows:
...
 
-`Message = SHA3-256(AATBytes)`
+The `Client` is needed to sign the actual relays while the `Application` gets
+charged on-chain. However, the `Application` only gets charged if the `Client`
+signed the relay. The `Client` public key is in the `AAT` and had to be signed
+by the `Application` during `AAT` generation.
```
adshmh: An extra explanation here, e.g. `by signing the client's public key, the application agrees to be charged for relays sent by the client`, would be very helpful.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +116,45 @@ The proper way to sign the token is as follows:
...
+    subgraph RPU["Unsigned RelayProof"]
+        requestHash
+        Entropy
+        SessionHeight
+        AAT
+        ServicerPubKey
+        BlockchainID
+    end
+
+    AAT <-.-> CG
```
adshmh: This link should be removed from this diagram IMO: it makes the diagram harder to understand as it looks to create a cycle. AAT generation is presented a bit earlier on this page.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +116,45 @@ The proper way to sign the token is as follows:
...
+    CG  <-.-> sig
+
+    RPU -- hash --> RPUB["RelayProof Bytes"]
+
+    RPUB -- "Sign with Client/Gateway\nPrivate Key" --> sig["Client/Gateway Signature"]
```
adshmh: This arrow and the `Signature` arrow are a bit confusing: both `ClientPrivateKey/GatePrivateKey` and `RelayProofBytes` are input to the Gateway, which generates the signature and adds it to the `Relay Proof` struct.

x/pocketcore/keeper/aat.go
```diff
@@ -8,23 +8,29 @@ import (
...
+	aatBytes := aat.Hash()
+
+	// sign the AAT
+	sig, err := appPrivKey.Sign(aatBytes)
 	if err != nil {
 		return pc.AAT{}, pc.NewSignatureError(pc.ModuleName, err)
 	}
-	// stringify the signature into hex
+
+	// This is where the `ApplicationPrivKey` signs (i.e. delegates trust) to
```
adshmh: This comment is very helpful, but a better position for it would be at line 26, i.e. just before `sign, err := appPrivKey.Sign(aatBytes)`

adshmh (changes requested):
Thank you for sharing the PR, it is very helpful in showing how AATs work.
One extra suggestion: would be great if we could reference several of the files that are touched here in a list of docs/code to look at to learn how the relays and proofs work.

commoddity (approved):
Agree with @adshmh here; this is great improvement in clarity and providing context. Thanks a lot.

app/util.go
Olshansk: > a link to a grove doc page that describes the relationship between apps, nodes, and relays, would be great.

I'm unsure what you mean here.

1. Should we link to this file from within the grove documentation?
2. Should we link to a grove doc in the comment?
3. Something else?

Olshansk:
@adshmh 

> One extra suggestion: would be great if we could reference several of the files that are touched here in a list of docs/code to look at to learn how the relays and proofs work.

Would you be open to tackle this one yourself?

app/cmd/cli/app.go
```diff
@@ -9,10 +9,11 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/spf13/cobra"
+
```
msmania, line 13: remove this empty line

app/cmd/cli/app.go
```diff
@@ -179,10 +180,19 @@ command, you must have the private key of the current staked app <fromAddr>
 var createAATCmd = &cobra.Command{
 	Use:   "create-aat <appAddr> <clientPubKey>",
 	Short: "Creates an application authentication token",
-	Long: `Creates a signed application authentication token (version 0.0.1 of the AAT spec), that can be embedded into application software for Relay servicing.
-Will prompt the user for the <appAddr> account passphrase.
-Read the Application Authentication Token documentation for more information.
-NOTE: USE THIS METHOD AT YOUR OWN RISK. READ THE APPLICATION SECURITY GUIDELINES IN ORDER TO UNDERSTAND WHAT'S THE RECOMMENDED AAT CONFIGURATION FOR YOUR APPLICATION.`,
+	Long: `Creates a signed Application Authentication Token.
+
+WARNING: USE THIS METHOD AT YOUR OWN RISK.
```
msmania: Let's remove this WARNING message.  Any pocket command is at a user's risk.  Moreover, this command is an offline command, just to print AAT.  I believe the risk of this command is much lower than other commands to submit a transaction.

app/cmd/rpc/client.go
```diff
@@ -5,15 +5,16 @@ import (
...
 	"fmt"
-	nodesTypes "github.com/pokt-network/pocket-core/x/nodes/types"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/julienschmidt/httprouter"
+
```
msmania, line 15: remove this empty line

app/cmd/rpc/client.go
```diff
@@ -65,6 +66,7 @@ func Relay(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
 		WriteJSONResponseWithCode(w, string(j), r.URL.Path, r.Host, 400)
 		return
 	}
+	// Handling the relay
```
msmania: Code is self-explanatory.  Remove this comment.

app/query.go
```diff
@@ -605,6 +606,7 @@ func (app PocketCoreApp) HandleRelay(r pocketTypes.Relay) (res *pocketTypes.Rela
 		return nil, nil, fmt.Errorf("pocket node is currently syncing to the blockchain, cannot service in this state")
 	}
 
+	// Calling HandleRelay here
```
msmania: Code is self-explanatory.  Remove this comment.

app/cmd/cli/app.go
```diff
@@ -203,15 +213,24 @@ NOTE: USE THIS METHOD AT YOUR OWN RISK. READ THE APPLICATION SECURITY GUIDELINES
...
-		aat, err := app.GenerateAAT(hex.EncodeToString(kp.PublicKey.RawBytes()), args[1], privkey)
+
+		// Generate the AAT
+		aat, err := app.GenerateAAT(pubKeyHexEncoded, args[1], privKey)
 		if err != nil {
 			fmt.Println(err)
 			return
 		}
+
+		// Print out the AAT
```
msmania: Code is self-explanatory.  Remove this comment.

app/cmd/cli/app.go
```diff
@@ -203,15 +213,24 @@ NOTE: USE THIS METHOD AT YOUR OWN RISK. READ THE APPLICATION SECURITY GUIDELINES
...
+		pubKey := kp.PublicKey.RawBytes()
+		pubKeyHexEncoded := hex.EncodeToString(pubKey)
+		privKeyArmored := kp.PrivKeyArmor
+		privKey, err := mintkey.UnarmorDecryptPrivKey(privKeyArmored, cred)
 		if err != nil {
 			return
 		}
-		aat, err := app.GenerateAAT(hex.EncodeToString(kp.PublicKey.RawBytes()), args[1], privkey)
+
+		// Generate the AAT
```
msmania: Code is self-explanatory.  Remove this comment.

app/cmd/rpc/client.go
```diff
@@ -5,15 +5,16 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	nodesTypes "github.com/pokt-network/pocket-core/x/nodes/types"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
 	"time"
```
msmania, line 13: remove this empty line

app/query.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	core_types "github.com/tendermint/tendermint/rpc/core/types"
 	tmtypes "github.com/tendermint/tendermint/types"
```
msmania, line 14: remove this empty line

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
 
 var _ Proof = RelayProof{} // ensure implements interface at compile time
 
-// "ValidateLocal" - Validates the proof object, where the owner of the proof is the local node
-func (rp RelayProof) ValidateLocal(appSupportedBlockchains []string, sessionNodeCount int, sessionBlockHeight int64, verifyAddr sdk.Address) sdk.Error {
-	//Basic Validations
+// ValidateLocal validates the Relay Proof object, where the owner of the proof is the local node
+
```
msmania: Remove empty lines between a function and its leading comment.

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
 
 var _ Proof = RelayProof{} // ensure implements interface at compile time
 
-// "ValidateLocal" - Validates the proof object, where the owner of the proof is the local node
-func (rp RelayProof) ValidateLocal(appSupportedBlockchains []string, sessionNodeCount int, sessionBlockHeight int64, verifyAddr sdk.Address) sdk.Error {
-	//Basic Validations
+// ValidateLocal validates the Relay Proof object, where the owner of the proof is the local node
```
msmania: This comment does not describe the function correctly.  This is a good time to correct it.

I think the reason behind "where the owner of the proof is the local node" is `Relay.Validate` calls this function with `servicerAddr`.

How about this?

```go
// ValidateLocal validates the Relay Proof object is aligned with
// `sessionBlockHeight` and `verifyAddr`.
```

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
+
+func (rp RelayProof) ValidateLocal(
+	appSupportedBlockchains []string,
+	sessionNodeCount int,
```
msmania: `sessionNodeCount` is never used by this function or `RelayProof.Validate`.  Can be renamed to `_` or removed.

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
+	sessionBlockHeight int64,
+	verifyAddr sdk.Address,
```
msmania: How about renaming `verifyAddr` to `expectedServicerAddr`?

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
+) sdk.Error {
+	// Basic validation of the relay proof
```
msmania: remove this comment

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
 	err := rp.ValidateBasic()
 	if err != nil {
 		return err
 	}
+	// Retrieve the servicer public key
```
msmania: remove this comment

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
 	if er != nil {
 		return NewInvalidNodePubKeyError(ModuleName)
 	}
 	// validate the public key correctness
 	if !sdk.Address(servicerPublicKey.Address()).Equals(verifyAddr) {
-		return NewInvalidNodePubKeyError(ModuleName) // the public key is not this nodes, so they would not get paid
+		// the public key is not this nodes, so they would not get paid
+		return NewInvalidNodePubKeyError(ModuleName)
 	}
+	// ValidateLocal calling `rp.Validate` is important
```
msmania: Remove this comment..  Every function call is important :)

x/pocketcore/types/proof.go
```diff
@@ -54,21 +55,30 @@ func (ps ProofIs) FromProofI() (res Proofs) {
...
 	}
+	// Retrieve the servicer public key
 	servicerPublicKey, er := crypto.NewPublicKey(rp.ServicerPubKey)
 	if er != nil {
 		return NewInvalidNodePubKeyError(ModuleName)
 	}
 	// validate the public key correctness
 	if !sdk.Address(servicerPublicKey.Address()).Equals(verifyAddr) {
-		return NewInvalidNodePubKeyError(ModuleName) // the public key is not this nodes, so they would not get paid
+		// the public key is not this nodes, so they would not get paid
```
msmania: "is not this nodes" is not correct, and whether get paid or not get paid is out of this function's scope.  I think the meaning of this line is obvious (verification has failed because servicerPublicKey doesn't match) and no need to have a comment.

x/pocketcore/keeper/proof.go
```diff
@@ -4,18 +4,23 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"math"
+	"reflect"
+	"time"
+
+	"github.com/tendermint/tendermint/rpc/client"
+
```
msmania, line 12: remove this empty line

x/pocketcore/keeper/querier.go
```diff
@@ -2,10 +2,12 @@ package keeper
 
 import (
 	"fmt"
+
+	abci "github.com/tendermint/tendermint/abci/types"
+
```
msmania, line 7: remove this empty line

doc/specs/application-auth-token.md
```diff
@@ -20,26 +35,80 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+that are used within this specification.
+
+## AAT Generation
+
+When generating a new AAT, the owner of the `Application` private key has two options:
+
+1. Use the same public key for signing the AAT and the relay requests
+2. Use a different public key for signing the AAT and the relay requests
+
+### 1. Application === Client (Gateway)
```
msmania: I think one figure is enough to explain the flow.  The case of `Application === Client` is just a special case.

doc/specs/cli/apps.md
```diff
@@ -61,38 +61,33 @@ Transaction submitted with hash: <Transaction Hash>
 pocket apps create-aat <appAddr> <clientPubKey>
 ```
 
-Creates a signed application authentication token \(version `0.0.1` of the AAT spec\), that can be embedded into
-application software for Relay servicing. Will prompt the user for the `<appAddr>` account passphrase.
+**WARNING: USE THIS METHOD AT YOUR OWN RISK.**
```
msmania: We shouldn't have the same description in two places here and in app/cmd/cli/app.go.

x/apps/types/application_test.go
```diff
@@ -3,16 +3,18 @@ package types
...
-	sdk "github.com/pokt-network/pocket-core/types"
-	types2 "github.com/tendermint/tendermint/abci/types"
 	"math/rand"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
+
+	types2 "github.com/tendermint/tendermint/abci/types"
+
```
msmania, line 13: remove this empty line

x/pocketcore/types/crypto.go
```diff
@@ -3,12 +3,14 @@ package types
...
 	"encoding/hex"
-	"github.com/pokt-network/pocket-core/crypto"
-	sdk "github.com/pokt-network/pocket-core/types"
+	"math/big"
+
 	"github.com/tendermint/tendermint/crypto/tmhash"
 	"github.com/tendermint/tendermint/privval"
 	_ "golang.org/x/crypto/sha3"
-	"math/big"
+
```
msmania, line 11: remove this empty line

x/pocketcore/types/proof.go
```diff
@@ -118,15 +128,25 @@ func (rp RelayProof) ValidateBasic() sdk.Error {
 	if rp.Entropy < 0 { // todo this is inefficient
 		return NewInvalidEntropyError(ModuleName)
 	}
-	// verify a valid token
+
+	// verify the AAT token
```
msmania: T in AAT means Token.  Simply say `// validates the AAT in the relay proof`?

doc/specs/application-auth-token.md
```diff
@@ -1,10 +1,25 @@
...
+
+  The Pocket Network protocol contemplates the use of Application Auth Tokens (AATs)
+  to allow Gateways (off-chain) access Service Nodes (a.k.a Suppliers) on behalf
+  of the Application.
 ---
 
-# Application Authentication Token
+# Application Authentication Token <!-- omit in toc -->
+
+- [Data Structure Schema](#data-structure-schema)
```
msmania, line 22: It'd be reader-friendly to have simple description about the purpose of AAT.  How about this?

---
The purpose of an AAT is to define two separate entities:

1. Application Owner to stake an application on the blockchain
2. Client Owner to submit relays through any type of dApp (from a command-line tool to Portal-like website)

The key point here is one entity doesn't have to share their private key with the other entity.  Application Owner can delegate the right to use their staked app to Client Owner by passing an AAT including the client public key signed by the application private key.

doc/specs/application-auth-token.md
```diff
@@ -20,26 +35,80 @@ A semver string specifying the spec version under which this AAT needs to be int
...
-The protocol wide ed25519 ECDSA will be used for any signatures and verifications that are used within this
-specification.
+The protocol wide ed25519 ECDSA will be used for any signatures and verifications
+that are used within this specification.
+
+## AAT Generation
+
+When generating a new AAT, the owner of the `Application` private key has two options:
+
+1. Use the same public key for signing the AAT and the relay requests
```
msmania: If it's "When generating a new AAT", what the application owner can choose is which public key to be set as clientPublicKey.

doc/specs/application-auth-token.md
msmania: In the protocol level, there is no "charge".  Once the client owner gets an AAT from the app owner, they can submit infinite (up to max relays per session) number of relays on behalf of the app until the app owner unstakes, right?

I think the essence of this section is

- AAT includes client public key, signed by application private key
- Relay request includes an AAT, signed by client private key

What can be achieved by this fact is delegation. as mentioned in the earlier comment.

msmania (changes requested):
I think the main purpose of this PR is to add description about AAT rather than general documentation improvement.  Can you update the PR description?

- This PR contains redundant descriptions.  Can you put the most detailed description in .md file under doc/specs or in https://github.com/pokt-network/docs-v2, and make other places refer to it?  Code comments should be kept simple.
- I see some comments are not necessary because they're just rephrasing a code.
- Regarding AAT, I know PNF or Grove are introducing a new term "Gateway" to call what is used to be called "Client".  Having a comment like "(a.k.a gatewayPubKey)" everywhere is confusing.  In v0 code, can we just keep using clientPubKey?  Since it's defined as clientPubKey in the code, it's just clientPubKey.
- You emphasize "appPubKey and clientPubKey may or may not be the same", but I think it's not that important.  It's ok to say it in application-auth-token.md as a note, but not as code comments.

app/cmd/cli/app.go
Olshansk: Keeping as is. This is actually automatically done by `goimports-reviser` which is a new linter I imported. It separates `pokt-network` impros from other imports.

app/util.go
Olshansk: @adshmh Bump

doc/specs/application-auth-token.md
Olshansk: Pushing back on this. It's precisely one of the things that led to confusion for Grove and other gateways, so calling it out explicitly (thought a bit redundant) helps the reader.

doc/specs/application-auth-token.md
Olshansk: Awesome. Ty!

app/cmd/cli/app.go
```diff
@@ -179,10 +180,17 @@ command, you must have the private key of the current staked app <fromAddr>
...
+
+This CLI is hard-coded to generate an AAT with spec version 0.0.1.
+
+The output is intended to be embedded into the Gateway for Relay servicing.
+
+Upon AAT generation, the user will be prompted for the <appAddr> account passphrase.
+
+<appAddr> is associated with the application
+
+Make sure to read doc/specs/application-auth-token.md to understand what's recommended during AAT configuration.`,
```
msmania: If we say this, application-auth-token.md needs to recommend something. :)

x/pocketcore/types/proof.go
```diff
@@ -118,15 +123,25 @@ func (rp RelayProof) ValidateBasic() sdk.Error {
...
 		return NewInvalidTokenError(ModuleName, err)
 	}
-	// verify the client signature on the Proof
-	if err := SignatureVerification(rp.Token.ClientPublicKey, rp.HashString(), rp.Signature); err != nil {
-		return err
-	}
-	return nil
+
+	// Retrieving the client public key (aka the gateway public key), which
+	// may or may not be the same as the application public key depending on
```
msmania: You shouldn't say "may or may not be the same" here.  This function doesn't care whether `rp.Token.ClientPublicKey` equals to `rp.Token.ApplicationPublicKey` or not.  What this code does is just to verify the signature `rp.Signature` is signed by `ClientPublicKey`.  Even though it's the same as `ApplicationPublicKey`, what we use to check is `ClientPublicKey`.  Adding more context confuses readers.

x/pocketcore/types/proof.go
```diff
@@ -54,9 +55,13 @@ func (ps ProofIs) FromProofI() (res Proofs) {
 
 var _ Proof = RelayProof{} // ensure implements interface at compile time
 
-// "ValidateLocal" - Validates the proof object, where the owner of the proof is the local node
-func (rp RelayProof) ValidateLocal(appSupportedBlockchains []string, sessionNodeCount int, sessionBlockHeight int64, verifyAddr sdk.Address) sdk.Error {
-	//Basic Validations
+// ValidateLocal validates the Relay Proof object is aligned with `sessionBlockHeight` and `verifyAddr`.
+func (rp RelayProof) ValidateLocal(
+	appSupportedBlockchains []string,
+	sessionNodeCount int, // TODO_TECHDEBT: This is not used an can be removed.
```
msmania, line 61: Instead of introducing TODO markers (which will be forgotten), can you open a new issue in the repo?

x/pocketcore/types/crypto.go
```diff
@@ -38,29 +40,31 @@ func NetworkIdentifierVerification(hash string) sdk.Error {
 	return nil
 }
 
-// "SignatureVerification" - Verify the signature using hex strings
-func SignatureVerification(publicKey, msgHex, sigHex string) sdk.Error {
-	// decode the signature from hex
-	sig, err := hex.DecodeString(sigHex)
+// SignatureVerification verifies that:
+// - The privKey associated with publicKey generated sigHex
+// - sigHex signed msgHex
```
msmania, line 45: How about saying this:

SignatureVerification verifies `sigHex` is a signature of `msgHex` signed by the private key associated with `publicKeyHex`.

I don't think all comments in the function are not needed because they're obvious, but it's up to you..

x/pocketcore/types/aat.go
```diff
@@ -104,13 +104,12 @@ func (a AAT) ValidateMessage() error {
 	return nil
 }
 
-// "ValidateSignature" - Confirms the signature field of the AAT
+// ValidateSignature confirms that the ApplicationSignature of the AAT is correct
```
msmania, line 107: How about this?

ValidateSignature validates the given AAT by confirming its `ApplicationSignature` is signed by its `ApplicationPublicKey`.

x/pocketcore/keeper/proof.go
```diff
@@ -4,18 +4,23 @@ import (
...
 	"github.com/pokt-network/pocket-core/x/auth"
 	"github.com/pokt-network/pocket-core/x/auth/util"
 	pc "github.com/pokt-network/pocket-core/x/pocketcore/types"
-	"github.com/tendermint/tendermint/rpc/client"
-	"math"
-	"reflect"
-	"time"
 )
 
+// This is how servicers (a.k.a suppliers a.k.a as nodes a.k.a node runners) send proofs to the network.
```
msmania: What's this comment??  I think we should drop this.

x/pocketcore/keeper/aat.go
```diff
@@ -8,23 +8,28 @@ import (
 	pc "github.com/pokt-network/pocket-core/x/pocketcore/types"
 )
 
-// "AATGeneration" - Generates an application authentication token with an application public key hex string
-// a client public key hex string, a passphrase and a keybase. The contract is that the keybase contains the app pub key
-// and the passphrase corresponds to the app public key keypair.
-func AATGeneration(appPubKey string, clientPubKey string, key crypto.PrivateKey) (pc.AAT, sdk.Error) {
-	// create the aat object
+// GenerateAAT generates an AAT to be used for relay servicing.
+// - appPubKey is the public key of the application that's paying for on-chain service.
```
msmania: Again, "paying" is an offchain concept.  We shouldn't say it in code.  Mainnet is not the only one network in the world.  From developer's perspective, appPubKey is just the public key of a staked app.

x/pocketcore/keeper/aat.go
```diff
@@ -8,23 +8,28 @@ import (
...
 	aat := pc.AAT{
 		Version:              pc.SupportedTokenVersions[0],
 		ApplicationPublicKey: appPubKey,
 		ClientPublicKey:      clientPubKey,
 		ApplicationSignature: "",
 	}
-	// marshal aat using json
-	sig, err := key.Sign(aat.Hash())
+
+	// marshal the AAT structure
```
msmania, line 23: If code looks like generating a hash but its comment says marshal, it's confusing.  This comment is not needed.

x/pocketcore/keeper/aat.go
```diff
@@ -8,23 +8,28 @@ import (
...
+	aatBytes := aat.Hash()
+
+	// This is where the `ApplicationPrivKey` signs (i.e. delegates trust) to
```
msmania, line 26: I understand your intention to add "(i.e. delegates trust)", but I think it's not good for general code readers.  Your comments seems to try to educate general people who want to understand what is AAT and the comments make sense when they read all AAT-related code comments.  We should do that in docs, not in code comments.  Code comment should be just to help developers understand unobvious context.

In this case, for example, code apparently looks signing `aatBytes` and getting its signature.  I don't think we need this comment.

app/cmd/cli/app.go
msmania: Can you elaborate that confusion a little bit more?  For example, if somebody is confused because they don't understand why there are two public keys in AAT, it's probably because they don't understand the role of those keys.  Saying Application may or may not be the same as Client doesn't explain it.

msmania: As I mentioned in another comment, we should recomment something in this doc.  In my understanding, application shouldn't be the same as client.  Here's why.

Application private key is very confidential.  Once it's stolen, they can make relays for free.  Relayer application needs to store client key in their application.  This means if they use the same key as app and their application is hacked, their application is hacked.  If they separate application from client, they can unstake app.

Is there any benefit from sharing app and client?

msmania (changes requested):

POKT-Discourse:

https://forum.pokt.network/t/block-sizes-claims-and-proofs-in-the-multi-gateway-era/5060/9

x/pocketcore/keeper/aat.go
Olshansk: Burn is disabled in Morse, but it is technically implemented (just old, untested, etc...).

Olshansk:
@msmania  @kutoft @ezeike @nodiesBlade @commoddity @adshmh Wanted to bump this PR. PTAL when you have a chance.

app/cmd/cli/app.go
```diff
@@ -179,10 +180,17 @@ command, you must have the private key of the current staked app <fromAddr>
 var createAATCmd = &cobra.Command{
 	Use:   "create-aat <appAddr> <clientPubKey>",
 	Short: "Creates an application authentication token",
-	Long: `Creates a signed application authentication token (version 0.0.1 of the AAT spec), that can be embedded into application software for Relay servicing.
-Will prompt the user for the <appAddr> account passphrase.
-Read the Application Authentication Token documentation for more information.
-NOTE: USE THIS METHOD AT YOUR OWN RISK. READ THE APPLICATION SECURITY GUIDELINES IN ORDER TO UNDERSTAND WHAT'S THE RECOMMENDED AAT CONFIGURATION FOR YOUR APPLICATION.`,
+	Long: `Creates a signed Application Authentication Token.
+
```
msmania, line 186: Why do you put an empty line between every sentence?

app/cmd/cli/app.go
```diff
@@ -179,10 +180,17 @@ command, you must have the private key of the current staked app <fromAddr>
...
+This CLI is hard-coded to generate an AAT with spec version 0.0.1.
```
msmania: Two corrections
- What is hardcoded is not "CLI" but the version
- Upon AAT generation, \<appAddr\> can be any address

How about this?

---
Creates a signed AAT (= Application Authentication Token) where the version is
hardcoded as "0.0.1" that is the only version supported by the protocol.

This command prompts you to input the \<appAddr\> account passphrase.
When you send a relay request with AAT, \<appAddr\> needs to be a staked
application.

Please read doc/specs/application-auth-token.md for additional details.

---

app/util.go
```diff
@@ -15,8 +15,12 @@ import (
 	pocketKeeper "github.com/pokt-network/pocket-core/x/pocketcore/keeper"
 )
 
-func GenerateAAT(appPubKey, clientPubKey string, key crypto.PrivateKey) (aatjson []byte, err error) {
-	aat, er := pocketKeeper.AATGeneration(appPubKey, clientPubKey, key)
+// GenerateAAT generates an AAT to be used for relay servicing.
```
msmania: "relay request authentication" rather than "relay servicing"?

doc/specs/application-auth-token.md
```diff
@@ -1,10 +1,35 @@
 ---
 description: >-
-  Version 0.0.1. The Pocket Network protocol contemplates the use of Application Auth Tokens to allow
-  Application Clients to access Service Nodes on behalf of the Application.
+  Version 0.0.1
+
+  The Pocket Network protocol contemplates the use of Application Auth Tokens (AATs)
```
msmania: s/Auth/Authentication/

doc/specs/application-auth-token.md
```diff
@@ -20,26 +45,82 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+The protocol wide ed25519 ECDSA will be used for any signatures and verifications
+that are used within this specification.
+
+## AAT Generation
+
+When generating a new AAT, the owner of the `Application` private key can choose
+which `Client` public key is used to sign Relays. More explicitly, this can be
+decomposed into two options:
+
+1. **Sovereign Application** - Use the same public key for signing the AAT and the relay requests
```
msmania: What is "Sovereign Application"?  What I expect to see here is in which case it's recommended or required to use the same key on app and client in AAT and why.

doc/specs/application-auth-token.md
```diff
@@ -20,26 +45,82 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+        App -- AppPublicKey --> AATU
+        App -- ClientPublicKey --> AATU
+        AATU -- hash --> AATUB[AAT Bytes]
+        AATUB -- Sign with AppPrivKey --> sig[Application Signature]
+        AATU -- AAT Structure --> AAT
+        sig -- Signature --> AAT
+    end
+```
+
+### 2. Application !== Gateway (Client)
```
msmania: `Client (Gateway)` to match the first bullet point

doc/specs/application-auth-token.md
```diff
@@ -20,26 +45,82 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+## AAT Generation
+
+When generating a new AAT, the owner of the `Application` private key can choose
+which `Client` public key is used to sign Relays. More explicitly, this can be
+decomposed into two options:
+
+1. **Sovereign Application** - Use the same public key for signing the AAT and the relay requests
+2. **Gateway** - Use a different public key for signing the AAT and the relay requests
+
+### 1. Application === Client (Gateway)
```
msmania: You should put the `Application !== Client` case first because that's a general case.  This case of `Application === Client` is just a special version of it.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +129,47 @@ The proper way to sign the token is as follows:
...
+### Relay Generation & Signing using AAT
 
-`Message = SHA3-256(AATBytes)`
+This section can be summarized by two key points:
 
-`AAT.Signature = ED25519.Sign(Message)`
+- AAT includes client public key, signed by application private key
+- Relay request includes an AAT, signed by client private key
+
+The `Client` is needed to sign the relays while the `Application` is the one that
```
msmania: How about rephrasing the first sentence like this?  It would be good to mention the max relays.

The `Client`, an application to send relays, needs to sign every relay request, while `Application` needs to be staked once with tokens on-chain to get access to Pocket Network's services.  In every session, the maximum number of relays for an app is proportional to the tokens to be staked with the app.  To make more relays, you need to stake more tokens.

doc/specs/application-auth-token.md
```diff
@@ -48,17 +129,47 @@ The proper way to sign the token is as follows:
...
+needs to stake on-chain to get access to Pocket Network's services. Recall that
```
msmania: AAT is always signed by the application private key, which is shown in the two diagram above.  No exception.  Let's cut the second sentence starting with "Recall that...".

doc/specs/cli/apps.md
```diff
@@ -20,13 +20,13 @@ just enter the current parameter value \(the same one you entered for your initi
 
 Arguments:
 
-* `<fromAddr>`: Target Address to stake.
-* `<amount>`: The amount of uPOKT to stake. Must be higher than the current value of the `ApplicationStakeMinimum`
+- `<fromAddr>`: Target Address to stake.
+- `<amount>`: The amount of uPOKT to stake. Must be higher than the current value of the `ApplicationStakeMinimum`
   parameter, found [here](https://docs.pokt.network/learn/protocol-parameters/#applicationstakeminimum).
```
msmania: This url is dead.  Let's remove.

doc/specs/cli/apps.md
```diff
@@ -20,13 +20,13 @@ just enter the current parameter value \(the same one you entered for your initi
...
-* `<relayChainIDs>`: A comma separated list of RelayChain Network Identifiers. Find the RelayChain Network
+- `<relayChainIDs>`: A comma separated list of RelayChain Network Identifiers. Find the RelayChain Network
   Identifiers [here](https://docs.pokt.network/supported-blockchains/).
```
msmania: This url is dead. Let's remove.

x/pocketcore/keeper/proof.go
```diff
@@ -4,18 +4,22 @@ import (
...
 	"github.com/pokt-network/pocket-core/x/auth"
 	"github.com/pokt-network/pocket-core/x/auth/util"
 	pc "github.com/pokt-network/pocket-core/x/pocketcore/types"
-	"github.com/tendermint/tendermint/rpc/client"
-	"math"
-	"reflect"
-	"time"
 )
 
+// Nothing actionable (requests have already been signed by this point), but just FYI.
```
msmania: Is this a copy-n-paste mistake?

msmania (changes requested):

app/util.go
```diff
@@ -15,8 +15,12 @@ import (
 	pocketKeeper "github.com/pokt-network/pocket-core/x/pocketcore/keeper"
 )
 
-func GenerateAAT(appPubKey, clientPubKey string, key crypto.PrivateKey) (aatjson []byte, err error) {
-	aat, er := pocketKeeper.AATGeneration(appPubKey, clientPubKey, key)
+// GenerateAAT generates an AAT to be used for relay request authentication.
+// - appPubKey is the public key of the application that's paying for on-chain service.
```
msmania: s/paying/staking/

x/apps/keeper/querier_test.go
```diff
@@ -30,7 +31,7 @@ func Test_queryApplications(t *testing.T) {
 		want  []byte
 		want1 sdk.Error
 	}{
-		{"Test query applicaitons", args{
+		{"Test query applications", args{
```
msmania, line 34: I grepped the code and found three more places with this typo.  Can you correct those too?

```
$ grep -inr applicaitons --include=*.go
x/apps/types/errors_test.go:63:func TestError_ErrApplicaitonStatus(t *testing.T) {
x/apps/keeper/appStateChanges_test.go:56:func TestAppStateChange_ValidateApplicaitonStaking(t *testing.T) {
x/apps/keeper/querier_test.go:47:                               t.Errorf("queryApplicaitons() got1 = %v, want %v", got1, tt.want1)
```

doc/specs/application-auth-token.md
```diff
@@ -20,26 +45,82 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+The protocol wide ed25519 ECDSA will be used for any signatures and verifications
+that are used within this specification.
+
+## AAT Generation
+
+When generating a new AAT, the owner of the `Application` private key can choose
+which `Client` public key is used to sign Relays. More explicitly, this can be
+decomposed into two options:
+
+1. **Application Key != Client Key ** - Recommended for all Gateways.
```
msmania: Text is not bolded here.  Remove a whitespace between "Key" and "**"

doc/specs/application-auth-token.md
```diff
@@ -20,26 +45,82 @@ A semver string specifying the spec version under which this AAT needs to be int
...
+2. **Application Key == Client Key** - Recommended for independent Applications that are not Gateways.
```
msmania: This part is still confusing to me.  Having "Recommended for all Gateways" and "Recommended for independent Applications that are not Gateways" in the same list doesn't sound right.  The first item says "for gateways" while the second item says "for Applications", but these are two different concepts.  Application is not Gateway.

I think it's fair to say we create AAT for gateways because a gateway (aka client, relayer) is a program to use AAT.  Then the second case confuses me.  Do you mean there is a situation where we can use AAT without any gateways?  Can you elaborate that case a little bit more?

doc/specs/cli/apps.md
```diff
@@ -61,38 +61,31 @@ Transaction submitted with hash: <Transaction Hash>
 pocket apps create-aat <appAddr> <clientPubKey>
 ```
 
-Creates a signed application authentication token \(version `0.0.1` of the AAT spec\), that can be embedded into
-application software for Relay servicing. Will prompt the user for the `<appAddr>` account passphrase.
+This CLI is hard-coded to generate an AAT with spec version 0.0.1.
```
msmania: Please update this to match the actual command output.

msmania (changes requested):

doc/specs/cli/apps.md
Olshansk: This was the output. How should I update the docs?

```bash
 pocket apps create-aat 0551ce49b10bd603c85b8de89434686efdbb9c7a aaa
2024/03/11 19:47:48 Initializing Pocket Datadir
2024/03/11 19:47:48 datadir = /Users/olshansky/.pocket
Enter passphrase:
{
  "version": "0.0.1",
  "app_pub_key": "1789ebc8b3a37ebb07b6a9bd8716f34a67004c61f2120a8c1f76560474d8e4f0",
  "client_pub_key": "aaa",
  "signature": "8f07d21bc90fe571fcc326d0ba281eb71f61ddc781abe8f125ebe27ea74a3a55c09f7e4c19744dc7488b362a022e767b0d4e57965c0461dfc97f11c0996d2101"
}
```

doc/specs/application-auth-token.md
Olshansk: You should watch my talk! https://www.youtube.com/watch?v=jVW-3lRzVT0&t=1s

Docs also updated.

Olshansk: I added this:

```
_Note: Please note that option 2 is only shown for theoretical purposes. Due to
the scaling limitations of Morse (v0) implemented in `pocket-core`, Gateways
have to be used. The original design / intention of Pocket, and how Shannon (v1)
will operate is support both modes._
```

If this is not enough and you want me to write out the whole story, I'd rather just delete these updates ot the docs altogether.

Lmk what you prefer.

doc/specs/cli/apps.md
msmania: I mean the output of `pocket apps create-aat --help`.  We updated the text in app/cmd/cli/app.go, so this needs to match it.

doc/specs/application-auth-token.md
msmania: Hmm, looks like we're not yet on the same page.  Before just stepping back, let's discuss it more.  I think it's important.  If I miss something, I want to know it.

First of all, I watched https://www.youtube.com/watch?v=jVW-3lRzVT0.  That was a great talk!

I watched "Sovereign Application" part at around 15:00 carefully and if I understand that part correctly, the Sovereign Application scenario still requires a gateway, right?  As you said "they talk to the suppliers themselves" in the talk, they have their own program to submit relays, which is a gateway.

I guess you're calling only a delegated client "gateway".  If so, "gateway" is just one usecase of "client" and we can't say "client, aka gateway".  Whether it's delegated or not, a client is a program to submit relays.  Actually I personally prefer calling it "relayer".

Anyway, the question is if "Sovereign Application" operator doesn't delegate the relayer part to anyone and opereates it themselves, is it recommended to use the same key for the Gateway and App?  My answer is no.

To me, this question is almost the same as "can we use the same password for multiple services, like Amazon.com account and Google account?".  The benefit from doing that is we can reduce the number of passwords to be managed, but of course the answer is "we can, but it's not recommended because it sacrifices security".

I think Application/Gateway has the same story.

In whatever scenario, a gateway program needs to access client private key to send a request.  It needs to load the key in a raw format at some point.  Access to application private key, on the other hand, is not required if they precreate an AAT.  And if they do so, even if their server is hacked, application private key is safe.  Good.  If they use the same key for Client and Application, application private key is leaked when their client is hacked, and the staked tokens will be at stake really.

Maybe I'm getting paranoid to this.  Probably the server will not be hacked.  Managing less private keys is more beneficial.  But if you *recommend* something, you need to clarify pros and cons, and explain why.  Otherwise people cannot choose which fits their system "Application Key != Client Key" or "Application Key == Client Key" .

x/apps/keeper/querier_test.go
msmania: Ah, my bad.  I only grepped a plural "applicaitons".  There are more singular ones.

Actually I don't care these typos at all.  It's totally fine to leave them.  This PR is to explain AAT, not a boring typo-fixing work (which should be done by AI...).  Instead of correcting them, it would be good to file a new issue and forget :)

```
$ grep -inr applicaiton
app/cmd/cli/query.go:538:               params := rpc.HeightAndApplicaitonOptsParams{
app/cmd/rpc/rpc_test.go:643:    var params = HeightAndApplicaitonOptsParams{
app/cmd/rpc/rpc_test.go:659:    params = HeightAndApplicaitonOptsParams{
app/cmd/rpc/query.go:78:type HeightAndApplicaitonOptsParams struct {
app/cmd/rpc/query.go:732:       var params = HeightAndApplicaitonOptsParams{}
x/apps/types/errors_test.go:163:func TestError_ErrApplicaitonPubKeyExists(t *testing.T) {
x/apps/types/errors_test.go:238:func TestError_ErrBadApplicaitonAddr(t *testing.T) {
x/apps/types/errors_test.go:258:                                t.Errorf("ErrBadApplicaitonAddr(): returns %v but want %v", got, tt.want)
x/apps/keeper/application_test.go:40:                           t.Errorf("Applicaiton.GetApplication() = got %v, want %v", found, tt.want)
x/apps/keeper/application_test.go:155:                          t.Errorf("Applicaiton.GetAllApplications() = got %v, want %v", got, tt.want)
x/apps/keeper/application_test.go:184:                          t.Errorf("Applicaiton.GetAllApplications() = got %v, want %v", got, tt.want)
x/apps/keeper/appStateChanges_test.go:15:func TestAppStateChange_ValidateApplicaitonBeginUnstaking(t *testing.T) {
x/apps/keeper/appUnstaked.go:19:// deleteUnstakingApplicaiton - DeleteEvidence an application address from the unstaking queue
```

msmania (changes requested):

doc/specs/application-auth-token.md
Olshansk: The goal is to deprecate `pocket-core` and focus on `poktroll`, so I don't want to make a big deal out of this. This was a "side quest".

How about I just delete this part of the documentation to avoid confusion and paranoia?
👍?

x/apps/keeper/querier_test.go
msmania: https://github.com/pokt-network/pocket-core/issues/1601

doc/specs/application-auth-token.md
msmania: We discussed in Discord and agreed:
- We should not bring Shannon concepts to this repo
- Client in v0 is not Gateway in Shannon
- Client pubkey may be the same as App pubkey, but it's not recommended in v0

msmania (approved):
Sorry for taking long.  Thanks!

PR #1603, opened 2024-05-02
Docker Image enhancement and better CLI support.

jorgecuesta:
## THIS PR INCLUDES BREAKING CHANGES
- Dockerfile now uses `app` user instead of `root`

#### Why?

A few Geo-Mesh users report to POKTscan about an issue with the new image after we adopt the one on the pokt-network/pocket-core repository on the latest RC.

They report that this image is using `root` as the user which is recommended to be avoided. There are a lot of blogs and documentation about this, [here](https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security) one of them from a well-known docker image user/company.

Also, we detected a few things that could be enhanced on both, entry point and docker context.

The problem with having a public image using root right now is that pocket binary generates folders and files that now belong to the `root` user, so they will need to modify those permissions to belong to the proper `app` user and group. To this, I added another optional entry point that could be used once to fix the permission issue and then start the container as before.

[Here](https://github.com/pokt-network/pocket-core/pull/1603/files#diff-8a9d880ecb3d20b3cfe8def4da1bd200fcbd44131caee95ea699a34faf9cfd6fR19) you can see how to use it with docker-compose or docker

#### Changes:

- Modifications to Dockerfile allow the container to run as `app` user instead of `root`.
- Added a new shell script named fix_permissions.sh to fix ownership issues related to running containers.
- Updated Dockerfile to use this new script.
- Added a .dockerignore file to help maintain a cleaner Docker build context, excluding unnecessary files. 
- Modifications to `entrypoint.sh` allows the user to run all its internal commands with the proper `--datadir` param. Now properly handle the start command when `--keybase=false` is sent. Also, allow the user to pass the `--datadir` as an env variable to omit it on the start command.

### DIFF
--- a/.github/workflows/entrypoint.sh
+++ b/.github/workflows/entrypoint.sh
@@ -54,17 +54,17 @@ if {$config != ""} {
 }
 
 if {[regexp -nocase "datadir=" $command] && ![regexp -nocase "datadir=${datadir}" $command]} {
-  send_user "WARNING: --datadir provided with a different path to the one defined on Dockerfile. This could lead to errors when use this entrypoint to run CLI commands.\n"
+  send_user "WARNING: --datadir provided with a different path than the one defined in the Dockerfile. This could lead to errors using pocket CLI commands.\n"
 } elseif {![regexp -nocase "datadir=" $command]} {
-  send_user "INFO: param --datadir was not provided; attaching ${datadirParam} on every command executed with this entrypoint\n"
+  send_user "INFO: param --datadir was not provided; attaching ${datadirParam} on every command\n"
   set command "${command} ${datadirParam}"
 }
 
 if {![regexp -nocase "datadir=${defaultDatadir}" $command]} {
 	send_user "WARNING: --datadir is not the default one
 Please review:
-1. Mount your config folder to the same path you specify on --datadir
-2. Review your config.json on the following points to match with the value of --datadir
+1. Mount your config folder to the same path you specify in --datadir
+2. Review your config.json to ensure the following configs match the value of --datadir
   2.1. tendermint_config.RootDir
   2.2. RPC.RootDir
   2.3. P2P.RootDir
--- a/.github/workflows/entrypoint.sh
+++ b/.github/workflows/entrypoint.sh
@@ -12,11 +12,12 @@ trap graceful_exit {SIGINT SIGTERM}
 set command $argv
 set timeout -1
 
-# Create work dir
-spawn sh -c "mkdir -p /home/app/.pocket/config"
-expect eof
-
 # Pull variables from env if set
+set defaultDatadir "/home/app/.pocket"
+set datadir $defaultDatadir
+catch {set datadir $env(POCKET_CORE_DATADIR)}
+set datadirParam "--datadir=${datadir}"
+
 set genesis ""
 catch {set genesis $env(POCKET_CORE_GENESIS)}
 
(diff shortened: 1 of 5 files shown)


Olshansk:
@okdas Please review this when you're back in office next week and see if any of the learnings / suggestions can transfer over to #495 which we should pick up then as well.

.github/workflows/fix_permissions.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+datadir=$1
+echo "Attempting to fix ${datadir} permissions to be own by app:app"
```
Olshansk:
```suggestion
echo "Attempting to fix ${datadir} permissions to be owned by app:app"
```

.github/workflows/fix_permissions.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
Olshansk: Can you rename this file to something more explicit please. For example: `change_datadir_permissions_to_app.sh`

.github/workflows/fix_permissions.sh
```diff
@@ -0,0 +1,7 @@
...
+datadir=$1
+echo "Attempting to fix ${datadir} permissions to be own by app:app"
+chown -R app:app $datadir
+echo "${datadir} permissions applied."
+echo "Please turn off entrypoint override and ensure you are using user app or user 1005 when start container."
```
Olshansk:
```suggestion
echo "Please turn off entrypoint override and ensure you are using user `app` or user `1005` when start container."
```

.github/workflows/fix_permissions.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+datadir=$1
+echo "Attempting to fix ${datadir} permissions to be own by app:app"
+chown -R app:app $datadir
```
Olshansk: Why do we need to do this?

No need to reply in PR, just #PleaseUpdateComment (#PUC) in the code..

.github/workflows/fix_permissions.sh
```diff
@@ -0,0 +1,7 @@
...
+echo "${datadir} permissions applied."
```
Olshansk: Update comment on what types of permissions.

.github/workflows/fix_permissions.sh
Olshansk: Overall, need to add comments & descriptions in this file to understand what & why we're doing so it's clear to people like me without needing to have access to ask questions from experts like you.

.github/workflows/entrypoint.sh
```diff
@@ -26,42 +27,83 @@ catch {set chains $env(POCKET_CORE_CHAINS)}
...
+    set config_file [open "${datadir}/config/config.json" w]
     puts $config_file $config
     close $config_file
     send_user "CONFIG loaded from env\n"
 }
 
-# If key isn't passed in, start the node
-if { $env(POCKET_CORE_KEY) eq "" }  {
+if {[regexp -nocase "datadir=" $command] && ![regexp -nocase "datadir=${datadir}" $command]} {
+  send_user "WARNING: --datadir provided with a different path to the one defined on Dockerfile. This could lead to errors when use this entrypoint to run CLI commands.\n"
```
Olshansk:
```suggestion
  send_user "WARNING: --datadir provided with a different path than the one defined in the Dockerfile. This could lead to errors when use this entrypoint to run CLI commands.\n"
```

.github/workflows/entrypoint.sh
```diff
@@ -26,42 +27,83 @@ catch {set chains $env(POCKET_CORE_CHAINS)}
...
+} elseif {![regexp -nocase "datadir=" $command]} {
+  send_user "INFO: param --datadir was not provided; attaching ${datadirParam} on every command executed with this entrypoint\n"
```
Olshansk: `with this entrypoint` - Which entrypoint? Consider updating the log.

.github/workflows/entrypoint.sh
```diff
@@ -26,42 +27,83 @@ catch {set chains $env(POCKET_CORE_CHAINS)}
...
+  set command "${command} ${datadirParam}"
+}
+
+if {![regexp -nocase "datadir=${defaultDatadir}" $command]} {
+	send_user "WARNING: --datadir is not the default one
+Please review:
+1. Mount your config folder to the same path you specify on --datadir
```
Olshansk:
```suggestion
1. Mount your config folder to the same path you specify in --datadir
```

.github/workflows/entrypoint.sh
```diff
@@ -26,42 +27,83 @@ catch {set chains $env(POCKET_CORE_CHAINS)}
...
+2. Review your config.json on the following points to match with the value of --datadir
```
Olshansk:
```suggestion
2. Review your config.json to ensure the following configs match the value of --datadir
```

Olshansk (changes requested):

Olshansk (approved):
I think this is great!

I do want one other pair of eyes from an operator to make sure I didn't miss anything, but it has a 👍  from me.

@okdas @msmania @nodiesBlade Can one of you PTAL and review?

nodiesBlade:
(Reviewed this on Geomesh branch as well), looks good to me.

nodiesBlade (approved):

Olshansk:
@okdas PTAL

okdas (approved):
I haven't pulled to verify locally but code changes look good.

Olshansk:
Re-running to make sure this passes but otherwise lgtm:

Screenshot 2024-05-17 at 3 38 42 PM

okdas:
@Olshansk it is not going to pass because the origin of the PR is a branch from `pokt-scan` organization which doesn't have permissions to push to our container registry.

Olshansk:
@okdas Appreciate the context. What do you recommend as the solution here?

For example:
1. We squash & merge as is?
2. Update permissions to allow their org to run CI on our repo?
3. Validate in some other way?
4. Etc?

I understand the risk is low but just looking to you to make a call here.

okdas:
@Olshansk the workaround would be to open a new PR under `pokt-network` org, but I don't think we need to do that and should proceed with merging as is. The CI job seems to be failing on the very last step - and the image itself is built successfully.

PR #1608, opened 2024-08-14
Add pwd flags to appTransferCmd

Olshansk:
Enable running `pocket apps transfer` with the `--pwd` flag so it can be automated.


msmania (approved):
Oh I didn't know this automation mechanism existed.  Looks good.

PR #1609, opened 2024-08-14
Update the seeds information in README.md

breezytm:
## Description

Replaced maintainers from Nodefleet.org to Stakenodes.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ An offline pruning tool is provided by [C0D3R](https://c0d3r.org/). The tool is 
 
 ## Accessing TestNet
 
-TestNet information can be found at [testnet.md](doc/guides/testnet.md) and is maintained by the [nodefleet.org](https://nodefleet.org/) team.
+TestNet information can be found at [testnet.md](doc/guides/testnet.md) and is maintained by the [Stakenodes](https://stakenodes.org/) team.
 
 ## Charts & Analytics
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,7 +117,7 @@ An offline pruning tool is provided by [C0D3R](https://c0d3r.org/). The tool is
 
 ## Accessing TestNet
 
-TestNet information can be found at [testnet.md](doc/guides/testnet.md) and is maintained by the [nodefleet.org](https://nodefleet.org/) team.
+TestNet information can be found at [testnet.md](doc/guides/testnet.md) and is maintained by the [Stakenodes](https://stakenodes.org/) team.
```
Olshansk, line 120: Please update https://github.com/pokt-network/pocket-core/blob/staging/doc/guides/testnet.md as well

Olshansk (changes requested):

README.md
breezytm: I will. I am going through the contents to determine what is actually needed and those that can be removed. I never had a request for tendermint access; therefore, it is unclear to me if that is something I should be making accessible.

Olshansk: > I never had a request for tendermint access; therefore, it is unclear to me if that is something I should be making accessible.

Yea, let's maintain parity with what the previous testnet maintainers had.

PR #1614, opened 2024-10-01
Parameter JSON serialization must be deterministic

msmania:
This is a fix for a regression introduced by Per-chain RTTM (#1580). When we store a map-type parameter, we use `Codec.MarshalJSON` that doen't guarantee the order of fields in a map.  This causes consensus failure because a different order of fields in the parameter store generates a different merkle tree (= AppHash), resulting in chainhalt.

The proposed fix is to canonicalize the marshaled bytes by calling `MustSortJSON` before storing it in the tree.

Changing `Subspace.Set` is enough and a change in ` Subspace.SetWithSubkey` won't be necessary because it's not used from anywhere, but we should keep consistency.

reviewpad:summary

### DIFF
--- a/types/param.go
+++ b/types/param.go
@@ -212,11 +212,11 @@ func (s Subspace) checkType(store KVStore, key []byte, param interface{}) {
 func (s Subspace) Set(ctx Ctx, key []byte, param interface{}) {
 	store := s.kvStore(ctx)
 	s.checkType(store, key, param)
-	bz, err := s.cdc.MarshalJSON(param)
-	if err != nil {
-		panic(err)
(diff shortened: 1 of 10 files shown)


Olshansk (approved):
@msmania LGTM and we can merge it as is.

In a followup PR, can we also tend to 

1. Fixing this: https://github.com/pokt-network/pocket-core/blob/staging/x/gov/keeper/subspace.go#L216

2. Migrating the recovery guide to docs: https://www.notion.so/Recovery-guide-641ec21aead74cae806166f5f9e61394

PR #1615, opened 2024-10-02
Version bump to RC-0.12.0

msmania:
Due to a consensus breaking change to fix a regression in RC-0.11.1 (#1614), this patch is upgrading the protocol version to RC-0.12.0.  That regression caused the chainhalt in the testnet at height block.  We're rolling back the testnet to recover it.  This version bump helps to make sure all validators are runing the latest version.

reviewpad:summary

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	AppVersion = "RC-0.11.1"
+	AppVersion = "RC-0.12.0"
 )
 
 // NewPocketCoreApp is a constructor function for PocketCoreApp


app/app.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	AppVersion = "RC-0.11.1"
```
Olshansk, line 30: This is a consensus breaking change, so we need to go with `0.12`

Olshansk (changes requested):
1. Can you also copy-pasta your summary from discord here (for the PR description).

```
Fix has been merged, but the version is still RC-0.11.1
PR of version bump to RC-0.11.2: https://github.com/pokt-network/pocket-core/pull/1615
We need to roll back the status earlier than 1550258
Snapshot at 155025: https://link.storjshare.io/s/jvws56e6ayjsukqualg36hq2j23a/pocket-snapshots/pocket-testnet@155025.zip?download=1
Validator set at 155025
 77e608d8ae4cd7b812f122dc82537e79dd3565cb breezytm
 7674a47cc977326f1df6cb92c7b5a2ad36557ea2 Ian
 c7b7b7665d20a7172d0c0aa58237e425f333560a Ian
 5b18a8c268ffcbf0530f61f70e3ee14f064bdf0f c0d3r (me)
 6925c38c9303a7a1864e9dfcc85b86f9c150519a MissKitty42
 ```

msmania:
> 1. Can you also copy-pasta your summary from discord here (for the PR description).

I rewrote the description.  I think that discord post doesn't fit description (for example, the snapshot url will be expired).

app/app.go
msmania: Ah right. Changed it to RC-0.12.  Ideally we should do BETA, but let's cut that step this time.

Olshansk (approved):

